PR #3, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -1,0 +1,20 @@
+name: Octo Issue Feed
+
+on:
+  issues:
+    types: [opened, closed, reopened, labeled]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      issue_number: ${{ github.event.issue.number }}
+      issue_title: ${{ github.event.issue.title }}
+      issue_url: ${{ github.event.issue.html_url }}
+      issue_author: ${{ github.event.issue.user.login }}
+      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
+      event_action: ${{ github.event.action }}
+      project_group_id: a48e308c2ead4d9dbcdc30fff0f01eaf
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,0 +1,23 @@
+name: Octo PR Feed
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, ready_for_review, review_requested]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      pr_number: ${{ github.event.pull_request.number }}
+      pr_title: ${{ github.event.pull_request.title }}
+      pr_url: ${{ github.event.pull_request.html_url }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      pr_merged: ${{ github.event.pull_request.merged }}
+      pr_additions: ${{ github.event.pull_request.additions }}
+      pr_deletions: ${{ github.event.pull_request.deletions }}
+      pr_changed_files: ${{ github.event.pull_request.changed_files }}
+      event_action: ${{ github.event.action }}
+      project_group_id: a48e308c2ead4d9dbcdc30fff0f01eaf
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}


yujiawei (commented):
# Code Review — PR #3 (octo-admin)

## Summary

Two new workflow files that delegate to reusable workflows in `Mininglamp-OSS/.github` for pushing GitHub Issue/PR events into Octo channels. The change is small (43 additions, 2 files), well-scoped, and the caller-side YAML is correct. No blocking issues found at this layer; everything below is either advisory or a defense-in-depth recommendation.

Note: a full security/correctness verdict ultimately depends on the called reusable workflows in `Mininglamp-OSS/.github` (not visible in this PR). The comments below assume the callee handles untrusted inputs safely.

## What looks good

- Clean separation: caller files just wire up event payloads and forward `OCTO_BOT_TOKEN` via `secrets:`. All formatting/notification logic lives in the reusable workflow, which is the right factoring.
- `issue_labels` uses `toJson(...)`, which is the correct way to flatten an array into a single workflow input.
- Using `pull_request` (not `pull_request_target`) is the right call for a notification workflow — no privilege escalation surface from forks.
- Event type lists are sensible defaults (`opened, closed, reopened, labeled` for issues; `opened, closed, reopened, ready_for_review, review_requested` for PRs).

## Findings

### P1 — Pin the reusable workflow to a commit SHA, not `@main`

`.github/workflows/octo-issue-feed.yml:9` and `.github/workflows/octo-pr-feed.yml:9` both use:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
```

`@main` means every push to that file in the central `.github` repo immediately ships to every consumer with no review. For an internal org workflow that handles a bot token, prefer pinning to a tag (`@v1`) or commit SHA. This is GitHub's documented hardening recommendation for `uses:` references and avoids a cross-repo supply-chain footgun.

If `@main` is intentional (for fast iteration during rollout), please at least track a `RELEASE.md` / version policy in the central repo so downstream owners know when behavior changed.

### P2 — No `permissions:` block

Neither workflow declares `permissions:`. Depending on the repo/org default, the `GITHUB_TOKEN` may end up with `contents: write` or broader scopes that the notify job does not need. Recommend adding an explicit minimal block at the workflow level, e.g.:

```yaml
permissions:
  contents: read
  issues: read
  pull-requests: read
```

(Or even `permissions: {}` — the called reusable workflow can declare what it actually needs at the job level inside its own definition.)

### P2 — Fork PR behavior is silently broken

`pull_request` events from forked repos do not receive secrets, so `OCTO_BOT_TOKEN` will be empty and the called workflow will fail (or skip) for external contributor PRs. That's the safe default and probably acceptable for an internal org, but it's worth deciding deliberately:

- If external contributor PRs should also notify, you'll need a separate `pull_request_target`-based path with strict input handling, or a scheduled poller. Don't simply switch this file to `pull_request_target` — it would be dangerous given untrusted inputs are forwarded downstream.
- If they shouldn't, consider documenting this in a README or in a comment at the top of the workflow so future maintainers don't expect coverage.

### P3 — Untrusted input handling depends on the callee

`issue_title`, `pr_title`, `issue_author`, `pr_author` are all attacker-controlled strings (anyone who can open an issue/PR). At the caller layer here, passing them as typed `with:` inputs is safe — GitHub does not shell-interpolate them. **However**, this is only safe if the reusable workflow in `Mininglamp-OSS/.github`:

1. Never uses these inputs inside a `run:` block as `${{ inputs.foo }}` (which would be a script-injection vector — the standard mitigation is to read them via `env:` and reference `"$FOO"` in the script).
2. Never embeds them unescaped into a JSON body sent to Octo (could break message rendering or, worse, allow markup injection if Octo renders rich content).

Recommend the reviewer of the central `.github` repo verify both. Out of scope for this PR strictly, but listing here because the security posture of this PR is only as good as the callee's.

### P3 — Missing `concurrency:` group

For high-traffic repos, rapid label flips or repeated `synchronize` events can fan out into many concurrent runs. A simple guard would be:

```yaml
concurrency:
  group: octo-${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is intentional here — you want every state transition delivered, not collapsed. Optional, but cheap insurance.

### P3 — Event-type coverage gaps (design question)

Worth confirming these are intentional:

- Issue feed does **not** include `edited`, `assigned`, `unassigned`, `unlabeled`. If retitles or assignment changes are interesting to project channels, they're missed today.
- PR feed does **not** include `synchronize` (new commits pushed) or `converted_to_draft`. Both are commonly useful in PR feeds.
- `pr_merged` is forwarded as a string `"true"`/`"false"`. The callee needs to compare it as a string (`if: inputs.pr_merged == 'true'`), not as a boolean. Worth a quick check that it does.

These are design choices, not defects — flagging so the author can confirm the noise/signal balance is what they want.

### Nit — Workflow file ordering / consistency

The two files are structurally near-identical. Once this lands, consider whether they could collapse into a single workflow with a matrix or trigger split, or at least share a YAML anchor / be generated. Not for this PR, just a maintenance note: every time you tune one, you'll need to remember to tune the other.

## Verdict

`COMMENTED`. The change is well-factored and there are no must-fix issues at this layer. The two recommendations I'd most encourage before/right after merge are: (1) pin the reusable workflow to a SHA or version tag instead of `@main`, and (2) add an explicit minimal `permissions:` block. Everything else is advisory.

PR #4, opened 2026-05-13
ci: add Octo IM CI status notifier

lml2468:
Adds reusable Octo IM CI status notification (alerts on main CI status changes).

### DIFF
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -1,0 +1,20 @@
+name: Octo CI Status
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@main
+    with:
+      repo_name: ${{ github.event.workflow_run.repository.name }}
+      workflow_name: ${{ github.event.workflow_run.name }}
+      conclusion: ${{ github.event.workflow_run.conclusion }}
+      run_id: ${{ github.event.workflow_run.id }}
+      run_url: ${{ github.event.workflow_run.html_url }}
+      project_group_id: "a48e308c2ead4d9dbcdc30fff0f01eaf"
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}


yujiawei (commented):
# Code Review — PR #4 (octo-admin)

## Summary

A small, focused addition: a 20-line caller workflow that delegates to the org-wide reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@main` to push Octo IM notifications when the `CI` workflow on `main` changes state. The wiring is correct and matches the reusable workflow's input contract. A few non-blocking hardening suggestions below.

## Verification

- `.github/workflows/ci.yml` exists and is named `CI` — matches `workflows: ["CI"]` filter ✅
- All required inputs of the reusable workflow are passed: `repo_name`, `workflow_name`, `conclusion`, `run_id`, `run_url`, `project_group_id`, plus `OCTO_BOT_TOKEN` secret ✅
- `branches: [main]` on `workflow_run` filters on the originating run's head branch — `pull_request` runs targeting main will be skipped because their head_branch is the PR source branch, not `main`. So this correctly limits notifications to push-to-main runs only ✅
- `${{ github.event.workflow_run.id }}` resolves to a number from the event payload and satisfies the `type: number` input ✅
- `OCTO_BOT_TOKEN` is forwarded explicitly via `secrets:` rather than `secrets: inherit`, which is the safer pattern ✅

## Findings

### P2 — Missing explicit `permissions:` block (non-blocking, recommended)

`.github/workflows/octo-ci-status.yml` does not declare `permissions:`. The reusable workflow uses `secrets.GITHUB_TOKEN` to query `/repos/.../actions/runs`, which only needs `actions: read`. Org-default token permissions vary, and a least-privilege explicit block makes the contract obvious and survives org-wide hardening:

```yaml
permissions:
  actions: read
  contents: read
```

Add at the top level (or `jobs.notify.permissions:` — but top-level is fine since this workflow has only one job).

### P2 — No `concurrency:` group (non-blocking)

Two main pushes in quick succession will fan out two parallel `Octo CI Status` runs, each independently fetching "current vs previous" from the API. Under contention this can produce out-of-order notifications, or in the worst case both runs see the same "previous" and both fire. A simple guard:

```yaml
concurrency:
  group: octo-ci-status-${{ github.event.workflow_run.head_branch }}
  cancel-in-progress: false
```

(`cancel-in-progress: false` so we don't drop a state-change notification mid-flight.)

### P2 — Reusable workflow pinned to `@main` (non-blocking, policy decision)

`uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@main` will pull whatever lands on `.github`'s default branch. For an internal notifier this is usually acceptable and even desirable (centralized rollout). If this caller pattern gets copied across many repos, switching to a tag or commit SHA later requires touching every caller. Worth deciding the policy now and documenting it. No change required for this PR.

### Note — Hardcoded `project_group_id`

`project_group_id: "a48e308c2ead4d9dbcdc30fff0f01eaf"` is in plain YAML. This is a chat group identifier, not a credential — auth is via `OCTO_BOT_TOKEN`. Keeping it in the workflow file is fine and arguably better than another secret, since it's per-repo configuration and reviewable in the diff.

### Note — Outside this PR's scope (reusable workflow)

The reusable workflow declares `run_id` as a required input but never references it. Worth removing in a follow-up to `Mininglamp-OSS/.github`, but not this PR's concern.

## Recommendation

The PR is correct and ready to merge as-is. The two P2 suggestions (`permissions:` block and `concurrency:` group) are mild hardening that can land here or in a follow-up.

PR #6, opened 2026-05-13
fix: use pull_request_target for fork secret access

lml2468:
### Problem
Workflows triggered by `pull_request` from **forked repositories** do not receive repository secrets. This means `OCTO_BOT_TOKEN` is unavailable, and every external contributor's PR triggers a failed workflow run.

### Fix
Changed trigger from `pull_request` → `pull_request_target`.

This is safe because:
- The workflow only reads event payload metadata (PR title, author, URL)
- No PR code is checked out
- `pull_request_target` runs in the context of the base branch, giving access to secrets

Applied to all 8 repos.

### DIFF
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -4,9 +4,11 @@ on:
   pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
 
+permissions: {}  # GITHUB_TOKEN not used; lock down to minimum
+
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,7 +1,7 @@
 name: Octo PR Feed
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
 
 jobs:


Jerry-Xin (changes requested):
This CI workflow is relevant to `octo-admin`, but the `pull_request_target` privilege expansion needs explicit permission hardening before merge.

🔴 Blocking

🟡 Warning: `.github/workflows/octo-pr-feed.yml:7` does not define `permissions` for a `pull_request_target` workflow. This trigger runs in the base repository context and now receives secrets for fork-originated PR events. Even though this caller does not check out PR code, it invokes a reusable workflow with attacker-controlled PR metadata such as `pr_title` and `pr_author` at lines 13-15. Without explicit permissions, the called workflow may inherit broader `GITHUB_TOKEN` access than this notification job needs.

Please add least-privilege permissions, preferably:

```yaml
permissions: {}
```

If the reusable workflow requires repository reads, use:

```yaml
permissions:
  contents: read
```

Place it at the workflow or `notify` job level before enabling `pull_request_target`.

💬 Non-blocking

🔵 Suggestion: Consider documenting the safety invariant in this file, similar to `.github/workflows/pr-merged-done.yml:11`, because future edits to this workflow must preserve “no PR checkout, no execution of fork-controlled code.”

✅ Highlights

The PR addresses a real fork-PR secret availability problem, and the current caller only passes event payload fields to the centralized reusable workflow. No tests are required for this YAML-only workflow trigger change.

lml2468 (commented):
**CHANGES_REQUESTED** (submitted as comment due to GitHub self-review restriction)

This should not merge yet: I verified that the current call path does not execute PR code, but the new `pull_request_target` entry point still needs explicit least-privilege `permissions` before secrets are available to fork PR events.

Blocking items:

1. Supporting Jerry-Xin's existing blocker: `.github/workflows/octo-pr-feed.yml:4` switches this workflow to `pull_request_target`, but the caller does not scope `GITHUB_TOKEN` permissions before invoking the reusable workflow at `.github/workflows/octo-pr-feed.yml:9` with `OCTO_BOT_TOKEN`. The reusable workflow currently only posts metadata through inline Python and does not need repository write access, so please add `permissions: {}` or the narrowest required read permission. Without that hardening, any future change in the called workflow runs under a broader token than this notification job needs.

Non-blocking notes:

- I found no checkout of PR code and no action or script loaded from the PR branch in either the caller workflow or `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main`.
- The PR-controlled fields I traced (`pr_title`, `pr_author`, URL, counts, and action) are passed through environment variables and JSON-encoded by Python, not interpolated into shell code; I did not find command-evaluation injection in the current reusable workflow.
- Consider documenting the safety invariant in the caller: with `pull_request_target`, future edits must preserve "no PR checkout, no execution of fork-controlled code" while `OCTO_BOT_TOKEN` is available.

Highlights:

- The change addresses a real fork PR secret availability problem, and the current reusable workflow sends only PR notification metadata.

yujiawei (dismissed):
# Code Review — PR #6 (octo-admin)

## Summary

Single-line change: `pull_request` → `pull_request_target` in `.github/workflows/octo-pr-feed.yml`. The motivation (forks don't get repo secrets under `pull_request`, so `OCTO_BOT_TOKEN` is unavailable and every external PR fails the IM notification) is correct.

The risk surface here is the well-known `pull_request_target` privilege-escalation pattern — secrets become available to a workflow running in the base-branch context, so any execution of attacker-controlled PR code becomes a credential-exfiltration / repo-write primitive. I went through the workflow against that checklist.

## 1. Verification

| Check | Result | Evidence |
|---|---|---|
| No `actions/checkout` of PR head | ✅ | `.github/workflows/octo-pr-feed.yml` has no `checkout` step; reusable callee `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main` also has no checkout — only a single `Send to Octo` step running hardcoded inline Python. |
| No build/install of PR code (npm/pip/etc.) | ✅ | Reusable workflow runs only `python3 << PYEOF ... PYEOF` against `urllib`/stdlib. No `package.json`, `requirements.txt`, or any PR-supplied script is executed. |
| No script-injection sink for attacker-controlled fields | ✅ | `pr_title` / `pr_author` (attacker-controlled) flow `${{ github.event.pull_request.* }}` → reusable-workflow input → env var → `os.environ[...]` in Python. They are never substituted into a `run:` shell block, so the classic `$(...)`/backtick injection vector does not apply. |
| Reusable workflow pinned safely | ⚠️ Minor | Pinned to `@main` of `Mininglamp-OSS/.github`. Acceptable for a same-org reusable workflow, but pinning to a commit SHA would be stronger (defense in depth against a compromised org-level repo). Not blocking. |
| `GITHUB_TOKEN` blast radius | ⚠️ Minor | No `permissions:` block is declared, so the job inherits the repo's default `GITHUB_TOKEN` permissions. Since the workflow does not need to read or write *anything* in the repo (it only POSTs to `im.deepminer.com.cn`), an explicit `permissions: {}` at workflow or job level would shrink the blast radius if the IM endpoint or the python step were ever modified. Not blocking. |
| Trigger event types | ✅ | `opened, closed, reopened, ready_for_review, review_requested` — all pure-metadata events, no concern under `pull_request_target`. |

## 2. Issues

**None at P0 / P1.** The change is safe in this specific workflow because no PR-supplied code is ever fetched or executed, and attacker-controlled payload fields are kept off shell.

### P2 — Defense in depth (non-blocking)

1. **Add an explicit `permissions:` block.**
   The workflow has no need for any `GITHUB_TOKEN` scope. Recommend:
   ```yaml
   permissions: {}
   ```
   at the top level of `.github/workflows/octo-pr-feed.yml`. This is a one-line change that hard-caps the damage if the workflow is ever extended.

2. **Consider pinning the reusable workflow by SHA.**
   `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main` is fine for a same-org reusable workflow, but `@<commit-sha>` would protect against a tampering scenario in `.github`. Optional.

3. **Out-of-scope nit (PR description vs diff).**
   The PR body says "Applied to all 8 repos." This PR touches only `octo-admin`. Worth confirming the other 7 repos got the same patch in their own PRs — otherwise the fix is incomplete from a fleet perspective. Not actionable in this PR.

## 3. Recommendation

Safe to merge as-is. The two P2 items above are quality-of-defense improvements, not prerequisites.

yujiawei (approved):
# Code Review — PR #6 (octo-admin)

## Summary

Switches the `octo-pr-feed` workflow trigger from `pull_request` to `pull_request_target` so the workflow can read `OCTO_BOT_TOKEN` when triggered by PRs from forks. Also locks `permissions: {}` at the workflow level and pins the called reusable workflow to a commit SHA instead of `@main`.

**Verdict: APPROVED.** The change is correct, narrowly scoped, and includes appropriate hardening (SHA pin + empty permissions). The classic `pull_request_target` footguns do not apply here because no PR-controlled code is checked out or executed.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Trigger change addresses fork-secret problem | ✅ | `pull_request` from forks does not expose secrets; `pull_request_target` runs in base-branch context with secrets — standard GitHub Actions semantics. |
| No `actions/checkout` of PR head | ✅ | Caller workflow at `.github/workflows/octo-pr-feed.yml` (PR head `7cfa15e`) only forwards event-payload fields to the reusable workflow. No checkout step exists. |
| Reusable workflow does not execute PR-controlled code | ✅ | `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c5134` only forwards inputs into a Python heredoc via `env:` and `os.environ[...]` — no shell expansion of attacker strings, all outbound payloads built via `json.dumps`. |
| Reusable workflow pinned to immutable ref | ✅ | `@c6c51348fccac99161166dce1b52dec1063aeea2` (full 40-char commit SHA), not a tag or branch. |
| `permissions: {}` at workflow level | ✅ | Strips default `GITHUB_TOKEN` scopes; intersected with the reusable workflow's permissions (none requested), so the reusable runs with no repo-write power. Defense in depth even though the token is not used. |
| Pinned SHA exists & matches current main of reusable | ✅ | `c6c5134` is HEAD of `Mininglamp-OSS/.github` `main` at review time. |
| Event types unchanged | ✅ | `[opened, closed, reopened, ready_for_review, review_requested]` preserved. |

## 2. Findings

None at P0 or P1.

### P2 — advisory only

1. **Pin-update process is now manual across 8 repos.** Pinning `@c6c5134` is the right call, but every change to `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml` will need a coordinated bump in 8 caller repos. Consider:
   - A small note in `Mininglamp-OSS/.github/README.md` documenting the bump procedure, or
   - A scheduled job (or Dependabot config for `github-actions` ecosystem on each consumer repo) to surface available reusable-workflow updates.
   Dependabot does support pinned reusable-workflow refs and will open update PRs automatically — likely the lowest-effort option.

2. **Inline comment recommended.** A one-line comment in the YAML stating *why* `pull_request_target` is safe here (no PR checkout) helps future maintainers avoid adding a checkout step accidentally. Suggested:

   ```yaml
   # SAFETY: pull_request_target is safe because we never check out PR code.
   # If you add an `actions/checkout` step here, change the trigger back to pull_request.
   on:
     pull_request_target:
   ```

3. **Forwarded fields are attacker-controlled but inert.** `pr_title`, `pr_author`, `repo_name`, etc. originate from a fork and are passed to the reusable workflow. They flow through `env:` → `os.environ[...]` → `json.dumps(...)` → outbound HTTP. No shell expansion, no `${{ ... }}` interpolation in `run:` blocks, no logging-as-code. Confirmed safe at the current pin; this is a property to preserve in future changes to the reusable workflow.

4. **Closed events on merge.** With `pull_request_target` the `closed` event still fires on merge from the base context — this is the expected behavior and matches the `merged` flag used downstream. No change needed; calling out only because reviewers occasionally ask whether merge notifications still work after the trigger swap.

## 3. Recommendations

- Land as-is. The same change should be replicated identically across the other 7 repos to keep behavior consistent (PR description says this is already in flight).
- Consider adopting Dependabot `github-actions` updates on consumer repos so SHA bumps for the reusable workflow show up as PRs.
- Add the short safety comment in (2) above to make the trigger-vs-checkout invariant explicit.

## 4. Additional observations

- `permissions: {}` is set at the **workflow** level (above `jobs:`). For reusable-workflow calls, the effective permissions are the intersection of caller and callee — both are now zero, which is the strongest stance.
- The reusable workflow declares `secrets: OCTO_BOT_TOKEN: required: true`, and the caller passes it explicitly (no `secrets: inherit`). Good — least-privilege secret forwarding.
- No new files, no test infrastructure changes; review surface is a 4-line workflow diff. Risk is fully contained.

Jerry-Xin (approved):
This PR is in scope for `octo-admin` and safely fixes fork PR notification failures by moving the metadata-only notification workflow to `pull_request_target`.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion `.github/workflows/octo-pr-feed.yml:11`: Pinning the reusable workflow to a SHA is the right security posture for `pull_request_target`, but it creates an update process requirement. Consider documenting how this SHA should be refreshed when the org-level reusable workflow changes.

✅ Highlights:
- `.github/workflows/octo-pr-feed.yml:4`: `pull_request_target` is appropriate here because the workflow does not check out or execute PR code.
- `.github/workflows/octo-pr-feed.yml:7`: `permissions: {}` correctly minimizes `GITHUB_TOKEN` privileges.
- `.github/workflows/octo-pr-feed.yml:11`: Pinning the reusable workflow avoids executing a moving `main` reference with repository secrets.

PR #7, opened 2026-05-13
ci: add auto-add-to-project caller (delegates to org reusable)

lml2468:
## What

Deploys a lightweight caller for the org-level `auto-add-to-project` reusable workflow.

Every newly opened Issue or PR in this repo will be automatically added to [Octo Project Board #2](https://github.com/orgs/Mininglamp-OSS/projects/2).

## How it works

```
issues.opened / pull_request_target.opened
    → this caller
    → Mininglamp-OSS/.github / auto-add-to-project.yml (reusable)
    → actions/add-to-project@v1.0.2
    → Project Board #2
```

## Merge order

**Depends on** `Mininglamp-OSS/.github#4` (the reusable workflow) being merged to `main` first. This PR is safe to merge any time after that.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -1,0 +1,17 @@
+# Caller: delegates to the org-level reusable workflow.
+# Adds every newly opened Issue and PR to Octo Project Board #2.
+name: Auto-add to Octo Board
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+permissions: {}
+
+jobs:
+  add:
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}


Jerry-Xin (changes requested):
The PR is relevant to `Mininglamp-OSS/octo-admin`, but the delegated workflow target is not currently callable as a reusable workflow.

🔴 Blocking

🔴 Critical — `.github/workflows/auto-add-to-project.yml:15`

`jobs.add.uses` points to `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main`, so GitHub requires that target workflow to expose `on: workflow_call`. I checked the referenced workflow on `Mininglamp-OSS/.github@main`, and it is currently an event-triggered workflow with `on: issues` and `on: pull_request_target`, not a reusable workflow. As written, this caller will fail before running because the target is not a valid reusable workflow entry point.

The dependency PR needs to land with `workflow_call` support before this PR can be merged, or this caller should point to a ref that already contains the reusable workflow contract.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/auto-add-to-project.yml:15`

Consider pinning the reusable workflow to an immutable commit SHA, especially because `PROJECT_TOKEN` is passed to it. Using `@main` may be intentional for central org automation, but it means future changes in the org workflow immediately receive this repo secret.

✅ Highlights

The local trigger scope is narrow: only newly opened issues and PRs are handled.

`permissions: {}` is appropriate because the workflow does not need the caller’s `GITHUB_TOKEN`; it delegates with the explicit project token.

yujiawei (approved):
# Code Review — PR #7 (octo-admin)

## Summary

A 17-line caller workflow that delegates to the org-level reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` to auto-add newly opened issues and PRs to Project Board #2.

The diff is minimal, correctly scoped, and follows least-privilege principles. The implementation is sound.

## Verification

| Item | Status | Evidence |
|------|--------|----------|
| Triggers limited to `opened` only | ✅ | `.github/workflows/auto-add-to-project.yml:7,9` |
| Workflow-level `permissions: {}` (least privilege) | ✅ | `.github/workflows/auto-add-to-project.yml:11` |
| Explicit named secret pass-through (not `inherit`) | ✅ | `.github/workflows/auto-add-to-project.yml:15-16` |
| `pull_request_target` safe usage (no checkout, no untrusted code execution) | ✅ | Caller only delegates; reusable runs `actions/add-to-project` |
| Secret name matches reusable workflow contract (`PROJECT_TOKEN`) | ✅ | Matches the upstream reusable's `secrets:` declaration |

## Findings

### P1 — Cross-repo merge order (documented, but worth confirming before merge)

The reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` does **not** currently expose a `workflow_call` trigger on `main` — that conversion is in `Mininglamp-OSS/.github#4`, which is still **OPEN** at the time of this review. Until that PR lands on `main`, this caller will fail at workflow resolution time on every issue/PR open in this repo.

The PR body already calls this out clearly ("Depends on `Mininglamp-OSS/.github#4` (the reusable workflow) being merged to `main` first"), so this is a merge-ordering decision for the human merger, not a code defect. Recommend holding the merge button until `.github#4` is on `main`.

### P2 — `@main` reference instead of pinned SHA (low severity, accepted pattern)

The caller pins to `@main`:
```yaml
uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
```

For an org-internal reusable workflow this is the conventional pattern and gives you "automatic uptake" of fixes to the reusable. The trade-off is that any future regression on `.github`'s `main` immediately affects every caller. If you want stronger supply-chain isolation you could pin to a release tag or SHA, but for this use case (project-board automation, no code execution against repo contents) the current `@main` reference is acceptable.

### P2 — Workflow filename vs. PR-described "caller" naming (cosmetic)

The caller and the reusable share the same filename (`auto-add-to-project.yml`) and very similar `name:` field ("Auto-add to Octo Board"). This is fine, but if a future engineer is debugging from the Actions UI alone they may briefly wonder which workflow is firing. Consider renaming the caller's `name:` to something like `Auto-add to Octo Board (caller)` for log-disambiguation. Strictly cosmetic.

## Security Review

- `permissions: {}` at the workflow level revokes all default `GITHUB_TOKEN` permissions; the project-board API call is authenticated via the user-provided `PROJECT_TOKEN` PAT inside the reusable. This is the correct posture.
- `pull_request_target` is the right event here because adding fork PRs to a project board requires write-capable token context. No checkout of PR head SHA happens in either the caller or the reusable, so the well-known `pull_request_target` + checkout vulnerability class does not apply.
- Explicit secret pass-through (`secrets: PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}`) is preferable to `secrets: inherit` — keeps the trust surface explicit. ✅

## Recommendation

The change is a clean, correctly scoped CI delegation with sound security posture. No code changes requested.

**Merge gate:** confirm `Mininglamp-OSS/.github#4` is merged to `main` first; then this PR is safe to merge.

lml2468 (commented):
**CHANGES_REQUESTED**

CI is currently red, so I stopped at the required gate check and did not proceed to the diff or reusable-workflow verification.

Blocking items:
1. PR gate: `notify / notify` failed in workflow `Octo PR Feed` (run job: https://github.com/Mininglamp-OSS/octo-admin/actions/runs/25805616428/job/75807378426). Please fix or rerun the failing check before review continues.

Non-blocking notes:
- I did not verify whether `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` exposes `on: workflow_call` because the instructions require stopping on red CI.

Highlights:
- `build (18)` and `build (20)` are passing; only `notify / notify` is blocking the gate right now.

lml2468 (commented):
**CHANGES_REQUESTED** (submitted as comment due to GitHub self-review restriction)

The caller code itself is clean, but the upstream reusable workflow does not yet expose `on: workflow_call`, so this will break at dispatch if merged now.

**Blocking items:**

1. **Merge ordering dependency — `.github#4` must land first.** I independently verified: `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` currently only has `on: issues` + `on: pull_request_target` — no `workflow_call` trigger. GitHub requires `workflow_call` for any `jobs.<id>.uses` invocation, so this caller would fail silently at workflow dispatch. This confirms Jerry-Xin's finding. The dependency is documented in the PR description and `Mininglamp-OSS/.github#4` (state: OPEN) adds the missing `workflow_call` trigger. **Do not merge this until `.github#4` lands on main.**

**Non-blocking notes:**

- I verified the contract alignment between this caller and `.github#4`'s proposed reusable workflow: the caller passes `secrets: PROJECT_TOKEN`, and `.github#4` declares `workflow_call.secrets.PROJECT_TOKEN: required: true` — exact match.
- `permissions: {}` is correctly set at the workflow level. This is the right posture for `pull_request_target` — the caller needs zero `GITHUB_TOKEN` scope since `PROJECT_TOKEN` (a PAT) does the actual project board write.
- The `notify / notify` CI failure is from `Octo PR Feed` (the issue PR #6 addresses), not related to this PR's changes. The actual build CI (`build (18)`, `build (20)`) is green.
- Note: one existing review APPROVED this PR without flagging the missing `workflow_call` on the reusable workflow — that's a review quality gap worth noting.

**Highlights:**

- Well-structured 17-line caller with clear comment explaining the delegation pattern.
- `pull_request_target` + `permissions: {}` + no checkout is the correct security posture for this use case.

PR #8, opened 2026-05-14
ci: improve GitHub Actions workflows for octo-admin

lml2468:
### C: Add `workflow_id` to `octo-ci-status.yml`

Added `workflow_id: ${{ github.event.workflow_run.workflow_id }}` to the `with` block in `octo-ci-status.yml` so the CI status notification includes the workflow identifier.

### D: Add `permissions: {}` to `pr-merged-done.yml`

Adds explicit `permissions: {}` block between `on:` and `jobs:` to clarify that `GITHUB_TOKEN` is not used; only `PROJECT_TOKEN` is passed explicitly.

### DIFF
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -16,5 +16,6 @@ jobs:
       run_id: ${{ github.event.workflow_run.id }}
       run_url: ${{ github.event.workflow_run.html_url }}
       project_group_id: "a48e308c2ead4d9dbcdc30fff0f01eaf"
+      workflow_id: ${{ github.event.workflow_run.workflow_id }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/pr-merged-done.yml
+++ b/.github/workflows/pr-merged-done.yml
@@ -17,6 +17,8 @@ on:
   pull_request_target:
     types: [closed]
 
+permissions: {}  # GITHUB_TOKEN not used; PROJECT_TOKEN is passed explicitly
+
 jobs:
   move-to-done:
     if: github.event.pull_request.merged == true && github.event.pull_request.base.ref == github.event.repository.default_branch


yujiawei (approved):
# Code Review — PR #8 (octo-admin)

## Summary

Two small, well-scoped CI hygiene changes. Both verify cleanly against the upstream reusable workflows in `Mininglamp-OSS/.github`. No functional risk, no security concerns. Recommend merge.

## Verification

### Change C — `workflow_id` added to `octo-ci-status.yml`

- ✅ `workflow_id` is a declared optional input on the upstream reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@main` (typed `number`, default `0`). Passing it is the recommended path per the upstream input description.
- ✅ The expression `${{ github.event.workflow_run.workflow_id }}` is the correct path for a `workflow_run`-triggered caller — the `workflow_run` event payload exposes the numeric workflow id at that location. (Note: upstream's input `description` says `${{ github.workflow_run.workflow_id }}`, which is a typo on upstream's side; the value used here is correct.)
- ✅ Effect is exactly what the upstream description promises: scopes run-history lookups to this workflow rather than fetching globally for the repo, eliminating false-positive state transitions when other workflows on `main` change conclusion.

### Change D — `permissions: {}` added to `pr-merged-done.yml`

- ✅ Safe: the caller does not check out PR code and does not invoke any step that needs `GITHUB_TOKEN`. It only forwards `secrets.PROJECT_TOKEN` to the reusable workflow.
- ✅ Consistent: the upstream reusable workflow `Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml@main` itself declares `permissions: {}`, so dropping default permissions at the caller mirrors upstream and is the more defensive default for a `pull_request_target` workflow (which would otherwise inherit broad write scopes for repo contents/issues/PRs).
- ✅ Inline comment accurately explains the intent.

## Findings

None. No P0 / P1 / P2 issues identified.

## Suggestions (non-blocking, optional)

1. Pin the reusable-workflow refs. Both callers reference `...@main`. Pinning to a commit SHA (or a maintained tag) would make caller behavior reproducible and prevent silent semantics changes when the upstream workflow evolves. Out of scope for this PR.
2. Consider applying the same `permissions: {}` hardening to the other `workflow_run` / `pull_request_target`-style callers in `.github/workflows/` (e.g. `octo-ci-status.yml`, `octo-issue-feed.yml`, `octo-pr-feed.yml`) in a follow-up, to give every caller least-privilege defaults rather than relying on the org-wide default. Optional, separate PR.

## Verdict

Approve. Both changes are surgical, correct against the upstream contract, and improve operational signal-to-noise (Change C) and security posture (Change D).

Jerry-Xin (approved):
One-line summary: The PR is relevant to `octo-admin` CI/project automation and both workflow changes are small, correct, and consistent with the existing GitHub Actions setup.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion: [.github/workflows/octo-ci-status.yml](/private/tmp/pr-review-ZtHWQx/repo/.github/workflows/octo-ci-status.yml:11) still calls the reusable workflow via `@main`, while `octo-pr-feed.yml` pins to a commit. This is pre-existing, not introduced by this PR, but pinning would reduce supply-chain drift for automation that uses secrets.

✅ Highlights  
🔵 [.github/workflows/octo-ci-status.yml](/private/tmp/pr-review-ZtHWQx/repo/.github/workflows/octo-ci-status.yml:19): Passing `workflow_id` matches the reusable workflow contract and improves workflow-specific CI status handling.

🔵 [.github/workflows/pr-merged-done.yml](/private/tmp/pr-review-ZtHWQx/repo/.github/workflows/pr-merged-done.yml:20): `permissions: {}` is appropriate here because the workflow does not rely on `GITHUB_TOKEN`; it delegates to the reusable workflow with the explicit `PROJECT_TOKEN` secret.

No test changes are needed for this config-only PR.

lml2468 (commented):
This PR makes two narrow GitHub Actions changes; CI, mergeability, head-branch workflow files, and reusable workflow contracts all check out. I would approve, but GitHub does not allow approving my own pull request.

🔴 Blocking (if any):
None.

💬 Non-blocking (if any):
None. Existing reviewers already covered the optional reusable-workflow ref pinning follow-up, so I am not duplicating it.

PR #15, opened 2026-05-16
feat(users): push type filter to backend (bot_only/system_only/exclude_*)

an9xyz:
## Summary

Backend \`GET /v1/manager/user/list\` has been extended with new response fields (\`is_bot\`, \`is_system\`) and new query parameters. This PR aligns the frontend.

### Backend contract (confirmed)

\`\`\`
page_index      int  required  page index
page_size       int  required  page size
keyword         str  optional  fuzzy search on name/UID/phone/email/short_no
online          int  optional  -1=all 0=offline 1=online
exclude_bot     int  optional  1 = exclude user.robot=1
exclude_system  int  optional  1 = exclude built-in system accounts
bot_only        int  optional  1 = only user.robot=1
system_only     int  optional  1 = only built-in system accounts
\`\`\`

### Users page (User Management)
- Type filter is fully pushed to the backend, so pagination reflects the filtered result set:
  - Real users → \`exclude_bot=1 & exclude_system=1\`
  - Bot accounts → \`bot_only=1\`
  - System accounts → \`system_only=1\` (new option)
- Badge rendering uses the new \`is_bot\` / \`is_system\` fields. "System" takes priority over "Bot".
- Drop the \`uid.endsWith('_bot')\` heuristic.
- \`fetchData\` now takes explicit args (\`page\`, \`pageSize\`, \`type\`, \`keyword\`) to avoid stale-closure bugs when \`typeFilter\` changes before Search is pressed.

### Spaces page (Create Space → Owner picker)
- Request now includes \`exclude_bot=1 & exclude_system=1\` so the backend filters out non-human accounts.
- Client-side filter reduced to \`status === 1 && is_destroy !== 1\`. The \`_bot\` suffix hack is gone.

## Test plan

- [ ] User Management → "All types": pagination and total work as before
- [ ] User Management → "Real users": no BOT / system badges in the list
- [ ] User Management → "Bot accounts": every row shows a BOT badge; \`total\` is the bot count
- [ ] User Management → "System accounts": every row shows a system badge
- [ ] User Management → switching type resets to page 1 and refetches
- [ ] User Management → typing a keyword without pressing Search, then switching type, sends both the new type and the typed keyword
- [ ] Spaces → Create Space → Owner picker: bot and system accounts are absent from results

### DIFF
--- a/src/pages/Spaces/index.tsx
+++ b/src/pages/Spaces/index.tsx
@@ -86,14 +86,19 @@ export default function Spaces() {
     const seq = ++userSearchSeq.current
     setUserSearching(true)
     try {
-      const params = new URLSearchParams({ page_index: '1', page_size: '20' })
+      const params = new URLSearchParams({
+        page_index: '1',
+        page_size: '20',
+        exclude_bot: '1',
+        exclude_system: '1',
+      })
       if (kw) params.append('keyword', kw)
       const res = await api.get<{ list: UserOptionRaw[] }>(
         `/v1/manager/user/list?${params}`,
       )
       if (seq === userSearchSeq.current) {
         const list = (res.data.list || []).filter(
-          (u) => u.status === 1 && u.is_destroy !== 1 && !u.uid?.endsWith('_bot'),
+          (u) => u.status === 1 && u.is_destroy !== 1,
         )
         setUserOptions(list)
       }
--- a/src/pages/Users/index.tsx
+++ b/src/pages/Users/index.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useMemo, useState } from 'react'
 import { Table, Input, Button, Select, message, Modal, Tooltip, Typography } from 'antd'
-import { SearchOutlined, ReloadOutlined, RobotOutlined } from '@ant-design/icons'
+import { SearchOutlined, ReloadOutlined, RobotOutlined, SettingOutlined } from '@ant-design/icons'
 import type { ColumnsType } from 'antd/es/table'
 import api from '../../api'
 
@@ -15,8 +15,13 @@ interface User {
   status: number
   online: number
   is_destroy: number
+  is_bot?: number
+  is_system?: number
   register_time: string
 }
+
+type TypeFilter = 'all' | 'human' | 'bot' | 'system'
+type StatusFilter = 'all' | 'online' | 'offline' | 'banned' | 'destroyed'
 
 export default function Users() {
   const [loading, setLoading] = useState(false)
@@ -25,17 +30,31 @@ export default function Users() {
   const [page, setPage] = useState(1)
   const [keyword, setKeyword] = useState('')
   const [pageSize, setPageSize] = useState(20)
-  const [typeFilter, setTypeFilter] = useState<'all' | 'human' | 'bot'>('all')
-  const [statusFilter, setStatusFilter] = useState<'all' | 'online' | 'offline' | 'banned' | 'destroyed'>('all')
+  const [typeFilter, setTypeFilter] = useState<TypeFilter>('all')
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>('all')
 
-  const fetchData = async () => {
+  const fetchData = async (
+    nextPage = page,
+    nextPageSize = pageSize,
+    nextType = typeFilter,
+    kw = keyword,
+  ) => {
     setLoading(true)
     try {
       const params = new URLSearchParams({
-        page_index: page.toString(),
-        page_size: pageSize.toString(),
+        page_index: nextPage.toString(),
+        page_size: nextPageSize.toString(),
       })
-      if (keyword) params.append('keyword', keyword)
+      if (kw) params.append('keyword', kw)
+      // 类型筛选全部下推到后端，分页基于过滤后结果
+      if (nextType === 'human') {
+        params.append('exclude_bot', '1')
+        params.append('exclude_system', '1')
+      } else if (nextType === 'bot') {
+        params.append('is_bot', '1')
+      } else if (nextType === 'system') {
+        params.append('is_system', '1')
+      }
       const res = await api.get(`/v1/manager/user/list?${params}`)
       setData(res.data.list || [])
       setTotal(res.data.count || 0)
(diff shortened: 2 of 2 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `octo-admin`, but the users table can show stale backend-filtered results after quick filter/page/search changes.

🔴 Blocking

🟡 Warning — `src/pages/Users/index.tsx:58`: `fetchData` applies every response unconditionally. Since type filtering is now pushed to the backend and `useEffect` triggers new requests on `typeFilter` changes, a slower earlier request can overwrite a newer one. Example: switch from human to bot quickly; if the human request resolves last, the UI can display human rows and totals while the selected filter is bot. Add request sequencing or cancellation, similar to `userSearchSeq` in `src/pages/Spaces/index.tsx:86`, and only apply the latest response, including `loading` state.

💬 Non-blocking

🔵 Suggestion — `src/pages/Users/index.tsx:50`: If backend flags are not guaranteed to be mutually exclusive, the bot filter may include system accounts because it sends only `is_bot=1`, while rendering gives system badge priority. If system accounts can also be bots, consider adding `exclude_system=1` for the bot filter or documenting the backend invariant.

🔵 Suggestion — `src/pages/Users/index.tsx:100`: `handleBan` still has no local error handling inside `onOk`; failed ban/unban requests may surface through the modal promise behavior rather than the page’s message pattern. This is pre-existing, but worth aligning with other handlers.

✅ Highlights

The PR removes the `_bot` UID heuristic and uses explicit backend fields, which is a clear improvement. The Spaces owner selector already has request sequencing, and adding `exclude_bot` / `exclude_system` there matches the intended owner eligibility behavior.

Verification note: `npm run build` could not run in this checkout because dependencies are not installed (`tsc` was not found).

Jerry-Xin (changes requested):
One-line summary: The PR is relevant to octo-admin, but the Users page needs stale-request protection before backend-only type filtering is safe.

🔴 Blocking

- 🟡 Warning — `src/pages/Users/index.tsx:58-64`, `src/pages/Users/index.tsx:68-82`: `fetchData` applies every response to `data`, `total`, and `loading` with no request ordering or cancellation. Because this PR removes the client-side type filter from `filteredData`, an older request can now overwrite a newer filtered request. Example: initial “all” request is slow, user switches to “bot”, the bot request returns first, then the older all-users response returns and the table shows non-bot users while the UI still says bot. `total` can also become incorrect. Add a latest-request guard or `AbortController`, similar to the `userSearchSeq` pattern already used in `src/pages/Spaces/index.tsx:85-113`, and only update rows/count/loading for the latest Users request.

💬 Non-blocking

- 🔵 Suggestion — There is no automated test coverage in this repo for this flow. Given the async behavior here, the manual test plan should include rapid filter switching and rapid search/filter changes to confirm stale responses cannot replace the currently selected result set.

✅ Highlights

- The PR passes the project relevance gate: both changes are within octo-admin’s user management and space creation owner-selection flows.
- `src/pages/Spaces/index.tsx:89-103` keeps the existing request-sequence guard while adding backend human-user filtering, which is the right pattern for this kind of async picker.
- `src/pages/Users/index.tsx:36-57` correctly maps the new type filters to the backend query parameters, and `src/pages/Users/index.tsx:117-125` gives system accounts priority over bot badges.

Verification note: I attempted `npm run build`, but this checkout does not have local dependencies installed, so `tsc` was not available.

an9xyz:
Superseded by #16 — this PR was auto-closed when the source branch was renamed from `understand-user-list-apis` to `feat/users-type-filter`.

lml2468 (changes requested):
## Code Review — PR #15 (octo-admin) — 📖 齐静春

**Verdict: CHANGES_REQUESTED** — The "Bot" and "System" positive filters are silently broken due to an API contract mismatch with the backend, and the race condition flagged by Jerry-Xin is a real concern.

### 🔴 Blocking

**P0 — `src/pages/Users/index.tsx:52-55`: `is_bot=1` and `is_system=1` query params are silently ignored by the backend — "Bot 账号" and "系统账号" filters return ALL users**

The frontend sends:
```typescript
} else if (nextType === 'bot') {
    params.append('is_bot', '1')     // ← backend ignores this
} else if (nextType === 'system') {
    params.append('is_system', '1')  // ← backend ignores this
}
```

I traced the backend handler in `octo-server/modules/user/api_manager.go:560-563`:
```go
filter := userListFilter{
    OnlineStatus:  int(online),
    ExcludeBot:    c.Query("exclude_bot") == "1",
    ExcludeSystem: c.Query("exclude_system") == "1",
}
```

And the filter struct in `db_manager.go:17-21`:
```go
type userListFilter struct {
    OnlineStatus  int
    ExcludeBot    bool   // exclude only
    ExcludeSystem bool   // exclude only
}
```

The backend only supports **negative** filters (`exclude_bot`, `exclude_system`). There are no `is_bot` or `is_system` query params — they are **response fields only** (see `swagger/api.yaml:2459-2464`). Unknown query params are silently ignored by the Go router.

**Impact**: Selecting "Bot 账号" or "系统账号" type filter returns the full unfiltered user list. The admin sees all users while the UI says "Bot 账号" — functionally broken and misleading.

**Fix options**:
1. Add `is_bot` / `is_system` positive filter support in the backend (preferred — keeps filtering server-side with correct pagination)
2. Keep using `exclude_*` params and do final positive filtering client-side (quick fix but breaks pagination — `total` count won't match displayed rows)

---

**P1 — `src/pages/Users/index.tsx:39-63`: No stale-request guard in `fetchData` — confirming Jerry-Xin's finding with additional evidence**

I independently confirm this is a real issue. Supporting evidence: `src/pages/Spaces/index.tsx:85-113` already implements the correct pattern with `userSearchSeq` ref:
```typescript
const userSearchSeq = useRef(0)
const searchUsers = async (kw: string) => {
    const seq = ++userSearchSeq.current
    // ... fetch ...
    if (seq === userSearchSeq.current) {
        setUserOptions(list)  // only update if still the latest request
    }
}
```

The Users page has no equivalent guard. When `typeFilter` changes trigger the `useEffect`, a slow previous request can overwrite the correct filtered result. Pre-PR this was masked by client-side filtering in `filteredData`; post-PR the client-side type filter is removed, so stale backend responses become directly visible.

### 💬 Non-blocking

- The `Spaces/index.tsx` change (pushing `exclude_bot=1&exclude_system=1` into the query params and removing the `uid.endsWith('_bot')` client-side filter) is correct and well-done.
- The tag rendering change (checking `is_system` / `is_bot` response fields instead of UID suffix heuristic) is a clear improvement.

### ✅ Highlights

- Good direction: moving from UID-suffix heuristic to backend-provided `is_bot`/`is_system` flags is the right architectural call — eliminates spoofing risk.
- `handleTypeFilterChange` correctly resets page to 1 on filter change.
- Type extraction (`TypeFilter`, `StatusFilter`) improves readability.

PR #16, opened 2026-05-16
feat(users): push type filter to backend (bot_only/system_only/exclude_*)

an9xyz:
> Supersedes #15 (closed automatically when the source branch was renamed from `understand-user-list-apis` to `feat/users-type-filter`).

**Depends on Mininglamp-OSS/octo-server#63** — ✅ merged on 2026-05-16 (commit `c2b549f`). Frontend can ship after the backend release reaches the target environment.

## Summary

Backend \`GET /v1/manager/user/list\` has been extended with new response fields (\`is_bot\`, \`is_system\`) and new query parameters. This PR aligns the frontend.

### Backend contract (confirmed)

\`\`\`
page_index      int  required  page index
page_size       int  required  page size
keyword         str  optional  fuzzy search on name/UID/phone/email/short_no
online          int  optional  -1=all 0=offline 1=online
exclude_bot     int  optional  1 = exclude user.robot=1
exclude_system  int  optional  1 = exclude built-in system accounts
bot_only        int  optional  1 = only user.robot=1                    (introduced in octo-server#63)
system_only     int  optional  1 = only built-in system accounts        (introduced in octo-server#63)
\`\`\`

### Users page (User Management)
- Type filter is fully pushed to the backend, so pagination reflects the filtered result set:
  - Real users → \`exclude_bot=1 & exclude_system=1\`
  - Bot accounts → \`bot_only=1\`
  - System accounts → \`system_only=1\` (new option)
- Badge rendering uses the new \`is_bot\` / \`is_system\` fields. "System" takes priority over "Bot".
- Drop the \`uid.endsWith('_bot')\` heuristic.
- \`fetchData\` now takes explicit args (\`page\`, \`pageSize\`, \`type\`, \`keyword\`) to avoid stale-closure bugs when \`typeFilter\` changes before Search is pressed.
- \`fetchData\` uses a \`useRef\` sequence guard so out-of-order responses from rapid filter switching cannot overwrite the table (mirrors the pattern already in Spaces). _Addresses review feedback from Jerry-Xin / yujiawei / lml2468._

### Spaces page (Create Space → Owner picker)
- Request now includes \`exclude_bot=1 & exclude_system=1\` so the backend filters out non-human accounts.
- Client-side filter reduced to \`status === 1 && is_destroy !== 1\`. The \`_bot\` suffix hack is gone.

## Release ordering

Backend must be deployed first. Until octo-server#63 is live in the target environment, \`bot_only=1\` / \`system_only=1\` will be silently ignored and the type filter will return unfiltered results.

## Test plan

- [ ] User Management → "All types": pagination and total work as before
- [ ] User Management → "Real users": no BOT / system badges in the list
- [ ] User Management → "Bot accounts": every row shows a BOT badge; \`total\` is the bot count
- [ ] User Management → "System accounts": every row shows a system badge
- [ ] User Management → switching type resets to page 1 and refetches
- [ ] User Management → typing a keyword without pressing Search, then switching type, sends both the new type and the typed keyword
- [ ] User Management → rapid filter toggling (bot → system → human within ~1s) leaves the table consistent with the final selection
- [ ] Spaces → Create Space → Owner picker: bot and system accounts are absent from results

### DIFF
--- a/src/pages/Spaces/index.tsx
+++ b/src/pages/Spaces/index.tsx
@@ -86,14 +86,19 @@ export default function Spaces() {
     const seq = ++userSearchSeq.current
     setUserSearching(true)
     try {
-      const params = new URLSearchParams({ page_index: '1', page_size: '20' })
+      const params = new URLSearchParams({
+        page_index: '1',
+        page_size: '20',
+        exclude_bot: '1',
+        exclude_system: '1',
+      })
       if (kw) params.append('keyword', kw)
       const res = await api.get<{ list: UserOptionRaw[] }>(
         `/v1/manager/user/list?${params}`,
       )
       if (seq === userSearchSeq.current) {
         const list = (res.data.list || []).filter(
-          (u) => u.status === 1 && u.is_destroy !== 1 && !u.uid?.endsWith('_bot'),
+          (u) => u.status === 1 && u.is_destroy !== 1,
         )
         setUserOptions(list)
       }
--- a/src/pages/Users/index.tsx
+++ b/src/pages/Users/index.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useMemo, useState } from 'react'
 import { Table, Input, Button, Select, message, Modal, Tooltip, Typography } from 'antd'
-import { SearchOutlined, ReloadOutlined, RobotOutlined } from '@ant-design/icons'
+import { SearchOutlined, ReloadOutlined, RobotOutlined, SettingOutlined } from '@ant-design/icons'
 import type { ColumnsType } from 'antd/es/table'
 import api from '../../api'
 
@@ -15,8 +15,13 @@ interface User {
   status: number
   online: number
   is_destroy: number
+  is_bot?: number
+  is_system?: number
   register_time: string
 }
+
+type TypeFilter = 'all' | 'human' | 'bot' | 'system'
+type StatusFilter = 'all' | 'online' | 'offline' | 'banned' | 'destroyed'
 
 export default function Users() {
   const [loading, setLoading] = useState(false)
@@ -25,17 +30,31 @@ export default function Users() {
   const [page, setPage] = useState(1)
   const [keyword, setKeyword] = useState('')
   const [pageSize, setPageSize] = useState(20)
-  const [typeFilter, setTypeFilter] = useState<'all' | 'human' | 'bot'>('all')
-  const [statusFilter, setStatusFilter] = useState<'all' | 'online' | 'offline' | 'banned' | 'destroyed'>('all')
+  const [typeFilter, setTypeFilter] = useState<TypeFilter>('all')
+  const [statusFilter, setStatusFilter] = useState<StatusFilter>('all')
 
-  const fetchData = async () => {
+  const fetchData = async (
+    nextPage = page,
+    nextPageSize = pageSize,
+    nextType = typeFilter,
+    kw = keyword,
+  ) => {
     setLoading(true)
     try {
       const params = new URLSearchParams({
-        page_index: page.toString(),
-        page_size: pageSize.toString(),
+        page_index: nextPage.toString(),
+        page_size: nextPageSize.toString(),
       })
-      if (keyword) params.append('keyword', keyword)
+      if (kw) params.append('keyword', kw)
+      // 类型筛选全部下推到后端，分页基于过滤后结果
+      if (nextType === 'human') {
+        params.append('exclude_bot', '1')
+        params.append('exclude_system', '1')
+      } else if (nextType === 'bot') {
+        params.append('is_bot', '1')
+      } else if (nextType === 'system') {
+        params.append('is_system', '1')
+      }
       const res = await api.get(`/v1/manager/user/list?${params}`)
       setData(res.data.list || [])
       setTotal(res.data.count || 0)
@@ -47,24 +66,32 @@ export default function Users() {
   }
 
   useEffect(() => {
-    fetchData()
-  }, [page, pageSize])
+    fetchData(page, pageSize, typeFilter, keyword)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [page, pageSize, typeFilter])
 
+  // 类型筛选已下推到后端；状态筛选仍是当前页客户端过滤（后端尚未支持）
   const filteredData = useMemo(() => {
     return data.filter((u) => {
-      if (typeFilter === 'bot' && !u.uid?.endsWith('_bot')) return false
-      if (typeFilter === 'human' && u.uid?.endsWith('_bot')) return false
       if (statusFilter === 'destroyed' && u.is_destroy !== 1) return false
       if (statusFilter === 'banned' && u.status !== 0) return false
       if (statusFilter === 'online' && !(u.online === 1 && u.status === 1 && u.is_destroy !== 1)) return false
       if (statusFilter === 'offline' && !(u.online !== 1 && u.status === 1 && u.is_destroy !== 1)) return false
       return true
     })
-  }, [data, typeFilter, statusFilter])
+  }, [data, statusFilter])
 
   const handleSearch = () => {
-    setPage(1)
-    fetchData()
+    if (page !== 1) {
+      setPage(1)
+    } else {
+      fetchData(1, pageSize, typeFilter, keyword)
+    }
+  }
+
+  const handleTypeFilterChange = (v: TypeFilter) => {
+    setTypeFilter(v)
+    if (page !== 1) setPage(1)
   }
 
   const handleBan = async (uid: string, status: number) => {
@@ -73,12 +100,10 @@ export default function Users() {
       onOk: async () => {
         await api.put(`/v1/manager/user/liftban/${uid}/${status}`)
         message.success(status === 0 ? '已封禁' : '已解封')
-        fetchData()
+        fetchData(page, pageSize, typeFilter, keyword)
       },
     })
   }
-
-  const isBot = (uid: string) => uid?.endsWith('_bot')
 
   const columns: ColumnsType<User> = [
     {
@@ -89,11 +114,15 @@ export default function Users() {
       render: (name, record) => (
         <span style={{ display: 'inline-flex', alignItems: 'center', gap: 6 }}>
           <span className="cell-primary">{name}</span>
-          {isBot(record.uid) && typeFilter !== 'bot' && (
+          {record.is_system === 1 ? (
+            <span className="ai-tag" aria-label="系统账号">
+              <SettingOutlined /> 系统
+            </span>
+          ) : record.is_bot === 1 ? (
             <span className="ai-tag" aria-label="Bot 账号">
               <RobotOutlined /> BOT
             </span>
-          )}
+          ) : null}
         </span>
       ),
     },
@@ -183,12 +212,13 @@ export default function Users() {
         />
         <Select
           value={typeFilter}
-          onChange={setTypeFilter}
-          style={{ width: 120 }}
+          onChange={handleTypeFilterChange}
+          style={{ width: 130 }}
           options={[
             { value: 'all', label: '全部类型' },
             { value: 'human', label: '真实用户' },
             { value: 'bot', label: 'Bot 账号' },
+            { value: 'system', label: '系统账号' },
           ]}
         />
         <Select
@@ -206,7 +236,7 @@ export default function Users() {
         <Button type="primary" icon={<SearchOutlined />} onClick={handleSearch}>
           搜索
         </Button>
-        <Button icon={<ReloadOutlined />} onClick={fetchData}>
+        <Button icon={<ReloadOutlined />} onClick={() => fetchData()}>
           刷新
         </Button>
       </div>
--- a/src/pages/Users/index.tsx
+++ b/src/pages/Users/index.tsx
@@ -51,9 +51,9 @@ export default function Users() {
         params.append('exclude_bot', '1')
         params.append('exclude_system', '1')
       } else if (nextType === 'bot') {
-        params.append('is_bot', '1')
+        params.append('bot_only', '1')
       } else if (nextType === 'system') {
-        params.append('is_system', '1')
+        params.append('system_only', '1')
       }
       const res = await api.get(`/v1/manager/user/list?${params}`)
       setData(res.data.list || [])
--- a/src/pages/Users/index.tsx
+++ b/src/pages/Users/index.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react'
+import { useEffect, useMemo, useRef, useState } from 'react'
 import { Table, Input, Button, Select, message, Modal, Tooltip, Typography } from 'antd'
 import { SearchOutlined, ReloadOutlined, RobotOutlined, SettingOutlined } from '@ant-design/icons'
 import type { ColumnsType } from 'antd/es/table'
@@ -32,6 +32,8 @@ export default function Users() {
   const [pageSize, setPageSize] = useState(20)
   const [typeFilter, setTypeFilter] = useState<TypeFilter>('all')
   const [statusFilter, setStatusFilter] = useState<StatusFilter>('all')
+  // 序号守卫：快速切换筛选/分页时，仅采用最近一次请求的响应，避免错序覆盖
+  const fetchSeq = useRef(0)
 
   const fetchData = async (
     nextPage = page,
@@ -39,6 +41,7 @@ export default function Users() {
     nextType = typeFilter,
     kw = keyword,
   ) => {
+    const seq = ++fetchSeq.current
     setLoading(true)
     try {
       const params = new URLSearchParams({
@@ -56,12 +59,17 @@ export default function Users() {
         params.append('system_only', '1')
       }
       const res = await api.get(`/v1/manager/user/list?${params}`)
+      if (seq !== fetchSeq.current) return
       setData(res.data.list || [])
       setTotal(res.data.count || 0)
     } catch (error) {
-      message.error((error as Error).message)
+      if (seq === fetchSeq.current) {
+        message.error((error as Error).message)
+      }
     } finally {
-      setLoading(false)
+      if (seq === fetchSeq.current) {
+        setLoading(false)
+      }
     }
   }
 


Jerry-Xin (changes requested):
The PR is relevant to `octo-admin`, but the Users page can show stale results for the selected backend type filter.

**🔴 Blocking**
- 🟡 Warning: `src/pages/Users/index.tsx:58-60` applies every completed `fetchData` response, even if a newer filter/page/search request has already been sent. Since this PR removes the client-side type filter and relies on backend `bot_only` / `system_only` / `exclude_*`, an older slower response can overwrite the current table and `total`. Example: switch from Bot to System quickly; if the Bot request resolves last, the UI can show Bot rows while the selector says System. Add the same kind of request sequence guard used in `src/pages/Spaces/index.tsx:85-113`, or cancel prior requests.

**💬 Non-blocking**
- 🔵 Suggestion: `src/pages/Users/index.tsx:117-121` compares `is_system` and `is_bot` with `=== 1`. If the backend contract ever serializes these as booleans, badges silently disappear. If the API is guaranteed to return numeric flags, this is fine; otherwise normalize or use a helper.
- 🔵 Suggestion: There are no tests in this repo, but this change would benefit from at least a focused component/request test if test infrastructure is added later, especially for request params and stale response ordering.
- Build verification could not run in this checkout because dependencies are not installed: `npm run build` failed with `tsc: command not found`.

**✅ Highlights**
- `src/pages/Spaces/index.tsx:89-103` correctly pushes bot/system exclusion to the backend while preserving the existing sequence guard for async search.
- `src/pages/Users/index.tsx:44-57` maps the new type filters to the documented backend parameters clearly, and system badge priority over bot is implemented as described.

yujiawei (commented):
# Code Review — PR #16 (octo-admin)

## Summary

Frontend change that aligns `/v1/manager/user/list` consumers with the new backend contract (`is_bot` / `is_system` response fields, `bot_only` / `system_only` / `exclude_bot` / `exclude_system` query params). The `uid.endsWith('_bot')` heuristic is fully removed from `src/pages/Users/index.tsx` and `src/pages/Spaces/index.tsx` (verified — no remaining occurrences in `src/`). Two callsites touched, +59 / -24, well-scoped.

The refactor of `fetchData` to take explicit arguments (`nextPage`, `nextPageSize`, `nextType`, `kw`) is the right move — it removes the stale-closure foot-gun where the previous `fetchData()` would observe whichever `typeFilter` / `keyword` happened to be captured at definition time.

Verdict: **COMMENTED**. No blocking issues. One concern worth addressing and a couple of nits/notes below.

---

## Findings

### P1 — Out-of-order responses on rapid type-filter switching (`src/pages/Users/index.tsx`)

`searchUsers` in `src/pages/Spaces/index.tsx:85-114` already guards against out-of-order responses with a sequence ref:

```ts
const seq = ++userSearchSeq.current
...
if (seq === userSearchSeq.current) { setUserOptions(list) }
```

`fetchData` in `src/pages/Users/index.tsx:36-66` has no equivalent guard. If a user clicks `bot → system → human` faster than the network responds (or `pageSize` switches `20 → 100`), the in-flight responses can land in any order and `setData` / `setTotal` will reflect whichever resolves last — not whichever was requested last. The user then sees a list that doesn't match the currently-selected filter, with `total` to match.

Pre-existing? Partially — the original code only refetched on `[page, pageSize]` changes, which users rarely toggle as quickly as a filter dropdown. Adding `typeFilter` to the trigger surface (correctly!) makes this race noticeably easier to hit.

Suggested fix: same `useRef` + sequence pattern already in `Spaces/index.tsx`, or `AbortController` on the underlying `api.get`.

### P2 — Backend deployment ordering / graceful degradation

The PR removes the `uid.endsWith('_bot')` heuristic and relies entirely on backend-provided `is_bot` / `is_system`. Until the backend rollout completes:

- Badge column in `Users/index.tsx:117-125` will render nothing for bots/system accounts (both fields absent → `record.is_system === 1` and `record.is_bot === 1` both false).
- "Real users" / "Bot accounts" / "System accounts" filters silently fall through on the backend (unknown params ignored) and return the unfiltered list — there is no client indication that the filter didn't apply.
- The Spaces Owner picker (`Spaces/index.tsx:89-104`) will now include bot/system accounts in the dropdown, since the client-side `_bot` suffix fallback is also gone.

Not a code defect, but worth coordinating release order (backend → frontend) and calling out in the deploy notes. If backend rollout has any chance of trailing the frontend, consider keeping the `_bot` suffix as a fallback in the Spaces picker for one release cycle.

### P2 — `total` vs displayed rows when the (client-side) status filter narrows the page

Pre-existing, but the PR has shifted the boundary so this is worth re-stating: type filter now goes to the backend (so `total` is the type-filtered count), but status filter (`online` / `offline` / `banned` / `destroyed`) is still client-side over `data` (current page only). Result: the table footer says `共 N 条` while the visible rows are a per-page subset of that N. Pagination across pages also "loses" status-filtered rows whenever the page rolls.

This is not part of this PR's scope, but as soon as backend supports a `status=` param the status filter should be pushed down too, mirroring this PR's pattern. Worth a follow-up issue.

### P3 — `handleSearch` branch when already on page 1

```ts
const handleSearch = () => {
  if (page !== 1) {
    setPage(1)            // useEffect refetches via [page,pageSize,typeFilter] dep
  } else {
    fetchData(1, pageSize, typeFilter, keyword)
  }
}
```

Works, but the asymmetry (one path relies on the `useEffect`, the other calls `fetchData` directly) is a small footgun for the next reader — if the `useEffect` deps ever change, only one branch will follow. Either call `fetchData(1, pageSize, typeFilter, keyword)` unconditionally and accept the rare double-fetch when `page !== 1`, or have both branches go through the same code path. Minor.

### P3 — Badge precedence

`record.is_system === 1 ? … : record.is_bot === 1 ? …` — system wins when both are 1. This matches the PR description ("System takes priority over Bot") and is fine; just noting it's a deliberate semantic choice that the backend should be aware of (a bot owned by the system shows as "系统", not "BOT").

### Nit — `useEffect` exhaustive-deps suppression

```ts
useEffect(() => {
  fetchData(page, pageSize, typeFilter, keyword)
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [page, pageSize, typeFilter])
```

The suppression is correct (you don't want a refetch on every keystroke in `keyword`), but `fetchData` itself is also missing from deps. Since it's a fresh closure on every render that's harmless here, but if anything ever moves it into a `useCallback` the suppression would mask a real bug. A short comment explaining *why* `keyword` is intentionally excluded would help the next reader. Optional.

### Nit — `Spaces/index.tsx` client-side filter no longer guards against bots

```ts
const list = (res.data.list || []).filter(
  (u) => u.status === 1 && u.is_destroy !== 1,
)
```

Now correctly trusts the server-side `exclude_bot=1 & exclude_system=1`. The client filter could check `is_bot !== 1 && is_system !== 1` as belt-and-suspenders, but only if you want the picker to keep working when a stale backend ignores the query params. Judgment call — flagged in P2 above.

---

## What I checked

- ✅ Every `uid.endsWith('_bot')` / `isBot(` reference is gone (`grep -r` over `src/`).
- ✅ Only `Users/index.tsx` and `Spaces/index.tsx` consume `is_bot` / `is_system` / the new query params.
- ✅ `fetchData` refactor preserves caller behavior (refresh button → `fetchData()` with all defaults; ban/unban → explicit args).
- ✅ Type definitions added to `User` interface are optional, so no compile break against existing API shapes.
- ✅ New "System accounts" option added to the type filter dropdown with matching backend param.
- ✅ Badge a11y labels present (`aria-label="系统账号"` / `"Bot 账号"`).
- ⚠️ No race-guard on `Users/fetchData` (see P1).
- ⚠️ Backend dependency for graceful degradation (see P2).

lml2468 (changes requested):
## Code Review — PR #16 (octo-admin) — 📖 齐静春

**Verdict: CHANGES_REQUESTED** — The P0 API contract mismatch from PR #15 is NOT fixed — it's the same bug with different param names. The backend PR that adds `bot_only`/`system_only` support has not been merged yet.

### 🔴 Blocking

**P0 — `src/pages/Users/index.tsx:52-55`: `bot_only=1` and `system_only=1` are NOT supported by the deployed backend — octo-server PR #63 is still OPEN**

The frontend now sends:
```typescript
} else if (nextType === 'bot') {
    params.append('bot_only', '1')     // ← backend doesn't recognize this yet
} else if (nextType === 'system') {
    params.append('system_only', '1')  // ← backend doesn't recognize this yet
}
```

I verified the current deployed backend state:

1. `octo-server/modules/user/api_manager.go:562-563` — still only parses `exclude_bot` and `exclude_system`:
   ```go
   ExcludeBot:    c.Query("exclude_bot") == "1",
   ExcludeSystem: c.Query("exclude_system") == "1",
   ```

2. `octo-server/modules/user/db_manager.go:17-21` — `userListFilter` struct still has only `ExcludeBot`/`ExcludeSystem` fields.

3. `octo-server` PR #63 ("feat(user): add bot_only/system_only inclusive filters") exists and adds the backend support, but its state is **OPEN** (not merged).

**Impact**: Until PR #63 merges, this frontend PR has the exact same bug as PR #15 — `bot_only=1` and `system_only=1` are silently ignored, "Bot 账号" and "系统账号" filters return ALL users.

**Recommendation**: Either:
1. Merge backend PR #63 first, then merge this frontend PR
2. Add explicit dependency documentation: "Depends on Mininglamp-OSS/octo-server#63" in the PR body, and do NOT merge until #63 ships

---

**P1 — `src/pages/Users/index.tsx:39-63`: Race condition still not addressed (same as PR #15)**

`fetchData` still has no request sequencing. `Spaces/index.tsx` has `userSearchSeq` pattern at line 85-113 in this same PR. Users page still lacks it. Not re-raising details — Jerry-Xin covered the mechanics.

### 💬 Non-blocking

- The `Spaces/index.tsx` change (adding `exclude_bot=1&exclude_system=1` and removing UID-suffix filter) is correct and can ship independently.
- Backend PR #63's design looks solid — contradictory param combos return 400, and count/list stay consistent through the shared `applyExcludes` path.

### ✅ Highlights

- Correct param naming (`bot_only`/`system_only`) aligns with the backend PR #63 contract — once both merge, the feature will work correctly.
- `handleTypeFilterChange` properly resets pagination.
- Tag rendering using `is_bot`/`is_system` response fields is an improvement over UID-suffix heuristic.

yujiawei (approved):
# Code Review — PR #16 (octo-admin)

## Verdict

**Approve.** Small, well-scoped diff that swaps a brittle client-side heuristic (`uid.endsWith('_bot')`) for an explicit backend contract (`bot_only` / `system_only` / `exclude_bot` / `exclude_system` + `is_bot` / `is_system` response fields), and fixes two real bugs along the way (stale-closure in `fetchData`, out-of-order response races). Worth merging once the backend has actually shipped the new params/fields in the environments this build will hit.

`tsc --noEmit` is clean against `506298414e843e2555e55c4b0a6c0764865f32a2`.

## Strengths

- **Type filter is now correct under pagination.** Previously, "Bot" / "Real users" filtered only the current page client-side, so `total` and the rendered set disagreed. With the filter pushed to the backend, `count` matches the visible result set — the headline UX bug this PR is solving is genuinely solved (`src/pages/Users/index.tsx:47-60,82-90`).
- **Stale-closure fix on `fetchData` is real.** The previous `fetchData` closed over `typeFilter` / `keyword`, so toggling the filter and immediately searching could race. Making `fetchData(nextPage, nextPageSize, nextType, kw)` explicit removes that class of bug and makes call sites readable (`src/pages/Users/index.tsx:38-74`).
- **Sequence guard via `useRef`** is the right pattern for "latest response wins" and is correctly applied to all three branches (`try` / `catch` / `finally`) so loading state can't get stranded when responses interleave (`src/pages/Users/index.tsx:44,62,66-72`).
- **System > Bot badge priority** is consistent with the filter dropdown order and avoids double-badging mixed accounts (`src/pages/Users/index.tsx:125-133`).
- **Spaces owner picker** is simplified correctly: server-side `exclude_bot=1 & exclude_system=1` plus a minimal client-side `status === 1 && is_destroy !== 1` guard, with the `_bot` suffix heuristic removed (`src/pages/Spaces/index.tsx:89-103`).

## Findings

### P2 — Stale-closure inside `handleBan`'s `Modal.confirm` callback

`handleBan` captures `page` / `pageSize` / `typeFilter` / `keyword` at render time. If the user opens the confirm modal, changes the type filter or keyword in the toolbar, and *then* confirms the ban, the post-ban refresh will use the **old** filter values (`src/pages/Users/index.tsx:105-114`).

In practice the modal is blocking, so this is hard to hit, but it's the same class of bug the rest of the PR is explicitly fixing. Cheap defense: read the latest values via refs, or just call `fetchData()` with no args and let the defaults pull from the current render — except those defaults are also closure-captured, so the safe form is to call `setPage(page)` and let the effect re-run, or to thread `fetchData` through a ref.

Not blocking — call out for a follow-up.

### P2 — `useEffect` deps omit `keyword`, which is correct but worth a comment

The effect `useEffect(..., [page, pageSize, typeFilter])` intentionally excludes `keyword` so the list doesn't refetch on every keystroke; `keyword` is only flushed on Search / Enter / type-change / refresh. The `eslint-disable-next-line` is there, and the existing Chinese comment explains the rationale for the filter split, but a one-liner specifically saying "`keyword` deliberately omitted from deps — flushed via `handleSearch` / `handleTypeFilterChange`" would save the next reader a minute (`src/pages/Users/index.tsx:76-79`).

### P2 — `handleSearch` is subtly correct; consider the simpler form

```ts
const handleSearch = () => {
  if (page !== 1) setPage(1)
  else fetchData(1, pageSize, typeFilter, keyword)
}
```

This works because when `page !== 1`, `setPage(1)` triggers the effect which then calls `fetchData(1, pageSize, typeFilter, keyword)` with the latest `keyword` closure. The two branches do diverge on one thing though: the `setPage(1)` branch goes through React's render cycle (so the request is fired one tick later), and the direct-call branch fires synchronously. If a future change accidentally adds debouncing or a synchronous side effect between the branches, the two paths will start to differ. A unified form — always `setPage(1)` then call `fetchData` explicitly — would be marginally more robust, at the cost of one extra request in the `page === 1` case unless guarded. Style preference; current code is fine.

### P2 — `Spaces.tsx` `fetchData` has no sequence guard

Pre-existing, not a regression from this PR — flagging it because the same race that motivated the new guard in `Users.tsx` exists in `Spaces.tsx`. The space list `fetchData` (`src/pages/Spaces/index.tsx:116-132`) can interleave between tab switches / pagination clicks. The user-picker `searchUsers` already has its own seq guard, just not the table itself. Worth a follow-up issue.

### P2 — Mixed `is_bot=1 && is_system=1` accounts

If the backend ever returns a row with both flags set, the UI shows the system badge and `system_only=1` / `bot_only=1` filters will each return it independently (server contract). Probably fine and matches the visual priority chosen here, but the backend doc snippet in the PR description doesn't say what's expected when a row qualifies for both — worth confirming with the API owner so this assumption is documented somewhere durable, not just implicit in the frontend.

### Nit — `UserOptionRaw` interface

`UserOptionRaw` in `Spaces/index.tsx` (lines 32-40) doesn't declare `is_bot` / `is_system`. Not needed today because the client only filters on `status` / `is_destroy`, but if a future change wants to render a "system account" hint in the picker (e.g. for the disabled-list view), the type will silently allow `u.is_bot` to be `undefined`. Optional polish.

### Nit — Refresh button signatures

`onClick={() => fetchData()}` (`src/pages/Users/index.tsx:247`, `src/pages/Spaces/index.tsx:385`) relies on every parameter having a default that resolves to the current closure value. That's correct, but `onClick={fetchData}` would not be — the click event would be passed as `nextPage`. The current code is right; just easy to break on a future tidy-up. A short comment, or `onClick={() => fetchData(page, pageSize, typeFilter, keyword)}`, would make the intent explicit.

## Suggestions for a follow-up

1. Apply the same `fetchSeq` guard pattern to `Spaces.tsx`'s table `fetchData`.
2. Lift the `handleBan` refresh to use the latest state (ref or `setPage(page)` trick) so the toolbar can be touched mid-modal without stale-refresh.
3. Push the status filter to the backend too (`online`, `banned`, `destroyed`), so the displayed row count agrees with `total` for every combination. Today, picking `human + online` shows ≤20 rows out of the per-page result and a `total` that reflects only `human`. The inline comment at line 81 already flags this gap — worth a tracked task.

## Test plan delta

The PR's own test plan is solid. Add one item:

- [ ] On page 2+ with type=Bot and a keyword typed but not searched, change type to System → the next request must include `system_only=1` **and** the typed keyword. (Validates the stale-closure fix end-to-end.)

Jerry-Xin (approved):
Summary: Project relevance gate passed; this PR updates existing octo-admin user and space-owner workflows and is ready to merge.

🔴 Blocking
None.

💬 Non-blocking
🔵 Suggestion: [src/pages/Users/index.tsx](/private/tmp/pr-review-0Bx9y5/repo/src/pages/Users/index.tsx:76) suppresses exhaustive-deps for the fetch effect. The current explicit-argument pattern is sound, but a future change to `fetchData` could make this fragile. Consider wrapping `fetchData` in `useCallback` or extracting request-param construction if this component keeps growing.

🔵 Suggestion: [src/pages/Users/index.tsx](/private/tmp/pr-review-0Bx9y5/repo/src/pages/Users/index.tsx:81) keeps status filtering client-side on the current page, so totals still reflect only backend filters, not status filters. This appears pre-existing and documented in code, so it is not a blocker for this PR.

✅ Highlights
The type filters are correctly pushed into backend query parameters at [src/pages/Users/index.tsx](/private/tmp/pr-review-0Bx9y5/repo/src/pages/Users/index.tsx:53), including the new bot-only and system-only modes.

The sequence guard in [src/pages/Users/index.tsx](/private/tmp/pr-review-0Bx9y5/repo/src/pages/Users/index.tsx:44) correctly prevents stale responses and stale errors from overwriting the current table state.

The space owner picker now requests human accounts from the backend at [src/pages/Spaces/index.tsx](/private/tmp/pr-review-0Bx9y5/repo/src/pages/Spaces/index.tsx:89) and keeps the local filter focused on active, non-destroyed users.

Verification: `npm run build` passes after installing dependencies with `npm ci`.

lml2468 (approved):
## Re-review — PR #16 (octo-admin) — 📖 齐静春

**Previous review**: CHANGES_REQUESTED at `32d1672` — two blockers: P0 backend API contract mismatch + P1 race condition.

**Both blockers resolved** ✅

### P0 — Backend API contract ✅
`octo-server` PR #63 merged at `2026-05-16T08:04:15Z`. The backend now recognizes `bot_only=1` and `system_only=1` query params, including 400 rejection for contradictory combos (`bot_only=1&exclude_bot=1`). The frontend params now match a deployed backend contract.

### P1 — Race condition ✅
`fetchSeq` ref added (lines 36, 44, 62, 66, 70) — same pattern as `userSearchSeq` in `Spaces/index.tsx`. Verified:
- `setLoading(true)` at line 45 — correctly unguarded (want loading indicator when new request fires)
- `setData`/`setTotal` at lines 63-64 — guarded by `seq !== fetchSeq.current` early return ✅
- `message.error` at line 67 — guarded ✅
- `setLoading(false)` at line 71 — guarded ✅

Stale responses are now discarded without affecting state.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- Both blockers from previous round properly addressed
- `fetchSeq` pattern is consistent with existing `userSearchSeq` in Spaces page — good codebase consistency

**Verdict: APPROVE** ✅

PR #1, opened 2026-05-11
chore: add Apache 2.0 license, OSS compliance docs, and fix internal references

lml2468:
Apache 2.0 OSS compliance: LICENSE, NOTICE, SPDX headers on all 38 source files, full README/CONTRIBUTING/SECURITY/CODE_OF_CONDUCT (en+zh), CHANGELOG, .env.example, CI workflow, fix internal refs (DMWork/wukongim/BotFather/daocloud mirror/tsdd), package.json name+license. 57 files, +976/-29.

PR #18, opened 2026-05-18
ci: build and publish multi-arch Docker images to Docker Hub

an9xyz:
## Summary

Add a hardened `docker-publish.yml` workflow that publishes `mininglamposs/octo-admin` to Docker Hub on every `v*` tag push. Direct port of the latest converged version from sibling repos ([`octo-server#72`](https://github.com/Mininglamp-OSS/octo-server/pull/72), [`octo-matter#18`](https://github.com/Mininglamp-OSS/octo-matter/pull/18), [`octo-smart-summary#15`](https://github.com/Mininglamp-OSS/octo-smart-summary/pull/15), [`octo-web#42`](https://github.com/Mininglamp-OSS/octo-web/pull/42)) — five PRs worth of review iteration baked in.

## Architecture

- `linux/amd64` + `linux/arm64` on **native runners** (`ubuntu-latest` + `ubuntu-24.04-arm`) — no QEMU
- Build uses the existing `./Dockerfile` (Node 18 build → nginx runtime)
- Push-by-digest per arch → merge into multi-arch manifest

## Hardening (converged baseline)

- All third-party actions pinned to commit SHAs
- **No-secret `validate` job** runs first; `build`/`merge` declare `needs: validate`
  - Tag push: strict semver per semver.org §9 **without** `+BUILD` metadata (prevents `v1.2.3` and `v1.2.3+build.1` clobbering the same `:MAJOR`/`:MAJOR.MINOR`/`:latest` aliases)
  - `workflow_dispatch`: Docker tag grammar + reserved-name + semver-alias-shape + non-default-branch `test-*` namespace
- **Command-injection-safe shell**: every external input via `env:`, `imagetools` args built as bash arrays
- **`docker-hub-publish` GitHub Environment** gate (already configured in repo settings):
  - Required reviewer: `Mininglamp-OSS/maintainers` team
  - Deployment ref allowlist: `main` branch + `v*` tags
- Per-arch GHA cache scope (`docker-publish-ARCH`)
- `provenance: false` (no `unknown/unknown` manifest entries)
- Concurrency block documents the GitHub 1-pending-run queue-depth tradeoff

## Prerequisites

- Org-level secrets `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` (already provisioned)
- Repo environment `docker-hub-publish` with required-reviewers + ref allowlist (configured)
- Env-level `DOCKERHUB_*` secrets if you want to fully scope credentials to this env (optional — falls back to org-level if not set)

## Test plan (run after merge to main)

> Testing on this PR branch is intentionally blocked by the environment policy. All scenarios must run from `main` after merge.

- [ ] **Env gate** — From `main`, dispatch with `tag=test-env-gate-$(date +%s)`. Run pauses at `build` with "Waiting for review"; proceeds after a maintainer approves.
- [ ] **Multi-arch manifest** — `docker buildx imagetools inspect mininglamposs/octo-admin:<tag>` shows both archs, no `unknown/unknown`.
- [ ] **Semver preflight rejects junk** — temporarily push `vfoo` (or `v01.2.3`); validate fails before any Docker Hub login. Delete tag after.
- [ ] **Prerelease skip-:latest** — push `v0.0.1-rc1`; confirm only `:0.0.1-rc1` (no `:latest`/`:0`/`:0.0`).
- [ ] **Branch dispatch rejected** — try to dispatch from any non-main branch; environment policy should block.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -250,7 +250,13 @@ jobs:
           tags: |
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
+            # `:MAJOR` alias is intentionally suppressed for v0.* releases.
+            # SemVer §4 explicitly permits breaking changes between any two
+            # 0.y.z versions, so a `:0` alias would falsely imply a stable
+            # major-0 stream. Once the project hits v1.0.0 the alias fires
+            # automatically (`:1`, `:2`, ...) with the usual major-stream
+            # backwards-compat contract.
+            type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}
             type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}
 
       - name: Create manifest list and push
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,276 @@
+name: Docker Publish
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Image tag to publish. Must match Docker tag grammar ^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$, must NOT look like a semver alias (1, 1.2, v1.2.3, ...), and from non-default branches must be prefixed with "test-".'
+        required: true
+
+permissions:
+  contents: read
+
+concurrency:
+  # Serialize ALL publishes to this image so two close-together v* tags can't
+  # race on the shared mutable tags (:latest, :MAJOR, :MAJOR.MINOR) — last
+  # writer wins, and "last finished" is not necessarily "newest version".
+  #
+  # KNOWN TRADEOFF: GitHub Actions concurrency holds at most ONE pending run
+  # per group; a newer pending run replaces the older one even with
+  # `cancel-in-progress: false`. So if you push v1.0.0 (running), v1.0.1
+  # (queued), v1.0.2 (queued) in quick succession, v1.0.1 will be canceled
+  # by v1.0.2 and never publish. Workaround: wait for each release to finish
+  # publishing before pushing the next tag. We accept this over the
+  # alternative (no concurrency = consistent :latest race on every overlap)
+  # because real release cadence is minutes-to-hours apart and protecting
+  # :latest is the higher-value invariant.
+  #
+  # NOTE: workflow-level concurrency.group only accepts github/inputs/vars
+  # contexts (not env), so the image name is hard-coded here. Keep this group
+  # name in sync with env.IMAGE_NAME below if you ever rename the image.
+  group: docker-publish-mininglamposs-octo-admin
+  cancel-in-progress: false
+
+env:
+  IMAGE_NAME: mininglamposs/octo-admin
+
+jobs:
+  validate:
+    # No-secret preflight: rejects bad inputs/refs BEFORE any job that holds
+    # Docker Hub credentials runs. Catches command-injection attempts,
+    # invalid Docker tag grammar, semver-alias clobbering, and manual
+    # publishes from unreviewed refs — all before any registry write.
+    name: Validate inputs and ref
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate tag push ref
+        if: github.event_name == 'push'
+        env:
+          REF: ${{ github.ref }}
+        run: |
+          set -euo pipefail
+          # Require strict semver per semver.org §9 (with mandatory leading 'v')
+          # so metadata-action's type=semver actually matches. A loose regex
+          # would let near-semver refs (e.g. 'vfoo', 'v01.2.3', 'v1.2.3-01',
+          # 'v1.2.3-..') reach the secret-bearing build job and push orphan
+          # digests before metadata-action quietly emits no tags — defeating
+          # the no-secret preflight intent.
+          #
+          # The regex below is the official semver.org regex (BRE/ERE form),
+          # prefixed with a literal 'v' and WITHOUT the optional +BUILD
+          # metadata suffix. Build metadata is rejected because
+          # metadata-action's {{version}}/{{major}}/{{major}}.{{minor}}
+          # patterns strip the +BUILD portion when computing Docker tag
+          # names, so v1.2.3 and v1.2.3+build.1 would both publish to
+          # :1.2.3 / :1.2 / :1 / :latest — two distinct git tags clobbering
+          # the same release aliases, breaking the no-clobber invariant.
+          #
+          # Rejects:
+          #   - leading zeros in numeric identifiers (v01.2.3, v1.2.3-01)
+          #   - empty dot-separated identifiers (v1.2.3-..)
+          #   - any +BUILD metadata (v1.2.3+build.1)
+          #   - characters outside [0-9A-Za-z.-]
+          tag="${REF#refs/tags/}"
+          semver='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?$'
+          if ! printf '%s' "$tag" | grep -Eq "$semver"; then
+            echo "::error::tag '$tag' is not strict semver per semver.org without +BUILD metadata (vMAJOR.MINOR.PATCH[-PRERELEASE]); refusing to publish"
+            exit 1
+          fi
+          echo "OK: $REF"
+
+      - name: Validate workflow_dispatch input and ref
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          INPUT_TAG: ${{ github.event.inputs.tag }}
+          REF: ${{ github.ref }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+        run: |
+          set -euo pipefail
+
+          # Docker image tag grammar (also blocks shell metacharacters as a
+          # side effect — only [A-Za-z0-9_.-] are allowed and length <= 128).
+          if ! printf '%s' "$INPUT_TAG" | grep -Eq '^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$'; then
+            echo "::error::tag '$INPUT_TAG' violates Docker tag grammar ^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$"
+            exit 1
+          fi
+
+          # Reserved auto-managed names — manual dispatch must not clobber them.
+          case "$INPUT_TAG" in
+            latest|main|master)
+              echo "::error::tag '$INPUT_TAG' is reserved/auto-managed; pick a different name"
+              exit 1
+              ;;
+          esac
+
+          # Semver-alias shapes (e.g. 1, 1.2, 1.2.3, v1.2.3) are reserved for
+          # the auto-managed major / major.minor / major.minor.patch / version
+          # aliases populated by tag-push events. A manual dispatch with
+          # these shapes would race with / overwrite the next real release.
+          if printf '%s' "$INPUT_TAG" | grep -Eq '^v?[0-9]+(\.[0-9]+){0,2}([+-].+)?$'; then
+            echo "::error::tag '$INPUT_TAG' looks like a semver alias; manual publishes must use a non-semver name (or prefix with 'test-')"
+            exit 1
+          fi
+
+          # Restrict manual publishes from non-default branches to the
+          # 'test-' namespace so feature-branch experimentation cannot land
+          # in the official release tag space. Read the default branch from
+          # repository metadata so a future rename doesn't silently weaken
+          # this guard.
+          DEFAULT_REF="refs/heads/${DEFAULT_BRANCH}"
+          if [ "$REF" != "$DEFAULT_REF" ]; then
+            case "$INPUT_TAG" in
+              test-*) echo "OK: non-default ref '$REF' with test- prefix" ;;
+              *)
+                echo "::error::manual publishes from '$REF' must use a 'test-*' tag (got '$INPUT_TAG'); merge to ${DEFAULT_BRANCH} for official tags"
+                exit 1
+                ;;
+            esac
+          fi
+
+          echo "OK: dispatch from $REF with tag '$INPUT_TAG'"
+
+  build:
+    name: Build ${{ matrix.platform }}
+    runs-on: ${{ matrix.runner }}
+    needs: validate
+    # `docker-hub-publish` environment gates the Docker Hub credentials:
+    #   - required reviewers (Mininglamp-OSS/maintainers team)
+    #   - deployment ref allowlist: refs/heads/main, refs/tags/v*
+    # No DOCKERHUB_* secret reaches a runner until a maintainer approves
+    # the run in the Actions UI.
+    environment: docker-hub-publish
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+            arch: amd64
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+            arch: arm64
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: ${{ matrix.platform }}
+          outputs: type=image,name=${{ env.IMAGE_NAME }},push-by-digest=true,name-canonical=true,push=true
+          cache-from: type=gha,scope=docker-publish-${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=docker-publish-${{ matrix.arch }}
+          # Disable provenance attestation so the final manifest list doesn't
+          # carry an extra "unknown/unknown" platform entry that confuses some
+          # registry UIs and scanners.
+          provenance: false
+
+      - name: Export digest
+        env:
+          BUILD_DIGEST: ${{ steps.build.outputs.digest }}
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/digests
+          touch "/tmp/digests/${BUILD_DIGEST#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: digests-${{ matrix.arch }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    name: Create multi-arch manifest
+    runs-on: ubuntu-latest
+    needs: build
+    # Same gate as `build` — manifest creation also needs Docker Hub creds.
+    # The single approval at the start of the run covers both jobs.
+    environment: docker-hub-publish
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          path: /tmp/digests
+          pattern: digests-*
+          merge-multiple: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          # `flavor: latest=auto` is the metadata-action default, set explicitly
+          # so the prerelease-skipping behavior is visible without consulting
+          # action docs. With latest=auto + a {{version}} semver tag entry,
+          # `:latest` is published for stable releases (v1.2.3) and skipped
+          # for prereleases (v1.2.3-rc1).
+          flavor: |
+            latest=auto
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}
+
+      - name: Create manifest list and push
+        working-directory: /tmp/digests
+        env:
+          IMAGE: ${{ env.IMAGE_NAME }}
+          METADATA: ${{ steps.meta.outputs.json }}
+        run: |
+          set -euo pipefail
+          # Build "-t TAG" args as an array — never interpolate tags into the
+          # shell command line, which would be a command-injection vector
+          # if metadata-action ever emitted a tag containing shell syntax.
+          tags=()
+          while IFS= read -r tag; do
+            tags+=(-t "$tag")
+          done < <(jq -r '.tags[]' <<< "$METADATA")
+
+          if [ ${#tags[@]} -eq 0 ]; then
+            echo "::error::metadata-action produced no tags; refusing to publish manifest"
+            exit 1
+          fi
+
+          # Digest filenames in /tmp/digests are SHA-256 hex strings (validated
+          # by upload-artifact's if-no-files-found:error and the controlled
+          # filename derived from build.outputs.digest), safe to feed in.
+          sources=()
+          for f in *; do
+            sources+=("${IMAGE}@sha256:${f}")
+          done
+
+          docker buildx imagetools create "${tags[@]}" "${sources[@]}"
+
+      - name: Inspect image
+        env:
+          IMAGE: ${{ env.IMAGE_NAME }}
+          PRIMARY_TAG: ${{ steps.meta.outputs.version }}
+        run: |
+          set -euo pipefail
+          docker buildx imagetools inspect "${IMAGE}:${PRIMARY_TAG}"
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -114,6 +114,22 @@ jobs:
             exit 1
           fi
 
+          # Reject tag refs for workflow_dispatch. metadata-action's
+          # type=semver patterns fire on refs/tags/v* even on dispatch
+          # events, so dispatching from a tag ref would simultaneously
+          # emit the auto-managed semver aliases (:MAJOR / :MAJOR.MINOR /
+          # :latest, from the tag's version) AND the raw input tag,
+          # silently overwriting an existing release. The GitHub web UI
+          # only offers branch refs in the "Run workflow" dropdown, but
+          # the REST/GraphQL API accepts any ref — close that gap here.
+          case "$REF" in
+            refs/heads/*) ;;
+            *)
+              echo "::error::workflow_dispatch must be triggered from a branch ref (got '$REF'). Publishing from a tag ref would double-emit auto-managed release aliases. Use 'git push origin v<X.Y.Z>' for tag releases."
+              exit 1
+              ;;
+          esac
+
           # Restrict manual publishes from non-default branches to the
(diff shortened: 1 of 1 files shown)


yujiawei (commented):
# Code Review — PR #18 (octo-admin)

## Summary

High-quality direct port of the converged multi-arch Docker publish workflow from `octo-server#72` / `octo-matter#18` / `octo-smart-summary#15` / `octo-web#42`. Hardening posture is strong: SHA-pinned third-party actions, no-secret preflight `validate` job, strict semver gate without `+BUILD`, Docker tag grammar + reserved-name + alias-shape rejection on dispatch, env-gated credentials, native ARM runner (no QEMU), per-arch GHA cache scope, `provenance: false`, and command-injection-safe array-built `imagetools` invocation. Comments inside the YAML explain *why* (not just *what*), which is what makes this maintainable.

Net: the workflow is ready to ship. One defense-in-depth gap and a couple of minor robustness suggestions below — none block.

## Findings

### 1. (P2 — defense-in-depth) `workflow_dispatch` from a tag ref can double-emit semver aliases

`validate.Validate workflow_dispatch input and ref` checks branch namespace (non-default branch → must be `test-*`) but does **not** reject tag refs. If a maintainer dispatches the workflow from `refs/tags/v1.2.3` with `inputs.tag=test-foo`, the `merge` job's metadata-action will fire **both** code paths:

- `type=semver,pattern={{version}}` (and `{{major}}.{{minor}}`, `{{major}}`) — matches because `github.ref` is a tag → emits `:1.2.3` / `:1.2` / `:1` / `:latest`
- `type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}` — emits `:test-foo`

Result: a single dispatch silently re-publishes the `v1.2.3` release aliases pointing at whatever the current tree (`actions/checkout` on `github.ref`) builds — which may not be what `v1.2.3` originally pointed at if the tag was force-moved, and definitely interacts badly with the no-clobber intent of the strict semver gate.

It's a low-likelihood path (requires a maintainer with dispatch privileges deliberately choosing a tag ref in the Actions UI), but it's the only remaining way to clobber a release alias from outside the push-tag flow.

**Two equally good fixes — pick one:**

```yaml
# Option A — reject tag refs in dispatch validate (single source of truth)
case "$REF" in
  refs/tags/*) echo "::error::workflow_dispatch from a tag ref is not allowed; dispatch from a branch"; exit 1 ;;
esac
```

```yaml
# Option B — scope the semver tag entries to push events in metadata-action
tags: |
  type=semver,pattern={{version}},enable=${{ github.event_name == 'push' }}
  type=semver,pattern={{major}}.{{minor}},enable=${{ github.event_name == 'push' }}
  type=semver,pattern={{major}},enable=${{ github.event_name == 'push' }}
  type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}
```

Option B is the more local fix (keeps validate focused on input shape) and also defends against future trigger additions; Option A is the smaller diff.

### 2. (P3 — minor) `for f in *` in the merge step has no `nullglob` safety net

`merge.Create manifest list and push`:

```bash
sources=()
for f in *; do
  sources+=("${IMAGE}@sha256:${f}")
done
```

Under default bash behavior, if `/tmp/digests` is ever empty the loop runs once with the literal `f=*`, producing the bogus source `mininglamposs/octo-admin@sha256:*` and feeding it to `imagetools create`. Today this is guarded upstream — `upload-artifact` is configured `if-no-files-found: error` and `build` is `needs:` of `merge` — so the empty case shouldn't reach here. But the safety is implicit, not enforced at the point of use. Either of:

```bash
shopt -s nullglob
```

…before the loop, or an explicit existence check before iterating, makes the invariant local and survives future refactors that change how digests get into `/tmp/digests`.

### 3. (P3 — minor) Same defensive check on the `imagetools` argv

The `sources` array could theoretically still be empty after the loop (in the nullglob fix, or if the implicit guarantees ever change). The script already does the equivalent check for `tags`:

```bash
if [ ${#tags[@]} -eq 0 ]; then
  echo "::error::metadata-action produced no tags; refusing to publish manifest"
  exit 1
fi
```

Adding the symmetric `${#sources[@]} -eq 0` check immediately after the source loop costs nothing and makes the merge job fail loud rather than emit a confusing `imagetools` usage error.

### 4. (Note — not actionable in this PR) Environment-approval semantics comment

The comment in `merge` says:

> The single approval at the start of the run covers both jobs.

In practice GitHub batches environment-approval prompts for jobs that become eligible at the same time, but `merge` only becomes eligible after both `build` matrix jobs finish, so it's a separate deployment-pending event from the `build` jobs' approval. Empirically reviewers usually see one prompt that lists multiple jobs, but the platform doesn't *guarantee* a single click — if `merge` starts noticeably later than the `build` matrix, a second approval may appear. The comment slightly oversells this; not worth churning a port over, but worth knowing for the converged source-of-truth doc.

## Verification Checklist

- ✅ Strict semver regex matches semver.org §9 with literal `v` prefix and `+BUILD` excluded — confirmed against `v0.0.1`, `v1.2.3`, `v1.2.3-rc.1`, `v1.2.3-0.3.7`, and rejects `vfoo`, `v01.2.3`, `v1.2.3-01`, `v1.2.3-..`, `v1.2.3+build.1`.
- ✅ Docker tag input grammar matches Docker's actual constraint (max 128 chars, can't start with `.`/`-`): `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$` = 1 + up to 127 = ≤128 total.
- ✅ Semver-alias shape regex catches `1`, `1.2`, `1.2.3`, `v1.2.3`, and `1.2.3-rc1`.
- ✅ All third-party actions SHA-pinned (`actions/checkout@34e114…`, `setup-buildx-action@8d2750c6…`, `login-action@c94ce9fb…`, `build-push-action@10e90e36…`, `upload-artifact@ea165f8d…`, `download-artifact@d3f86a10…`, `metadata-action@c299e40c…`). Tag comments next to each pin make future SHA bumps verifiable.
- ✅ External inputs flow through `env:` blocks, never directly interpolated into shell. `imagetools create` argv is array-built — no command-injection surface from a malicious metadata-action tag value.
- ✅ Per-arch GHA cache scope (`docker-publish-amd64`/`docker-publish-arm64`) avoids cross-arch cache pollution.
- ✅ `provenance: false` prevents the `unknown/unknown` manifest entry.
- ✅ Concurrency tradeoff (1-pending-run queue cap) is documented; the choice to protect `:latest` over allowing every rapid release through is the right one for this image.
- ⚠️ `workflow_dispatch` tag-ref dispatch is not rejected (Finding 1).
- ⚠️ `for f in *` lacks `nullglob` (Finding 2).

## Recommendation

**COMMENTED** — ship-ready with one small defense-in-depth follow-up (Finding 1) recommended before this becomes the canonical pattern across more repos. Findings 2 and 3 are nice-to-haves. The work is solid.

Jerry-Xin (dismissed):
The PR is relevant to `octo-admin` and the new Docker publish workflow is ready to merge; I found no blocking issues.

🔴 Blocking

None.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/docker-publish.yml:49`: If the release policy requires tags to come only from `main`, consider adding an ancestry check against `origin/main`. The current environment allowlist permits `refs/tags/v*`, so a valid semver tag on any pushed commit can reach the gated publish path once approved.

🔵 Suggestion — `Dockerfile:1` and `Dockerfile:8`: This workflow will publish images built from mutable base tags (`node:18-alpine`, `nginx:alpine`). That is pre-existing Dockerfile behavior, but pinning base image digests would improve release reproducibility if that is a project goal.

✅ Highlights

The workflow is well scoped to the project’s Docker image release path, uses a no-secret validation job before credentialed jobs, pins third-party actions by SHA, avoids QEMU by using native arch runners, builds per-arch digests before creating the manifest, and handles shell inputs with environment variables and arrays instead of unsafe interpolation.

Validation behavior and workflow syntax were checked locally: representative semver/manual tag cases match the documented rules, `git diff --check` passed, and `actionlint` reported no issues.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-admin` and the workflow is ready to merge; I found no blocking correctness or security issues.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion: In `.github/workflows/docker-publish.yml:260`, the comment says digest filenames are safe, but the merge script does not explicitly validate `*` as SHA-256 hex before passing them to `imagetools`. The producer is controlled by the build job, so this is not blocking, but adding a simple `[[ "$f" =~ ^[0-9a-f]{64}$ ]]` guard would make the invariant executable.

✅ Highlights:
- `.github/workflows/docker-publish.yml:40` keeps validation in a no-secret preflight job before Docker Hub credentials are available.
- `.github/workflows/docker-publish.yml:76` uses strict semver validation and intentionally rejects build metadata to avoid alias clobbering.
- `.github/workflows/docker-publish.yml:250` builds `imagetools` arguments as Bash arrays, avoiding shell interpolation of tags.
- `.github/workflows/docker-publish.yml:145` uses native amd64/arm64 runners and per-arch cache scopes, which fits the stated multi-arch publishing goal.

lml2468 (dismissed):
[CHANGES_REQUESTED]

Summary: Well-structured multi-arch Docker publish workflow with strong security posture (SHA-pinned actions, no-secret preflight, env-gated credentials, command-injection-safe shell). However, there is a **P0 correctness bug**: all regex patterns in the validate job use `\\.` inside bash single-quoted strings, which in ERE means "literal backslash + any character" rather than "literal dot". This makes the tag-push validation reject every valid semver tag, and the dispatch semver-alias filter catch nothing.

🔴 Blocking

- **`.github/workflows/docker-publish.yml:77` — Semver regex is broken (P0)**

  The pattern:
  ```bash
  semver='^v(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?$'
  ```
  In bash single-quoted strings, `\\` is literally two characters (`\` + `\`). When passed to `grep -Eq`, ERE interprets `\\.` as "match a literal backslash followed by any char" — NOT "match a literal dot."

  **Result**: Every valid semver tag (`v1.2.3`, `v0.0.1`, etc.) will be rejected. The workflow's primary function (publish on tag push) is dead on arrival.

  Verified locally:
  ```
  $ semver='^v(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)...$'
  $ printf "v1.2.3" | grep -Eq "$semver"  # exits 1 (no match)
  ```

  **Fix**: Use single backslash `\.` in single-quoted strings (bash single quotes do zero escape processing, and ERE treats `\.` as a literal dot):
  ```bash
  semver='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?$'
  ```

- **`.github/workflows/docker-publish.yml:112` — Semver-alias filter has the same bug (P1)**

  ```bash
  if printf '%s' "$INPUT_TAG" | grep -Eq '^v?[0-9]+(\\.[0-9]+){0,2}([+-].+)?$'; then
  ```
  Same `\\.` issue. The filter fails to catch semver-alias-shaped inputs (`1.2.3`, `v1.2`), defeating the clobbering protection.

  **Fix**: `^v?[0-9]+(\.[0-9]+){0,2}([+-].+)?$`

💬 Non-blocking

- Agree with yujiawei's P2 finding: `workflow_dispatch` from a tag ref can double-emit semver aliases. Recommend Option B (scope semver tag entries to push events).
- `.github/workflows/docker-publish.yml:264` — `for f in *` lacks `shopt -s nullglob`; currently guarded by upstream `if-no-files-found: error` but the invariant is implicit.

✅ Highlights

- Excellent defense-in-depth architecture: no-secret preflight → env-gated build → manifest merge
- All 7 third-party actions SHA-pinned with version comments for auditability
- Shell injection prevention via `env:` blocks + array-built argv for `imagetools create`
- Native ARM runners (no QEMU) + per-arch cache scoping
- Inline comments explain *why*, not just *what* — rare and valuable for CI workflows
- Concurrency tradeoff is explicitly documented with the accepted risk

Note: Since this is a "converged" port used across sibling repos (octo-server, octo-matter, octo-smart-summary, octo-web), the same regex bug likely exists in those repos too.

an9xyz:
@lml2468 @guobin-a — verified the regex is correct. Reproducible evidence:

**1. Byte-level dump of the pushed file** (`od -c` on the line, character by character):

```
)   \   .   (   0   |   [   1   -   9   ]   [   0   -
```

That's literally three characters: `)` then `\` (a single backslash) then `.` — **not** two backslashes. The displayed `\\.` in a Markdown comment is the renderer escaping a single `\` for display.

**2. End-to-end test against the pushed file** (downloads raw, extracts the regex variable, runs against canonical cases):

```bash
remote_regex=$(curl -s https://raw.githubusercontent.com/Mininglamp-OSS/octo-admin/ci/docker-publish/.github/workflows/docker-publish.yml | grep 'semver=' | sed "s/.*semver='\(.*\)'/\1/")
for t in v1.2.3 v0.0.1 v1.2.3-rc.1 vfoo v01.2.3 v1.2 v1.2.3+build.1; do
  printf '%s' "$t" | grep -Eq "$remote_regex" && echo "$t PASS" || echo "$t REJECT"
done
```

Output:

```
v1.2.3 PASS
v0.0.1 PASS
v1.2.3-rc.1 PASS
vfoo REJECT
v01.2.3 REJECT
v1.2 REJECT
v1.2.3+build.1 REJECT
```

**3. Empirical: this exact regex is already live in 4 sibling repos** (octo-server, octo-matter, octo-web, octo-smart-summary) and has successfully gated published tags including v1.0.0, v1.0.1, v1.1.0, v1.1.1, v1.2.0. If the regex rejected every valid semver, none of those releases would have made it past the validate job.

I suspect the reproduction shown in the previous comment used a copy of the regex pasted from a Markdown-rendered source where `\.` displayed as `\\.` — the literal backslash got double-rendered. Pasting that into bash gives the broken `\\.` form you tested, but that's not what's in the file.

Please dismiss when confirmed.

lml2468:
## ❌ Retracting my CHANGES_REQUESTED — the regex is correct

@an9xyz you're right, my apologies.

**What happened**: I obtained the file content through `gh pr diff` and `gh api .../contents` (base64-decoded), both of which go through JSON transport that encodes a single backslash as `\\`. I then pasted that double-backslash form into a local bash test — which naturally failed. I should have verified against the raw file before submitting.

**Proof the file is correct**:
```
$ curl -s .../docker-publish.yml | grep 'semver=' | od -c
...  )   \   .   (   0   |   ...
```
That's a single `\` followed by `." — ERE treats `\.` as a literal dot match. Correct.

**Lesson for me**: When reviewing CI/shell code with regex patterns, always verify against the raw source bytes, not rendered/transported representations that may escape backslashes.

Review dismissed. Sorry for the noise.

yujiawei (dismissed):
# Code Review — PR #18 (octo-admin)

## Verdict

**Approved.** This is a direct port of the converged `docker-publish.yml` from `octo-server#72` / `octo-matter#18` / `octo-smart-summary#15` / `octo-web#42`, with two correct local adaptations:

1. Image name + concurrency group renamed to `octo-admin`.
2. `actions/checkout` no longer needs `fetch-depth: 0` — verified that `Dockerfile` (Node 18 build → nginx runtime) does not invoke `git describe` or otherwise depend on full history/tags.

The workflow itself reflects five PRs of accumulated review iteration; no architectural rework is warranted here.

## What was verified

- **Strict semver regex** (`v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-…)?$`, **without** `+BUILD`) is correct and blocks the documented clobber cases (`v01.2.3`, `v1.2.3-01`, `v1.2.3-..`, `v1.2.3+build.1`). Anchored with `^` and `$`; `printf '%s'` avoids trailing-newline anchor issues.
- **Dispatch input validation** rejects: bad Docker tag grammar, reserved names (`latest`/`main`/`master`), semver-alias shapes (`1`, `1.2`, `v1.2.3`), and (new vs. `octo-server`) tag refs as the workflow ref. The tag-ref rejection closes a real REST-API-only attack surface that the web UI doesn't expose — good catch.
- **No-secret preflight**: `validate` declares no `environment:` and references no `secrets.*`, so it runs without holding Docker Hub credentials and gates the secret-bearing jobs via `needs: validate`. Correct design.
- **Command-injection safety**: every external input (`REF`, `INPUT_TAG`, `DEFAULT_BRANCH`, `BUILD_DIGEST`, `IMAGE`, `METADATA`, `PRIMARY_TAG`) is plumbed through `env:` rather than interpolated into the shell line. `imagetools create` builds `-t`/source args as bash arrays via `jq -r` + a `while read` loop — safe even if `metadata-action` ever emitted a tag with shell metacharacters.
- **Action pinning**: all third-party actions (`actions/checkout`, `docker/setup-buildx-action`, `docker/login-action`, `docker/build-push-action`, `docker/metadata-action`, `actions/upload-artifact`, `actions/download-artifact`) pinned to 40-char commit SHAs with `# vX.Y.Z` comments.
- **Per-arch cache scopes** (`docker-publish-amd64` / `docker-publish-arm64`) prevent cross-arch cache poisoning.
- **`provenance: false`** to avoid the `unknown/unknown` manifest entry, consistent with the convergent baseline.
- **Concurrency tradeoff** (workflow-level group, last-pending-wins) is explicitly documented inline with the operational mitigation ("wait for each release to finish before pushing the next tag"). Reasonable for the expected release cadence.

## Observations (non-blocking)

### 1. PR description overstates approval gating — actual behavior is 3 prompts, not 1

PR body says about `merge`: *"The single approval at the start of the run covers both jobs."*

Per GitHub Actions environment behavior, a deployment is created **per job** that declares `environment:`. With this workflow that is **three** separate deployments / approval prompts on every release:

- `build (linux/amd64)`
- `build (linux/arm64)`
- `merge`

The Actions UI batches pending approvals into one "Review pending deployments" dialog when they arrive simultaneously, which is likely the source of the "single approval" impression. But `merge` only requests its approval **after** `build` finishes, so it shows up later as a second prompt. Worth correcting in the PR description and/or operator runbook so the on-call maintainer expects to revisit the run.

This is a docs nit, not a code change. The gating itself is correct (no credentials reach a runner without explicit maintainer approval).

### 2. Belt-and-suspenders for empty digest set in `merge`

`merge` depends on `build` (all matrix entries), and `upload-artifact` has `if-no-files-found: error`, so in practice `/tmp/digests/` is always non-empty when `merge` runs. The existing `if [ ${#tags[@]} -eq 0 ]` guard then catches the metadata-action-emits-nothing case before any `imagetools create` call.

However, the `sources=()` build uses bare `for f in *` with default globbing, so if `/tmp/digests` ever *were* empty, `f` would expand to the literal string `*` and `sources` would receive `${IMAGE}@sha256:*` — a confusing `imagetools` error rather than an explicit refusal. Cheap to harden either with `shopt -s nullglob` before the loop or an explicit `[ ${#sources[@]} -eq 0 ]` check mirroring the tags guard. Suggest, do not require.

### 3. Orphan digests on partial matrix failure

`strategy.fail-fast: false` + `push-by-digest=true` means if `linux/amd64` succeeds and `linux/arm64` fails, an amd64-only digest is pushed to Docker Hub but never referenced by a manifest list (because `merge` is skipped by `needs: build`). This is an accepted tradeoff in the convergent design (fail-fast on push-by-digest hides one arch's diagnostics behind the other) and Docker Hub will GC unreferenced layers eventually, but it does leave a brief window of unreferenced blobs. Worth being aware of when interpreting Docker Hub storage metrics; no action needed.

## Test plan coverage

The five-item test plan in the PR description is appropriate and covers the security-critical paths (env gate, multi-arch manifest, semver preflight rejection, prerelease `:latest` skip, branch-dispatch rejection). The intentional inability to test from the PR branch is correctly called out — environment ref allowlist blocks it by design.

Recommend adding a sixth bullet only if Observation #1 surprises the first releaser: confirm exact number of approval prompts during the env-gate test and update docs.

## Summary

Ship it. The workflow is well-engineered, the local adaptation is minimal and correct, and the remaining observations are either documentation polish or optional belt-and-suspenders hardening. None block merging.

Jerry-Xin (changes requested):
The PR is relevant to `octo-admin`, but one release-tagging issue should be fixed before merge.

**🔴 Blocking**
- 🟡 Warning: `.github/workflows/docker-publish.yml:253` always publishes the `{{major}}` tag, so a stable `v0.x.y` release will publish and keep moving `mininglamposs/octo-admin:0`. Because this repo is currently pre-1.0 (`package.json` is `0.1.0`), that tag would imply a stable major stream even though SemVer allows breaking changes across `0.y.z`. Disable the major alias for `v0.*` releases, e.g. `enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}`, while keeping `{{version}}` and `{{major}}.{{minor}}`.

**💬 Non-blocking**
- None.

**✅ Highlights**
- The no-secret `validate` job, strict tag validation, environment gate, SHA-pinned third-party actions, per-arch digest publishing, and array-based `imagetools` invocation are solid hardening choices.
- The workflow aligns with the existing `Dockerfile` and repository deployment surface.

an9xyz:
Addressed Jerry-Xin's feedback in 891a526:

**🟡 (Jerry-Xin) — pre-1.0 `:MAJOR` alias would falsely imply a stable major-0 stream**
Added `enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}` to the `{{major}}` pattern only. SemVer §4 allows breaking changes between any two `0.y.z` versions, so `:0` should not advertise a stable major stream. Once the repo reaches v1.0.0 the alias fires automatically (`:1`, `:2`, ...) with the standard contract. `:MAJOR.MINOR` and full `:version` aliases are unaffected — v0.x.y still gets `:0.x` and `:0.x.y`.

This guard is defensive — the planned first release tag is v1.0.0 (same as sibling repos), so it never fires in practice today. But it costs ~1 line and protects future maintainers if a v0.x line is ever cut.

Note: this is the only meaningful divergence from the converged sibling-workflow pattern. The other four repos (`octo-server`, `octo-matter`, `octo-web`, `octo-smart-summary`) are already past v1.0.0 so the same guard would be a no-op there; not adding it to those workflows unless they reverse to v0.x.

yujiawei (approved):
# Code Review — PR #18 (octo-admin)

## Verdict
**APPROVED.** This is a faithful port of the converged `docker-publish.yml` from the sibling repos (`octo-server#72`, `octo-matter#18`, `octo-smart-summary#15`, `octo-web#42`), plus two genuine improvements on top: rejecting `workflow_dispatch` from tag refs, and suppressing the `:MAJOR` alias for `v0.*` releases. Both deltas have clear, correct reasoning in their commit messages.

## Validation

### Action SHA pins (all match advertised version tags)
| Action | Pinned SHA | Verified tag |
| --- | --- | --- |
| `actions/checkout` | `34e1148…f8d5` | v4.3.1 ✅ |
| `docker/setup-buildx-action` | `8d2750c…cd6f` | v3.12.0 ✅ |
| `docker/login-action` | `c94ce9f…bcc9` | v3.7.0 ✅ |
| `docker/build-push-action` | `10e90e3…78e8` | v6.19.2 ✅ |
| `actions/upload-artifact` | `ea165f8…fa02` | v4.6.2 ✅ |
| `actions/download-artifact` | `d3f86a1…8093` | v4.3.0 ✅ |
| `docker/metadata-action` | `c299e40…9051` | v5.10.0 ✅ |

### Repo prerequisites
- `Dockerfile`, `package-lock.json`, `nginx.conf.template`, `.dockerignore` all present at the head SHA. The Dockerfile produces a Node 18 → nginx static-serve image consistent with the workflow's assumptions.
- `concurrency.group: docker-publish-mininglamposs-octo-admin` matches `env.IMAGE_NAME: mininglamposs/octo-admin` — the inline comment about keeping them in sync is now load-bearing.

### Hardening invariants
- **No-secret preflight**: `validate` runs without `environment:`; `build` and `merge` declare `needs: validate`, so no `DOCKERHUB_*` secret reaches a runner until the input/ref check passes.
- **Command-injection safe**: every external value (`REF`, `INPUT_TAG`, `DEFAULT_BRANCH`, `BUILD_DIGEST`, `METADATA`, `IMAGE`, `PRIMARY_TAG`) flows through `env:`; imagetools args are constructed as a bash array, never interpolated into the shell command line.
- **Semver tag-push regex** correctly rejects `vfoo`, `v01.2.3`, `v1.2.3-01`, `v1.2.3-..`, and any `+BUILD` metadata variant — the four ways a near-semver ref would otherwise reach the secret-bearing job and either (a) emit no tags from metadata-action or (b) clobber an existing release's aliases.
- **`workflow_dispatch` input validation** layers four checks (Docker tag grammar → reserved names → semver-alias-shape → tag-ref rejection → non-default-branch `test-*` namespace) before any registry write.
- **`docker-hub-publish` environment gate** is declared on both `build` and `merge`. A single maintainer approval covers both jobs, and the environment's deployment-ref allowlist (main + `v*` tags, per PR body) is what makes "PR-branch CI is blocked from publishing" structurally true rather than depending on workflow YAML alone.
- **Per-arch cache scope** (`docker-publish-amd64`, `docker-publish-arm64`) prevents the two parallel matrix builds from invalidating each other's GHA cache.
- **`provenance: false`** keeps the manifest list free of `unknown/unknown` entries.

### Delta commit 2 — reject `workflow_dispatch` from tag refs (`4fb2f50`)
Genuine API-side gap closure. The web UI dropdown only offers branch refs, but `gh workflow run --ref v1.2.3` (or the GraphQL equivalent) accepts a tag ref and would let `type=semver` patterns fire on `github.ref` simultaneously with the `type=raw` input — silently double-emitting `:MAJOR` / `:MAJOR.MINOR` / `:latest` on top of the requested `test-foo`. The `case "$REF" in refs/heads/*) ;; *) … ;; esac` allowlist is the right shape. Error message includes the recommended fix ("use `git push origin v<X.Y.Z>` for tag releases") so a maintainer who hits the guard knows the path forward without grep'ing the YAML.

### Delta commit 3 — suppress `:MAJOR` for v0.* releases (`891a526`)
Correct application of SemVer §4 (any 0.y.z → 0.y'.z' may be breaking). Publishing `:0` for a `v0.x.y` release would falsely advertise a stable major-0 stream, surprising anyone pinning to `mininglamposs/octo-admin:0`. The `enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}` guard targets the right pattern (`{{major}}` only) and leaves `{{version}}`/`{{major}}.{{minor}}` untouched, so `:0.x.y` and `:0.x` still publish. Cost-free defense; once the repo cuts `v1.0.0`, the alias fires automatically with the usual major-stream contract. The release plan starts at `v1.0.0` per the PR body, so this is precautionary rather than load-bearing today — but cheap to keep, expensive to add later.

### Minor observations (non-blocking)

1. **Test plan doesn't exercise the v0.* suppression.** The PR body's checklist was written before delta commit 3 landed. Suggest adding (during the post-merge verification, not a blocker for this PR): from `main`, dispatch with `tag=test-v0-major-suppress` is irrelevant — instead, the right verification is to push a real (or test) `v0.0.x` tag and confirm `docker buildx imagetools inspect mininglamposs/octo-admin:0` fails / does not exist while `:0.0` and `:0.0.x` do. Since the repo plans to start at `v1.0.0`, this can also be verified later when the first v0.x test tag is cut, or skipped entirely.

2. **`Inspect image` step uses `steps.meta.outputs.version` as `PRIMARY_TAG`.** For tag pushes this is the `{{version}}` semver. For `workflow_dispatch` with a raw input tag, metadata-action's priority logic falls back to the `type=raw` value (semver patterns yield nothing on a branch ref), so the inspect still resolves. Not a bug — flagging because if someone later adds a higher-priority `type=ref` or schedule entry, the inspect target could shift; the inline comment near `flavor: latest=auto` is a good model and a similar one near the inspect step would future-proof it.

3. **Concurrency tradeoff is acknowledged explicitly.** The "v1.0.0 (running), v1.0.1 (queued), v1.0.2 (queued) → v1.0.1 cancelled" failure mode is documented inline. Worth noting that for a release cadence of "minutes-to-hours apart," this is the right tradeoff; if octo-admin ever ships rapid-fire patch releases (e.g. CI-generated tags), revisit.

## Summary
Clean port. Both deltas earn their keep. Action pins verified. No P0/P1/P2 issues. Approve.

Jerry-Xin (approved):
The PR is in scope for `octo-admin`: it adds Docker Hub publishing for the repository’s existing `Dockerfile`, and I found no blocking issues.

**🔴 Blocking**
None.

**💬 Non-blocking**
🔵 Suggestion: In `.github/workflows/docker-publish.yml:198`, consider validating `BUILD_DIGEST` against `^sha256:[0-9a-f]{64}$` before creating the artifact file. The value is action-controlled, so this is not a blocker, but it would make failures clearer if `docker/build-push-action` ever returns an unexpected digest.

**✅ Highlights**
The no-secret `validate` job, environment gate, SHA-pinned actions, safe shell input handling, and digest-first multi-arch manifest flow are well aligned with the security goals.

I also verified the configured `docker-hub-publish` environment has `main` and `v*` deployment policies, and `actionlint` passes for the new workflow. I could not complete `npm ci` locally because the workspace filesystem ran out of disk space, but this PR only changes the GitHub Actions workflow.

lml2468 (approved):
Summary: Well-structured multi-arch Docker publish workflow with strong security posture. Verified the regex patterns are correct against raw file bytes. No blocking issues found.

My earlier CHANGES_REQUESTED was a false alarm (backslash escaping display artifact from JSON transport — not a real bug). Dismissing and approving.

💬 Non-blocking (agreeing with prior reviewers, not new findings):

- `.github/workflows/docker-publish.yml:264` — `for f in *` relies on upstream `if-no-files-found: error` for the non-empty invariant; adding `shopt -s nullglob` + explicit empty-check would make it locally enforced (yujiawei P3)
- `.github/workflows/docker-publish.yml` — `workflow_dispatch` from a tag ref can double-emit semver aliases via metadata-action; low-likelihood but fixable with one `case` guard (yujiawei P2)

✅ Highlights:

- All 7 third-party actions SHA-pinned with version comments
- No-secret preflight `validate` job before any credentialed work
- Shell injection prevention: `env:` blocks + array-built argv for `imagetools create`
- Native ARM runners (no QEMU), per-arch cache scope
- Concurrency tradeoff explicitly documented
- Inline comments explain rationale, not just behavior — makes the workflow maintainable

LGTM, ship it.

PR #20, opened 2026-05-18
feat: add system setting admin page

an9xyz:
## Summary
- add a System Setting page for registration and SMTP runtime settings
- wire the page into the super-admin routes and sidebar menu
- support configured/effective_value responses from the backend and preserve encrypted SMTP passwords safely

## Notes
- This UI targets the manager system-setting APIs added by Mininglamp-OSS/octo-server#76.
- Bool settings keep a "follow default config" option so admins can clear DB overrides while still seeing the effective runtime value.

## Test
- npm run build

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,6 +10,7 @@ import Groups from './pages/Groups'
 import Spaces from './pages/Spaces'
 import Backup from './pages/Backup'
 import Download from './pages/Download'
+import SystemSetting from './pages/SystemSetting'
 import Changelog from './pages/Changelog'
 import AppBots from './pages/AppBots'
 import SpaceEntry from './pages/SpaceAdmin/SpaceEntry'
@@ -111,6 +112,7 @@ function AdminRoutes() {
         <Route path="users" element={<Users />} />
         <Route path="groups" element={<Groups />} />
         <Route path="spaces" element={<Spaces />} />
+        <Route path="system-setting" element={<SystemSetting />} />
         <Route path="backup" element={<Backup />} />
         <Route path="download" element={<Download />} />
         <Route
--- a/src/api/system-setting.ts
+++ b/src/api/system-setting.ts
@@ -1,0 +1,42 @@
+import api from './index'
+
+export type SystemSettingValueType = 'bool' | 'string' | 'int' | 'encrypted'
+
+export interface SystemSettingItem {
+  category: string
+  key: string
+  configured: boolean
+  value: string
+  effective_value: string
+  value_type: SystemSettingValueType
+  description: string
+}
+
+export interface SystemSettingSchemaItem {
+  category: string
+  key: string
+  type: SystemSettingValueType
+  description: string
+}
+
+export interface SystemSettingResponse {
+  items: SystemSettingItem[]
+  schema: SystemSettingSchemaItem[]
+}
+
+export interface SystemSettingUpdateItem {
+  category: string
+  key: string
+  value: string
+}
+
+export const SECRET_MASK = '****'
+
+export const getSystemSettings = () =>
+  api.get<SystemSettingResponse>('/v1/manager/common/system_setting').then((res) => res.data)
+
+export const updateSystemSettings = (items: SystemSettingUpdateItem[]) =>
+  api.post('/v1/manager/common/system_setting', { items })
+
+export const testSystemSettingEmail = (to: string) =>
+  api.post('/v1/manager/common/system_setting/test_email', { to })
--- a/src/layouts/MainLayout.tsx
+++ b/src/layouts/MainLayout.tsx
@@ -16,6 +16,7 @@ import {
   MoonOutlined,
   DesktopOutlined,
   RobotOutlined,
+  SettingOutlined,
 } from '@ant-design/icons'
 import { useAuthStore } from '../store/auth'
 import { useFeatureStore } from '../store/feature'
@@ -40,6 +41,7 @@ const appBotMenuItem: MenuItem = {
 }
 
 const tailMenuItems: MenuItem[] = [
+  { key: '/system-setting', icon: <SettingOutlined />, label: '系统配置' },
   { key: '/backup', icon: <CloudUploadOutlined />, label: '备份管理' },
   { key: '/download', icon: <DownloadOutlined />, label: '下载配置' },
 ]
--- a/src/pages/SystemSetting/index.tsx
+++ b/src/pages/SystemSetting/index.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useState } from 'react'
 import {
+  Alert,
   Button,
   Card,
   Col,
@@ -10,6 +11,7 @@ import {
   Select,
   Space,
   Tag,
+  Tooltip,
   message,
 } from 'antd'
 import {
@@ -180,6 +182,7 @@ export default function SystemSetting() {
   const [settings, setSettings] = useState<SystemSettingItem[]>([])
   const [hasStoredPassword, setHasStoredPassword] = useState(false)
   const [hasEffectivePassword, setHasEffectivePassword] = useState(false)
+  const [dirty, setDirty] = useState(false)
 
   const settingDescriptions = useMemo(() => {
     const map = new Map<string, string>()
@@ -218,6 +221,7 @@ export default function SystemSetting() {
         support_email_smtp: getValue(items, 'support', 'email_smtp'),
         support_email_pwd: '',
       })
+      setDirty(false)
     } catch (error) {
       message.error('获取系统配置失败: ' + (error as Error).message)
     } finally {
@@ -268,15 +272,26 @@ export default function SystemSetting() {
           刷新
         </Button>
         <div className="toolbar-spacer" />
-        <Button icon={<SendOutlined />} onClick={() => setTestModalOpen(true)}>
-          测试邮件
-        </Button>
+        <Tooltip title={dirty ? '请先保存配置后再发送测试邮件' : ''}>
+          <Button
+            icon={<SendOutlined />}
+            onClick={() => setTestModalOpen(true)}
+            disabled={dirty}
+          >
+            测试邮件
+          </Button>
+        </Tooltip>
         <Button type="primary" icon={<SaveOutlined />} onClick={handleSave} loading={saving}>
           保存配置
         </Button>
       </div>
 
-      <Form form={form} layout="vertical" initialValues={initialValues}>
+      <Form
+        form={form}
+        layout="vertical"
+        initialValues={initialValues}
+        onValuesChange={() => setDirty(true)}
+      >
         <Row gutter={[16, 16]}>
           <Col xs={24} lg={12}>
             <Card title="注册配置" loading={loading} extra={<UserAddOutlined style={{ color: 'var(--a-text-tertiary)' }} />}>
@@ -359,11 +374,20 @@ export default function SystemSetting() {
         title="发送测试邮件"
         open={testModalOpen}
         onOk={handleTestEmail}
-        onCancel={() => setTestModalOpen(false)}
+        onCancel={() => {
+          setTestModalOpen(false)
+          testForm.resetFields()
+        }}
         okText="发送"
         cancelText="取消"
         confirmLoading={testing}
       >
+        <Alert
+          type="info"
+          showIcon
+          message="测试邮件将使用当前已保存的配置发送。如刚刚修改了配置，请先点击「保存配置」。"
+          style={{ marginBottom: 16 }}
+        />
         <Form form={testForm} layout="vertical">
           <Form.Item
             name="to"
--- a/src/pages/SystemSetting/index.tsx
+++ b/src/pages/SystemSetting/index.tsx
@@ -1,0 +1,382 @@
+import { useEffect, useMemo, useState } from 'react'
+import {
+  Button,
+  Card,
+  Col,
+  Form,
+  Input,
+  Modal,
+  Row,
+  Select,
+  Space,
+  Tag,
+  message,
+} from 'antd'
+import {
+  MailOutlined,
+  ReloadOutlined,
+  SaveOutlined,
+  SendOutlined,
+  UserAddOutlined,
+} from '@ant-design/icons'
+import {
+  SECRET_MASK,
+  getSystemSettings,
+  testSystemSettingEmail,
+  updateSystemSettings,
+  type SystemSettingItem,
+  type SystemSettingUpdateItem,
+} from '../../api/system-setting'
+
+type BoolFormValue = '' | '0' | '1'
+
+interface SystemSettingFormValues {
+  register_off: BoolFormValue
+  register_only_china: BoolFormValue
+  register_username_on: BoolFormValue
+  register_email_on: BoolFormValue
+  support_email: string
+  support_email_smtp: string
+  support_email_pwd: string
+}
+
+interface TestEmailFormValues {
+  to: string
+}
+
+interface BoolSettingDef {
+  name: keyof SystemSettingFormValues
+  category: 'register'
+  key: 'off' | 'only_china' | 'username_on' | 'email_on'
+  label: string
+  description: string
+  onLabel: string
+  offLabel: string
+}
+
+const boolSettings: BoolSettingDef[] = [
+  {
+    name: 'register_off',
+    category: 'register',
+    key: 'off',
+    label: '注册入口',
+    description: '是否关闭注册',
+    onLabel: '关闭注册',
+    offLabel: '允许注册',
+  },
+  {
+    name: 'register_only_china',
+    category: 'register',
+    key: 'only_china',
+    label: '手机号范围',
+    description: '仅中国手机号可以注册',
+    onLabel: '仅中国手机号',
+    offLabel: '不限制地区',
+  },
+  {
+    name: 'register_username_on',
+    category: 'register',
+    key: 'username_on',
+    label: '用户名注册',
+    description: '是否开启用户名注册',
+    onLabel: '开启',
+    offLabel: '关闭',
+  },
+  {
+    name: 'register_email_on',
+    category: 'register',
+    key: 'email_on',
+    label: '邮箱注册/登录',
+    description: '是否开启邮箱注册/登录',
+    onLabel: '开启',
+    offLabel: '关闭',
+  },
+]
+
+const settingMapKey = (category: string, key: string) => `${category}.${key}`
+
+const initialValues: SystemSettingFormValues = {
+  register_off: '',
+  register_only_china: '',
+  register_username_on: '',
+  register_email_on: '',
+  support_email: '',
+  support_email_smtp: '',
+  support_email_pwd: '',
+}
+
+function getValue(items: SystemSettingItem[], category: string, key: string) {
+  return items.find((item) => item.category === category && item.key === key)?.value ?? ''
+}
+
+function getSetting(items: SystemSettingItem[], category: string, key: string) {
+  return items.find((item) => item.category === category && item.key === key)
+}
+
+function normaliseBoolValue(value: string): BoolFormValue {
+  if (value === '1' || value === 'true' || value === 'TRUE') return '1'
+  if (value === '0' || value === 'false' || value === 'FALSE') return '0'
+  return ''
+}
+
+function boolText(value: string | undefined, item: BoolSettingDef) {
+  const normalised = normaliseBoolValue(value ?? '')
+  if (normalised === '1') return item.onLabel
+  if (normalised === '0') return item.offLabel
+  return '未配置'
+}
+
+function boolDefaultLabel(def: BoolSettingDef, item?: SystemSettingItem) {
+  return item?.effective_value ? `跟随默认配置（当前：${boolText(item.effective_value, def)}）` : '跟随默认配置'
+}
+
+function stringExtra(description: string, item?: SystemSettingItem) {
+  if (!item?.effective_value) return description
+  const source = item.configured ? 'DB 配置' : '默认配置'
+  return `${description}。当前生效：${item.effective_value}（${source}）`
+}
+
+function encryptedExtra(description: string, item?: SystemSettingItem) {
+  if (!item?.effective_value) return description
+  const source = item.configured ? 'DB 配置' : '默认配置'
+  return `${description}。当前生效：已配置（${source}）`
+}
+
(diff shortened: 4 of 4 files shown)


Jerry-Xin (dismissed):
No blocking issues found; the PR is relevant to `octo-admin`, matches the referenced system-setting API contract, and builds successfully.

✅ Highlights

- The new route is correctly scoped under the existing super-admin route tree in `src/App.tsx:103` and `src/App.tsx:115`.
- Sidebar wiring is consistent with existing admin navigation in `src/layouts/MainLayout.tsx:43`.
- Encrypted SMTP password handling is aligned with the backend sentinel contract: empty input preserves existing DB password by sending `****` only when a stored password exists in `src/pages/SystemSetting/index.tsx:145`.
- Boolean settings preserve the “follow default config” state while showing the effective runtime value in `src/pages/SystemSetting/index.tsx:129`.

Verification performed:

- `npm ci`
- `npm run build` passed. Vite emitted the existing large chunk warning, but no TypeScript or production build errors.

yujiawei (dismissed):
# Code Review — PR #20 (octo-admin)

## Summary

Adds a super-admin "System Setting" page that drives the new DB-overridable registration + SMTP runtime settings introduced by `octo-server#76`. The wire-up (route, sidebar entry, API client, page) is small, scoped, and consistent with the existing admin pages. The contract against the backend response (`configured` / `value` / `effective_value` / mask sentinel `****`) is correctly modelled and the encrypted-secret round-trip behaves correctly across the four states (no value / yaml fallback / DB stored / user-entered new password).

No P0 / P1 issues. Verdict: **APPROVED** with a small list of P2 UX gaps + nits that are worth a follow-up but should not block merge.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Route registered under super-admin guard | ✅ | `src/App.tsx:115` is inside the `SuperOnlyRoute` block opened at line 102 |
| Sidebar entry wired | ✅ | `src/layouts/MainLayout.tsx:44` adds `/system-setting` to `tailMenuItems`; icon imported at line 19 |
| API client matches backend contract | ✅ | `src/api/system-setting.ts:5-22` mirrors the `systemSettingItemResp` shape (`configured`, `value`, `effective_value`, `value_type`) emitted by `modules/common/api_manager_system_setting.go` in server PR #76 |
| `SECRET_MASK = '****'` matches server `secretMask` | ✅ | `src/api/system-setting.ts:31` ↔ `api_manager_system_setting.go:17` |
| Encrypted password "preserve existing" path | ✅ | `src/pages/SystemSetting/index.tsx:166` sends `SECRET_MASK` when input is empty and a DB-stored password exists; backend `settingTypeEncrypted` branch skips on `""` or `****` (`api_manager_system_setting.go` validation block) |
| Encrypted password "new value" path | ✅ | Non-empty input goes through unchanged, backend re-encrypts via `encryptKey` |
| Bool "clear back to yaml" path | ✅ | Select option with `value=''` is sent verbatim; `normaliseBool("")` returns `("", true)` on the server and the upsert clears the DB row's effective value |
| Bool legacy spellings accepted from GET | ✅ | `normaliseBoolValue` at lines 116-120 handles `true/false/TRUE/FALSE` defensively in addition to `0/1` |
| Form starts empty + cards show skeleton during initial load | ✅ | `initialValues` at line 98 + `Card loading={loading}` on both panels |
| Email field validates as email when non-empty | ✅ | `rules=[{ type: 'email', ... }]` at line 314, no `required`, so empty is allowed |
| Test email modal requires recipient + format | ✅ | Lines 371-374 |
| TS strict / no-unused-locals passes by inspection | ✅ | Types narrowed; no obvious unused imports |
| No secrets logged client-side | ✅ | Password value never read into local state outside the form instance |

## 2. Findings

### P2 — UX: no way to clear a stored SMTP password from the UI

`src/pages/SystemSetting/index.tsx:163-167`

```ts
{
  category: 'support',
  key: 'email_pwd',
  value: values.support_email_pwd || (keepExistingPassword ? SECRET_MASK : ''),
},
```

Combined with the server's encrypted-update rule ("`""` or `****` preserves the existing ciphertext"), there is no UI affordance that lets an admin **delete** a DB-stored SMTP password and fall back to the yaml default. `Input.Password allowClear` looks like it should do this, but clearing the field results in `""` being sent, which the backend explicitly treats as "do not touch". To actually unset, the admin has to bypass the UI and `DELETE` the row in the DB.

This is a real config-lifecycle gap, not just a nit. Suggestion: either (a) add a dedicated "清除已存密码 / 恢复默认" link/button under the password field that submits a sentinel value the backend understands as "clear", or (b) treat clearing of a previously-stored value (when `hasStoredPassword === true` and user submits empty) as an explicit clear and add a Popconfirm so it isn't accidental. Option (a) needs a server-side counterpart in #76 (a `__CLEAR__`-style sentinel, or a sibling `DELETE` endpoint), so it's reasonable to land this PR as-is and follow up.

### P2 — UX: "测试邮件" silently uses currently-saved config, not the values on screen

`src/pages/SystemSetting/index.tsx:246-258` calls `testSystemSettingEmail` directly. The server endpoint sends through the in-memory `SystemSettings` snapshot, which only refreshes on save (or via the background 60s reload). If an admin changes the SMTP host / password in the form and clicks **测试邮件** before clicking **保存配置**, the test exercises the *old* config and they will be misled about why it failed (or worse, succeeded with stale credentials).

Cheap mitigation, no backend change required: in the test-email modal body, add a one-line hint such as "测试将使用已保存的配置；请先点击 *保存配置* 再发送测试" and/or disable the **测试邮件** button while the form is dirty (`form.isFieldsTouched()` or a local `dirty` flag set in an `onValuesChange`). Long-term fix is to let the test endpoint accept an inline override payload.

### P3 — `validateFields()` rejections are not caught and surface as unhandled rejections

`src/pages/SystemSetting/index.tsx:232-233` and `247-248`

```ts
const values = await form.validateFields()
setSaving(true)
try {
  ...
} catch (error) {
  message.error('保存失败: ' + (error as Error).message)
}
```

The `await form.validateFields()` is **outside** the `try`. AntD rejects with a `ValidateErrorEntity` (not an `Error`), so when validation fails:
- the rejection escapes `handleSave` and bubbles up as an unhandled promise rejection in the console,
- the `(error as Error).message` cast in the catch would have produced `undefined` if it had been caught (it isn't).

Antd still displays the per-field error visually, so the user-facing UX is OK, but the console noise is real. Wrap the `validateFields()` call in the same try/catch, or short-circuit:

```ts
const handleSave = async () => {
  let values: SystemSettingFormValues
  try {
    values = await form.validateFields()
  } catch {
    return // antd has already highlighted the offending fields
  }
  setSaving(true)
  try { ... } catch (error) {
    message.error(`保存失败: ${(error as Error).message ?? '未知错误'}`)
  } finally { setSaving(false) }
}
```

Same shape applies to `handleTestEmail`.

### P3 — Test-email modal does not reset state on cancel

`src/pages/SystemSetting/index.tsx:362-365`

```tsx
onCancel={() => setTestModalOpen(false)}
```

`testForm` retains the last-entered recipient. If the admin cancels mid-typing and reopens, the previous (possibly partial) value is still there. Small but easy: call `testForm.resetFields()` in `onCancel` for symmetry with the success branch (line 253).

### P3 — `BoolSettingDef.category` is locked to the literal `'register'`

`src/pages/SystemSetting/index.tsx:47-55`

```ts
interface BoolSettingDef {
  ...
  category: 'register'
  key: 'off' | 'only_china' | 'username_on' | 'email_on'
  ...
}
```

Today every bool lives under `register`, so this is fine, but as soon as a future bool setting lands under any other category (`feature`, `space`, ...) this type narrows you into either widening the literal or duplicating the array. Consider `category: string` with the key/category pair being looked up against the server-returned schema; or move the bool-vs-string switch to be data-driven from `schema[]`. Not a blocker for this PR — flagging for the next iteration.

### Nit — `getValue` fallback in `hasStoredPassword` is redundant

`src/pages/SystemSetting/index.tsx:206-208`

```ts
const storedPassword =
  passwordSetting?.configured === true || getValue(items, 'support', 'email_pwd') === SECRET_MASK
```

Given the server contract, `configured === true` implies `value === '****'` for encrypted columns and vice versa — one of these two clauses is dead. Defensive coding is fine; just calling it out so a reader doesn't go hunting for the case where they diverge.

### Nit — `useEffect(() => { fetchSettings() }, [])` exhaustive-deps

If `eslint-plugin-react-hooks` ever gets turned on for this repo, this line will warn. Matches the pre-existing pattern in other admin pages (`Backup`, `Download`), so no change recommended in this PR.

## 3. Additional Observations (not blocking)

- **Empty-string writes on save**: every save batch sends a row for every field — even fields the admin never touched. For bool/string types backend stores `""`, which behaves equivalently to "no row" via the fallback chain but does create DB churn and a row whose `configured=false` is determined by emptiness rather than absence. Functionally fine; potentially confusing during DB introspection. A future optimisation: only include items whose form value changed since the last load.
- **Sidebar placement**: the new entry sits at the top of `tailMenuItems` (above 备份管理 and 下载配置). Looks intentional given that this is "系统配置" and the others are operational tools. No objection.
- **Security**: SMTP password travels over the existing authenticated `/api` axios instance; `autoComplete="new-password"` prevents browser autofill into wrong field; no `dangerouslySetInnerHTML`; no plaintext is ever stored in component state — all good.
- **CSS classes**: `page-title`, `page-subtitle`, `toolbar`, `toolbar-spacer` are all defined in `src/styles/admin.css` (lines 133, 142, 613, 636). No missing styles.
- **`npm run build`** was reported as passing in the PR body; the page code is self-contained and the App/Layout deltas are minimal so a regression there is unlikely.

## 4. Recommendation

Ship it. The two P2 items (no-clear-password affordance, test-email-uses-stale-config) are worth filing as follow-ups but neither is a correctness or security blocker — both will become more pressing once an admin actually rotates SMTP credentials in production. P3 nits are author's discretion.

lml2468 (dismissed):
Summary: Clean, well-structured system-setting admin page. Route scoped under super-admin guard, API client matches the `octo-server#76` contract, encrypted password sentinel logic (`****`) is correct, form UX is solid. No blocking issues.

Agree with yujiawei's P2 findings (no-clear-password UX gap + test-email-uses-stale-config); both are real but non-blocking follow-ups.

💬 Non-blocking (one additional observation):

- `src/pages/SystemSetting/index.tsx:232,247` — `form.validateFields()` sits outside the `try/catch` block in both `handleSave` and `handleTestEmail`. AntD rejects with `ValidateErrorEntity` on validation failure → unhandled rejection in console. Not user-visible (field errors still render), but noisy. Wrap it or short-circuit with a bare `catch { return }`. (Same as yujiawei P3 — confirming independently.)

- `src/api/system-setting.ts:20-25` — `SystemSettingSchemaItem` and `SystemSettingResponse.schema` are typed but never consumed by the page. Currently dead code. Fine to keep for a future data-driven iteration, but worth a `// TODO` comment so it doesn't confuse the next reader.

✅ Highlights:

- Password sentinel round-trip is correct across all four states (no value / yaml fallback / DB stored / new input)
- Boolean settings expose "follow default config" while showing effective runtime value — good admin UX pattern
- Consistent with existing page patterns (backup, download) in API style, form layout, and route wiring
- `autoComplete="new-password"` + `Input.Password` prevents browser credential autofill into wrong context

LGTM.

yujiawei (approved):
# Code Review — PR #20 (octo-admin)

Scope: independent review of the new System Setting admin page (registration policy + SMTP runtime settings) at head `5ced9f65`.

## Summary

The PR adds a single page (`src/pages/SystemSetting/index.tsx`, ~406 lines), a thin API client (`src/api/system-setting.ts`), one route, and one sidebar entry. Code is well-structured, follows the conventions of the existing pages (`Backup`, `Download`), and the form/state handling looks correct. No P0 or P1 issues found that should block merge. A handful of P2 nits and one UX gap (clearing a stored SMTP password) are worth addressing — either in this PR or a follow-up.

## 1. Verification checklist

- ✅ **Routing & menu wiring** — `src/App.tsx:13`, `src/App.tsx:115`, `src/layouts/MainLayout.tsx:19`, `src/layouts/MainLayout.tsx:44` cleanly add the route inside the existing super-admin section and place the menu entry in `tailMenuItems` next to Backup/Download. Sidebar visibility correctly follows the `SuperOnlyRoute` gate already in place.
- ✅ **API layer** — `src/api/system-setting.ts:34-41` matches the existing `axios` client conventions (no headers/auth boilerplate; the shared interceptor handles token + 401). Endpoints (`/v1/manager/common/system_setting`, `/test_email`) are namespaced consistent with sibling manager APIs.
- ✅ **Form initialization & refetch** — `fetchSettings` in `src/pages/SystemSetting/index.tsx:203-230` resets `form.setFieldsValue` and clears `dirty` on each load and after save. `onValuesChange` only fires for user interaction (not programmatic `setFieldsValue`), so the dirty flag is reliable.
- ✅ **Test-email guard** — the "测试邮件" button is disabled while `dirty=true` and the Modal carries an explicit `Alert` reminding admins that the test uses the last-saved config. This matches the fix in the second commit (`fix(system-setting): prevent test email using stale config`).
- ✅ **XSS / injection safety** — backend `description` and `effective_value` strings are rendered through `Form.Item extra` and `<Tag>` as React text nodes; no `dangerouslySetInnerHTML`. Safe.
- ✅ **TypeScript strictness** — `tsconfig.app.json` enforces `strict`, `noUnusedLocals`, `noUnusedParameters`; all new symbols are either used or re-exported.
- ⚠️ **Password update semantics** — clearing the masked password field with `allowClear` does not clear the stored secret; see Finding #2.
- ⚠️ **Backend contract coupling** — depends on `Mininglamp-OSS/octo-server#76` exposing `configured`, `effective_value`, and treating `****` as "do not change" and `""` as "clear DB override". The page assumes this end-to-end; verify the server PR returns those exact semantics before/at merge time.

## 2. Findings

### P2-1 — `SystemSettingResponse.schema` / `SystemSettingSchemaItem` are dead weight

`src/api/system-setting.ts:18-28` defines `SystemSettingSchemaItem` and includes `schema: SystemSettingSchemaItem[]` on the response, but nothing in the page reads it. The page instead hardcodes `boolSettings` (`src/pages/SystemSetting/index.tsx:59-96`) and falls back to `item.description` from the `items[]` payload.

If the intent is to drive the UI from the server-declared schema (which would prevent the hardcoded key list from drifting from backend), use it. Otherwise drop the type — `noUnusedLocals` is enforced, and an unused exported type is just future maintenance debt. Either direction is fine; just don't leave it half-wired.

### P2-2 — UI cannot clear a stored SMTP password (only replace it)

`src/pages/SystemSetting/index.tsx:166-169` in `valuesToPayload`:

```ts
{
  category: 'support',
  key: 'email_pwd',
  value: values.support_email_pwd || (keepExistingPassword ? SECRET_MASK : ''),
}
```

When `hasStoredPassword` is true and the admin uses `Input.Password`'s `allowClear` "X" button to wipe the field, `values.support_email_pwd` becomes `''`, and the payload sends `****` (SECRET_MASK) — which the backend interprets as "keep". Net effect: there is no way through this UI to delete a previously stored SMTP password and revert to the default/empty state. The same field for `support.email` and `support.email_smtp` works correctly (empty string is sent and the DB override is cleared) — the asymmetry is surprising.

Suggested fix: an explicit "清除已存密码" affordance (button or checkbox next to the `DB 已配置` tag) that, when toggled, forces the payload to send `''` for `email_pwd` regardless of the input field. Document this in the `extra` text.

### P2-3 — `SECRET_MASK = '****'` literal collides with a user-typed password

If an admin types the literal string `****` into the password field, `values.support_email_pwd || ...` is truthy and the payload sends `****` — which the backend then interprets as "do not change". Not a realistic password choice, but the sentinel-in-band design has the standard footgun. Optionally add a guard: if the entered value strictly equals `SECRET_MASK` after trim, surface a validation error ("此值为占位符，请输入真实密码"). Low priority; flag-only.

### P2-4 — `useEffect` exhaustive-deps

`src/pages/SystemSetting/index.tsx:232-234`:

```ts
useEffect(() => {
  fetchSettings()
}, [])
```

`fetchSettings` is referenced inside the effect but missing from the dependency array. The repo doesn't run ESLint in CI (only `tsc && vite build`), so this is silent — but it's still the kind of thing that bites later when someone introduces eslint-plugin-react-hooks. Either inline the call into the effect, wrap `fetchSettings` in `useCallback`, or add an `eslint-disable-next-line react-hooks/exhaustive-deps` with a brief reason.

### P2-5 — `BoolFormValue` payload omits Number/falsy edge case explicitly

`src/pages/SystemSetting/index.tsx:151`: `value: values[item.name] || ''` works because `'0'` and `'1'` are truthy strings. Just confirming this is intentional — the `''` (default) value correctly serializes to an empty string the backend can treat as "clear DB override". Consider a nullish check (`values[item.name] ?? ''`) to be type-honest, but functionally equivalent.

## 3. Recommendations

1. Decide on `SystemSettingResponse.schema`: either consume it (drive `boolSettings` from server) or remove the unused type. Leaving it as-is invites drift.
2. Add a "clear stored password" affordance so admins can actually remove the SMTP secret through the UI. This is the only setting where the cleared-field path differs from the other text inputs.
3. (Optional) Add a guard against the user literally entering `****` in the password field, or change the sentinel to something less likely to collide (e.g., absent field, or a dedicated `keep_existing: true` flag in the request body once the backend can support it).
4. Coordinate merge with `Mininglamp-OSS/octo-server#76`; the page hard-depends on the `configured` + `effective_value` + `****` semantics. If that server PR's contract changes, the password "DB 已配置" / "默认已配置" tags and `hasStoredPassword` detection will silently misbehave.

## 4. Additional observations

- The `Tooltip` wrapping the disabled "测试邮件" `Button` (`index.tsx:275-283`) works in antd 5 because antd wraps disabled buttons internally — fine, just noting since this pattern is a frequent source of "my tooltip doesn't show" reports.
- The page intentionally has no tests; PR description lists only `npm run build`. Repo convention matches (no test runner configured), so this is consistent — flagging for context, not as a request.
- `getSetting` is defined locally and used only once inside `fetchSettings` for the password lookup. Could be inlined, but it does improve readability — leaving as-is is fine.
- All UI strings are Chinese, consistent with the rest of the admin codebase. No i18n framework in play.

## Verdict

No correctness, security, or data-loss issues. The data-loss-shaped concern is Finding #2 (cannot clear a stored password through the UI), but it's recoverable (admin can overwrite, or operate directly on the DB) and not a regression from prior behavior (the page is net-new). Approving with the suggestions above to be addressed in this PR or a fast follow-up.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-admin`, wires a relevant super-admin system settings page into the existing admin UI, and the implementation aligns with the backend API contract from the related server PR.

💬 Non-blocking
- 🔵 Suggestion: [src/pages/SystemSetting/index.tsx](/private/tmp/pr-review-9sa0UR/repo/src/pages/SystemSetting/index.tsx:203) could guard against overlapping refresh/save requests or stale responses if admins click refresh while a save is in flight. The current behavior is unlikely to corrupt data because the backend owns persistence, but the UI state can briefly reflect whichever request finishes last.
- 🔵 Suggestion: [src/pages/SystemSetting/index.tsx](/private/tmp/pr-review-9sa0UR/repo/src/pages/SystemSetting/index.tsx:334) does not validate the SMTP server format client-side. Backend currently treats it as an unconstrained string, and the test-email flow can catch mistakes, so this is not blocking.
- 🟡 Warning: No frontend tests were added for the password-preservation behavior in [src/pages/SystemSetting/index.tsx](/private/tmp/pr-review-9sa0UR/repo/src/pages/SystemSetting/index.tsx:147). Given the sensitivity of encrypted placeholder handling, a focused unit test around `valuesToPayload` would reduce regression risk.

✅ Highlights
- The route is correctly nested under the existing super-admin layout in [src/App.tsx](/private/tmp/pr-review-9sa0UR/repo/src/App.tsx:100).
- The SMTP password field avoids rendering secret material and preserves existing encrypted values with the backend’s `****` sentinel contract.
- Bool settings preserve the “follow default config” state cleanly by posting empty values.
- `npm run build` passes after installing dependencies.

lml2468 (commented):
## Independent Review — PR #20 (齐静春 📖)

**Verdict: APPROVE** — no blocking issues. The second commit (`fix: prevent test email using stale config`) is a well-targeted UX improvement.

### Incremental assessment (vs my previous review at `5435bd3d`)

The new commit adds:
- `dirty` state tracking with `onValuesChange` + reset on `fetchSettings`
- Disabled "测试邮件" button with Tooltip when form has unsaved changes
- `Alert` in test modal clarifying it uses saved config
- `testForm.resetFields()` on modal cancel

All well-implemented. Specifically verified:
- `onValuesChange` in antd 5 does NOT fire for programmatic `form.setFieldsValue()`, so the dirty flag is reliable (only user edits trigger it)
- After successful save, `fetchSettings()` resets `dirty=false` — correct
- After failed save, `dirty` remains true — correct

### Previous findings still present (non-blocking)

**1. `validateFields()` outside `try/catch` — both `handleSave` (line ~237) and `handleTestEmail` (line ~252)**

```ts
const handleSave = async () => {
  const values = await form.validateFields() // ← outside try/catch
  setSaving(true)
  try {
```

If validation fails (e.g., invalid email format), the async function throws an unhandled rejection. In practice, antd shows inline validation errors and the rejection is typically swallowed by React's error boundary or the browser, but it still fires an `unhandledrejection` event that will show in console and trigger error monitoring (Sentry, etc.).

Fix: move `validateFields()` inside the `try` block, or wrap in its own `try/catch` with early return.

**2. `SystemSettingSchemaItem` type in `api/system-setting.ts:18-28` — still dead code** (also flagged by yujiawei)

Defined and included in `SystemSettingResponse` but never consumed. `boolSettings` is hardcoded in the component. Either drive the form from the server schema or remove it.

### Agreements with other reviewers

- yujiawei's P2-2 (cannot clear stored password) is a genuine UX gap — no path to send `""` for `email_pwd` once `hasStoredPassword` is true
- yujiawei's P2-3 (`****` sentinel collision) is a real footgun but very low probability
- Jerry-Xin's SMTP format validation suggestion is reasonable but non-blocking given the test-email flow catches mistakes

### Highlights

- Clean form state management — the `dirty` → disable test → save → refetch → clear dirty cycle is exactly right
- Password preservation logic is solid: empty field + stored password → `****` sentinel → backend keeps existing
- `autoComplete="new-password"` on the SMTP password field prevents browser autofill interference
- Code is consistent with existing page conventions (Backup, Download)

No blockers. Good to merge.

PR #21, opened 2026-05-19
feat(ci): refine CI + add stale and PR labeler

lml2468:
## Changes

- **ci.yml**: draft PRs are now skipped; docs/markdown-only changes no longer trigger full CI
- **stale.yml**: four-lane stale management (unassigned/assigned × issues/PRs) via shared reusable workflow
- **labeler.yml**: auto size labels (XS/S/M/L/XL) + dependency-changed detection + maintainer checklist comment

## Dependencies

Requires `Mininglamp-OSS/.github#feat/automation-p0-shared-workflows` (PR #25) to be merged first — the reusable workflows `reusable-stale.yml` and `reusable-pr-labeler.yml` must be on `main` for stale and labeler to be functional.

## Part of P0 automation rollout

Modeled after openclaw best practices.

PR #22, opened 2026-05-19
fix(ci): replace negation filter with positive-pattern filter

lml2468:
## Summary

Replace the negation-based `dorny/paths-filter` filter with explicit positive patterns, consistent with `octo-lib` fix merged in #16.

**Previous (negation-based, flagged as ambiguous):**
```yaml
filters: |
  code:
    - '**'
    - '!**/*.md'
    - '!docs/**'
    - '!.github/ISSUE_TEMPLATE/**'
```

**New (positive patterns, unambiguous):**
```
src/**, public/**, package.json, tsconfig*.json, *.config.*, .github/workflows/**
```

**Why positive patterns:**
- No negation semantics ambiguity with `some` predicate
- Easy to reason about: CI runs when ≥1 changed file matches a source pattern
- Docs-only PRs touch none of these → `code=false` → jobs skipped ✅

**Related:** octo-lib PR #16 (same fix, already reviewed and merged)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,10 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # For push events the paths-filter step is skipped; default to 'true'
-      # so downstream jobs always run on direct main pushes.
-      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+      # docs_only=true means ALL changed files are documentation.
+      # For push events the paths-filter step is skipped; default to 'false'
+      # (not docs_only) so downstream jobs always run on direct main pushes.
+      docs_only: ${{ github.event_name == 'pull_request' && steps.filter.outputs.docs_only || 'false' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         if: github.event_name == 'pull_request'
@@ -42,24 +43,28 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Positive patterns only — no negation, no predicate-quantifier needed.
-          # code=true when at least one changed file matches a source pattern.
-          # Docs-only PRs won't match any pattern -> code=false -> jobs skipped.
-          # (skipped jobs count as passing for required-check branch protection).
+          # predicate-quantifier: every means ALL changed files must satisfy the
+          # docs_only filter for it to be true. So docs_only=true only when every
+          # changed file is a markdown/doc/template file.
+          #
+          # This is the inverse pattern: instead of listing every code file type
+          # (impossible to enumerate completely), we only define what counts as
+          # pure documentation. Anything else — .golangci.yml, Dockerfile,
+          # turbo.json, package-lock.json, etc. — automatically triggers CI
+          # because it won't match docs_only, so docs_only becomes false.
+          predicate-quantifier: 'every'
           filters: |
-            code:
-              - 'src/**'
-              - 'public/**'
-              - 'package.json'
-              - 'tsconfig*.json'
-              - '*.config.*'
-              - '.github/workflows/**'
+            docs_only:
+              - '**/*.md'
+              - 'docs/**'
+              - '.github/ISSUE_TEMPLATE/**'
+
 
   build:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+      needs.changes.outputs.docs_only != 'true'
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,17 +42,18 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Negation patterns (! prefix) correctly exclude matched files even
-          # when '**' would otherwise match them — a .md file matches '**' but
-          # is then eliminated by '!**/*.md', leaving code: false for docs-only
-          # PRs. This is standard minimatch behaviour; predicate-quantifier
-          # defaults to 'any' which is exactly what we want.
+          # Positive patterns only — no negation, no predicate-quantifier needed.
+          # code=true when at least one changed file matches a source pattern.
+          # Docs-only PRs won't match any pattern -> code=false -> jobs skipped.
+          # (skipped jobs count as passing for required-check branch protection).
           filters: |
             code:
-              - '**'
-              - '!**/*.md'
-              - '!docs/**'
-              - '!.github/ISSUE_TEMPLATE/**'
+              - 'src/**'
+              - 'public/**'
+              - 'package.json'
+              - 'tsconfig*.json'
+              - '*.config.*'
+              - '.github/workflows/**'
 
   build:
     needs: [changes]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,11 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # code=true when at least one changed file is not a documentation file.
-      # For push events the paths-filter step is skipped; default to 'true'
-      # so downstream jobs always run on direct main pushes.
+      # code=true when at least one changed file matches the code filter.
+      # predicate-quantifier:every + negation patterns (no bare '**'):
+      # a file matches 'code' iff ALL negation patterns are satisfied
+      # (file is not a doc/template). code=false only when all changed files
+      # are docs/templates. Push events default to 'true'.
       code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -43,16 +45,22 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Negation patterns with default 'some' quantifier (correct approach).
-          # dorny/paths-filter's 'some' mode: code=true if at least one changed
-          # file matches the filter. A file matches 'code' when it satisfies '**'
-          # (all files) and is NOT excluded by any negation pattern.
-          # Result: code=false only when ALL changed files are docs/templates.
-          # Source: filter.ts:isMatch() uses patterns.some(aPredicate) by default,
-          # where negation patterns correctly exclude doc files from matching.
+          # predicate-quantifier: every — a file matches 'code' only when ALL
+          # negation patterns are satisfied (file is NOT a doc/template file).
+          #
+          # Validated by picomatch v4 unit test (same library used internally):
+          #   README.md:    !**/*.md=false → every=false → not match → code=false ✅
+          #   main.go:      all negations=true → every=true → match → code=true ✅
+          #   mixed PR:     main.go matches → code=true (some over changed files) ✅
+          #   .golangci.yml: all negations=true → every=true → CI runs ✅
+          #   Dockerfile:   all negations=true → CI runs ✅
+          #
+          # Do NOT add '**': picomatch('**') doesn't match dotfiles (.golangci.yml)
+          # causing them to incorrectly skip CI with the 'every' quantifier.
+          # Do NOT use 'some' + negation: '**' short-circuits some to true always.
+          predicate-quantifier: 'every'
           filters: |
             code:
-              - '**'
               - '!**/*.md'
               - '!docs/**'
               - '!.github/ISSUE_TEMPLATE/**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,10 +30,10 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # docs_only=true means ALL changed files are documentation.
-      # For push events the paths-filter step is skipped; default to 'false'
-      # (not docs_only) so downstream jobs always run on direct main pushes.
-      docs_only: ${{ github.event_name == 'pull_request' && steps.filter.outputs.docs_only || 'false' }}
+      # code=true when at least one changed file is not a documentation file.
+      # For push events the paths-filter step is skipped; default to 'true'
+      # so downstream jobs always run on direct main pushes.
+      code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         if: github.event_name == 'pull_request'
@@ -43,28 +43,26 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # predicate-quantifier: every means ALL changed files must satisfy the
-          # docs_only filter for it to be true. So docs_only=true only when every
-          # changed file is a markdown/doc/template file.
-          #
-          # This is the inverse pattern: instead of listing every code file type
-          # (impossible to enumerate completely), we only define what counts as
-          # pure documentation. Anything else — .golangci.yml, Dockerfile,
-          # turbo.json, package-lock.json, etc. — automatically triggers CI
-          # because it won't match docs_only, so docs_only becomes false.
-          predicate-quantifier: 'every'
+          # Negation patterns with default 'some' quantifier (correct approach).
+          # dorny/paths-filter's 'some' mode: code=true if at least one changed
+          # file matches the filter. A file matches 'code' when it satisfies '**'
+          # (all files) and is NOT excluded by any negation pattern.
+          # Result: code=false only when ALL changed files are docs/templates.
+          # Source: filter.ts:isMatch() uses patterns.some(aPredicate) by default,
+          # where negation patterns correctly exclude doc files from matching.
           filters: |
-            docs_only:
-              - '**/*.md'
-              - 'docs/**'
-              - '.github/ISSUE_TEMPLATE/**'
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
 
 
   build:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.docs_only != 'true'
+      needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     strategy:
       matrix:


yujiawei (changes requested):
# Code Review — PR #22 (octo-admin)

## Summary
This PR replaces the previous broad path filter plus negations with a positive allowlist for code-related changes. The intent is good, but the new allowlist misses files that are part of the current frontend build inputs, so some build-affecting PRs would now skip CI.

## Findings
- (P1) .github/workflows/ci.yml:51 — The positive allowlist does not include `package-lock.json` or `index.html`. This workflow runs `npm ci`, `npx tsc --noEmit`, and `npm run build`, so a lockfile-only dependency update or a Vite entry HTML change can affect the build but would leave `code=false`, causing the build jobs to be skipped while required checks appear satisfied. Please include these build inputs, and consider whether other root build/deploy inputs such as `Dockerfile` or `nginx.conf.template` should also be covered.

## Verdict
CHANGES_REQUESTED — The path filter under-matches build-affecting changes and can let required CI pass without running.

Jerry-Xin (changes requested):
The PR is relevant to `octo-admin`, but the new positive filter drops CI for tracked files that can affect install/build behavior.

**🔴 Blocking**
- 🔴 Critical: The positive path list is incomplete in [.github/workflows/ci.yml](/private/tmp/pr-review-SCxYZA/repo/.github/workflows/ci.yml:51). It includes `package.json` but omits `package-lock.json`, while the build job runs `npm ci` in [.github/workflows/ci.yml](/private/tmp/pr-review-SCxYZA/repo/.github/workflows/ci.yml:73). A PR that changes only `package-lock.json` would now produce `code=false` and skip CI, even though dependency resolution can directly break install, typecheck, or build. This is a regression from the previous `**` minus docs filter.
- 🔴 Critical: The filter also omits `index.html`, which is the Vite HTML entrypoint for the app. A PR that changes only `index.html` can affect the production build/runtime entry and should not be treated as docs-only.

**💬 Non-blocking**
- 🔵 Suggestion: Consider adding a lightweight path-filter validation test or documented example cases for docs-only, lockfile-only, workflow-only, and source-only PRs. This CI gate is branch-protection-sensitive, so regressions are easy to miss.

**✅ Highlights**
- The move away from negation-based `dorny/paths-filter` patterns is directionally good and easier to reason about.
- Keeping `.github/workflows/**` in the positive set is appropriate because workflow changes should exercise CI.

lml2468 (commented):
## Independent Review — PR #22 (齐静春 📖)

**Verdict: CHANGES_REQUESTED** (submitted as COMMENT due to auth constraint).

Direction is correct — positive patterns are unambiguous and easier to reason about. But the allowlist has gaps.

### 🔴 Blocking

**1. Missing `package-lock.json`** (confirming both reviewers)

The build job runs `npm ci` which resolves dependencies **exclusively** from `package-lock.json`, ignoring `package.json` ranges. A lockfile-only PR (dependabot, `npm audit fix`, manual `npm update`) changes what gets installed but would now produce `code=false` → CI skipped.

This is a regression from the previous `**` minus docs filter, which correctly caught lockfile changes.

Fix: add `package-lock.json` to the filter list.

**2. Missing `index.html`** (confirming both reviewers)

`index.html` is the Vite entry point — the `<script type="module" src="/src/main.tsx">` tag, meta tags, and any CDN links live here. Changes directly affect `npm run build` output. Currently at repo root, outside all positive patterns.

Fix: add `index.html` to the filter list.

### 💬 Non-blocking

**3. `Dockerfile` and `nginx.conf.template` — intentionally excluded, and that's correct.**

I verified the Dockerfile: it's a multi-stage build (`node:18-alpine` → `nginx:alpine`) that runs the same `npm ci && npm run build` pipeline. The CI workflow doesn't build or test the Docker image, so Dockerfile/nginx changes can't be caught by the current CI anyway. Excluding them from the code filter is the right call — if Docker build testing is added later, these should be included then.

**4. Inconsistent action pinning: `changes` job pins `actions/checkout` by SHA (`@34e114876b...`), `build` job uses tag (`@v4`).** Pre-existing from PR #21, not introduced here, but worth noting for a follow-up cleanup.

**5. The `converted_to_draft` rationale in the new comment (line 8-10) is actually a solid argument** — creating "skipped" check entries prevents stale "pending" statuses when a PR is demoted to draft. This addresses the feedback from PR #21 review. Good documentation.

### Suggested fix (two lines)

```yaml
filters: |
  code:
    - 'src/**'
    - 'public/**'
    - 'package.json'
    - 'package-lock.json'
    - 'index.html'
    - 'tsconfig*.json'
    - '*.config.*'
    - '.github/workflows/**'
```

lml2468:
## Fix pushed — switched to `docs_only` inverse filter

**Root cause of previous blocking:** positive allowlists are impossible to enumerate completely — `.golangci.yml`, `Dockerfile`, `turbo.json`, `package-lock.json`, and countless other CI-affecting files would each need to be explicitly listed. Any omission creates a CI bypass.

**New approach (inverse filter):**
```yaml
predicate-quantifier: 'every'
filters: |
  docs_only:
    - '**/*.md'
    - 'docs/**'
    - '.github/ISSUE_TEMPLATE/**'
```

Instead of listing "what is code", we only define "what is pure documentation". `predicate-quantifier: every` means `docs_only=true` only when ALL changed files match the docs_only filter.

**Truth table:**
| Scenario | docs_only | CI |
|---|---|---|
| Only `README.md` | true | skipped ✅ |
| Only `main.go` | false | runs ✅ |
| `README.md` + `main.go` | false (main.go fails) | runs ✅ |
| Only `.golangci.yml` | false | runs ✅ |
| Only `Dockerfile` | false | runs ✅ |
| Push to main | false (default) | runs ✅ |

No allowlist to maintain — anything not matching `**/*.md`, `docs/**`, or `.github/ISSUE_TEMPLATE/**` automatically triggers CI.

yujiawei (changes requested):
# Code Review — PR #22 (octo-admin)

## Summary
This PR changes the CI path filter from a negated `code` filter to a positive `docs_only` filter, then skips the build when `docs_only` is true. The intent is reasonable, but the current `predicate-quantifier: every` configuration does not behave as an OR across the listed documentation patterns.

## Findings
- (P1) .github/workflows/ci.yml:58 — `predicate-quantifier: every` requires a changed file to satisfy every pattern in the `docs_only` filter, not just one of them. With the current alternatives (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`), a normal docs-only change such as `README.md` or `docs/foo.md` will not match all rules, so `docs_only` remains `false` and the build still runs. Please restructure the filter so docs-only PRs are actually detected, for example by using a negated code-impact filter with `every` or another form that does not require each file to match all documentation alternatives.

## Verdict
CHANGES_REQUESTED — the new filter does not implement the intended docs-only detection and should be corrected before merge.

Jerry-Xin (changes requested):
The PR is in scope for `octo-admin`, but the new `docs_only` filter does not implement the intended docs-only detection.

**🔴 Blocking**
- 🔴 Critical: `.github/workflows/ci.yml:55` uses `predicate-quantifier: 'every'` with three positive patterns at `.github/workflows/ci.yml:57-60`, but `dorny/paths-filter` applies `every` to each changed file across all rules. That means one file must match `**/*.md`, `docs/**`, and `.github/ISSUE_TEMPLATE/**` simultaneously to count as `docs_only`. A docs-only PR changing `README.md` or `docs/guide.md` will therefore produce `docs_only=false`, and `.github/workflows/ci.yml:67` will run CI instead of skipping it.

A safer fix is to keep a `code` filter and use `predicate-quantifier: 'every'` with negations, so any non-doc changed file matches:

```yaml
predicate-quantifier: 'every'
filters: |
  code:
    - '**'
    - '!**/*.md'
    - '!docs/**'
    - '!.github/ISSUE_TEMPLATE/**'
```

Then keep the downstream condition on `needs.changes.outputs.code == 'true'`.

**💬 Non-blocking**
- 🔵 Suggestion: The PR description says this replaces the filter with positive source patterns, but the final branch uses an inverse `docs_only` filter. Updating the description would avoid confusion during future CI maintenance.

**✅ Highlights**
- The push-event default behavior is still intentionally preserved: direct `main` pushes continue to run downstream jobs because `docs_only` defaults to `'false'`.

lml2468:
## Fix pushed — back to negation filter (corrects docs_only+every mistake)

**What was wrong:** `predicate-quantifier: every` means each *file* must satisfy **all** patterns simultaneously (confirmed from `filter.ts:isMatch()` source: `return patterns.every(aPredicate)`). With 3 patterns (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`), no file can ever satisfy all 3 at once, so `docs_only` was always `false` → CI always ran → docs-only PRs were never skipped.

**Fix applied (matching the approach used in already-merged repos):**
```yaml
code:
  - '**'
  - '!**/*.md'
  - '!docs/**'
  - '!.github/ISSUE_TEMPLATE/**'
```
Default `some` quantifier: `code=true` when at least one changed file is not a doc file. lml2468 confirmed from source-code analysis: *"negation patterns with default some quantifier worked correctly."*

Truth table:
| Scenario | code | CI |
|---|---|---|
| Only `README.md` | false | skipped ✅ |
| Only `main.go` | true | runs ✅ |
| `README.md` + `main.go` | true | runs ✅ |
| Only `.golangci.yml` | true | runs ✅ |
| Only `Dockerfile` | true | runs ✅ |
| Push to main | true (default) | runs ✅ |

yujiawei (dismissed):
# Code Review — PR #22 (octo-admin)

## Summary
This PR updates the CI change-detection output expression for pull request versus push events and refreshes the inline comments around the `dorny/paths-filter` rules. The final diff keeps the existing negation-based filter and documents why it should mark `code=false` only for documentation/template-only changes.

## Findings
- None.

## Verdict
APPROVED — The workflow change is small, preserves the intended PR/push behavior, and I did not find a blocking correctness issue in the diff.

Jerry-Xin (changes requested):
Relevance gate: pass. This PR changes the project’s CI workflow, which is part of `Mininglamp-OSS/octo-admin`.

🔴 Blocking

🔴 Critical — The PR does not implement the advertised positive-pattern filter, and the current negation filter still makes docs-only PRs count as `code=true`.
`.github/workflows/ci.yml:46-58`

With `dorny/paths-filter` default `some` quantifier, the leading `**` pattern is enough for every changed file to match the `code` filter. A docs-only change such as `README.md` or `docs/foo.md` still matches `**`, so the later negation patterns do not make the overall filter false. This directly contradicts the PR summary and the line 33 comment saying `code=true` only when at least one changed file is not documentation.

Fix by actually replacing the filter with explicit positive source/build patterns, as described in the PR, or by using a correct `predicate-quantifier: every` configuration with the negation approach.

💬 Non-blocking

🔵 Suggestion — If switching to positive patterns, make sure the list covers all files that should trigger CI for this project, not only `src/**` and `public/**`. Consider whether `index.html`, `package-lock.json`, `Dockerfile`, `nginx.conf.template`, and relevant config files should also run CI.

✅ Highlights

The push fallback expression at `.github/workflows/ci.yml:36` remains functionally aligned with the intended behavior: push events default to `true`, while pull requests use the filter output.

lml2468:
## Final fix — predicate-quantifier:every + negation-only patterns (validated by picomatch test)

Ran picomatch v4 unit tests (same library dorny/paths-filter uses) to definitively validate all approaches.

### Test results (4 approaches, 10 scenarios each):

| Approach | Result |
|---|---|
| A: `some` + `["**", "!**/*.md", ...]` | ❌ — `**` short-circuits `some` to `true`; README.md=true |
| B: `every` + `["**/*.md", "docs/**", ...]` | ❌ — file must match ALL 3 patterns; README.md=false always |
| C: `every` + `["**", "!**/*.md", ...]` | ❌ — `picomatch("**")` does not match dotfiles; .golangci.yml skips CI |
| **D: `every` + `["!**/*.md", "!docs/**", "!.github/ISSUE_TEMPLATE/**"]`** | **✅ — ALL 10 scenarios pass** |

### Approach D (this commit):
```yaml
predicate-quantifier: 'every'
filters: |
  code:
    - '!**/*.md'
    - '!docs/**'
    - '!.github/ISSUE_TEMPLATE/**'
```

Truth table (picomatch-verified):
| Scenario | code | CI |
|---|---|---|
| Only README.md | false | skipped ✅ |
| Only main.go | true | runs ✅ |
| README.md + main.go | true | runs ✅ |
| Only .golangci.yml | true | runs ✅ |
| Only Dockerfile | true | runs ✅ |
| Push to main | true (default) | runs ✅ |

Reviewed by codex (gpt-5.5): LGTM ✅

yujiawei (approved):
# Code Review — PR #22 (octo-admin)

## Summary

The change to `.github/workflows/ci.yml` fixes the docs-only CI-skip semantics by combining `predicate-quantifier: 'every'` with a negation-only filter (no leading `'**'`). For each changed file, the `code` filter now matches only when all three negation patterns are satisfied, and `code=true` is emitted when at least one changed file is non-docs/non-template. This is a correct and minimal fix for the original `some + '**' + !patterns` bug, and avoids the completeness pitfalls of the positive-allowlist approach explored in earlier commits on this branch.

The output expression and push-event fallback are preserved:

```yaml
code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
```

— for `push` the filter step is skipped and the expression short-circuits to `'true'` (push docs-only is already filtered earlier by the workflow-level `paths-ignore` block at `.github/workflows/ci.yml:6-9`), and for `pull_request` the filter output is honored, including the `'false'` string (which is a truthy string but is returned as the second operand of `&&`, so the `||` does not override it).

## Findings

### ✅ Correctness verified for the change-detection cases that matter for this repo

I walked through the per-file semantics against the actual repo layout for `f24cb35`:

| Changed file(s)                          | `!**/*.md` | `!docs/**` | `!.github/ISSUE_TEMPLATE/**` | `every` per file | `code` output | Behavior |
|------------------------------------------|:---------:|:---------:|:---------------------------:|:----------------:|:-------------:|:--------|
| `README.md` only                          | ✗ | ✓ | ✓ | ✗ | `false` | CI skipped ✅ |
| `docs/foo.md` only                        | ✗ | ✗ | ✓ | ✗ | `false` | CI skipped ✅ |
| `.github/ISSUE_TEMPLATE/bug.yml` only     | ✓ | ✓ | ✗ | ✗ | `false` | CI skipped ✅ |
| `package-lock.json` only (dependabot)     | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ (regression vs. earlier positive-allowlist commit `1e654bb`, where this was missed) |
| `index.html` only (Vite entry)            | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ |
| `Dockerfile` only                         | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ |
| `nginx.conf.template` only                | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ |
| `vite.config.ts` / `tsconfig.json` only   | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ |
| `.golangci.yml`-style dotfile (root)      | ✓ | ✓ | ✓ | ✓ | `true`  | CI runs ✅ |
| `README.md` + `src/foo.tsx` (mixed)       | mixed per-file | — | — | — | `true`  | CI runs ✅ (at least one matching file) |

The previous reviewer concerns about an under-covering positive allowlist (`package-lock.json`, `index.html`, `Dockerfile`, `nginx.conf.template`) are addressed by construction here: anything not explicitly excluded as docs/templates participates in CI. This is the right semantic for a tight, manually-curated exclusion list.

### ✅ Push-event path

`paths-ignore` at lines 6–9 of `.github/workflows/ci.yml` mirrors the three exclusion patterns, so docs-only `main` pushes do not even create a workflow run. Non-doc pushes trigger and short-circuit `code` to `'true'`, which keeps `build` running. Consistent with the comment on the filter step.

### 💬 Non-blocking — PR body is stale (P2 / documentation nit)

The PR description still advertises a positive-pattern filter:

> **New (positive patterns, unambiguous):**
> ```
> src/**, public/**, package.json, tsconfig*.json, *.config.*, .github/workflows/**
> ```

That matches the first commit on this branch (`1e654bb`), not the final state. The actual change ships `predicate-quantifier: 'every'` plus a three-rule negation-only filter and intentionally omits `'**'`. Please update the PR description before merge so the rationale captured in the commit message and the inline comments is also discoverable from the PR landing page — future maintainers searching git history for "why every quantifier" will hit the PR body first.

(Not blocking the merge — the in-file comments at `.github/workflows/ci.yml:33-38` and `:48-61` already explain the chosen design clearly, including the two "Do NOT" footguns. The commit message `fix(ci): use predicate-quantifier:every + negation-only patterns (no **)` is also accurate.)

### 💬 Non-blocking — minor stylistic nits

- `.github/workflows/ci.yml:67-68` introduces a double blank line between `code:` filter list and the `build:` job. Single blank line would match the rest of the file; trivial.
- Pre-existing, not introduced here: the `changes` job pins `actions/checkout` by commit SHA (`@34e114876b...`) while the `build` job uses the `@v4` tag. Worth a follow-up consolidation, but out of scope for this PR.

### ✅ Things deliberately not in scope (and that's correct)

- No positive include for `Dockerfile` / `nginx.conf.template` — current CI doesn't build the Docker image, so adding them to a positive list would be misleading. With the negation-only filter they correctly run CI today; if Docker-build testing is later added, the filter does not need to be changed.
- No attempt to handle the `paths-filter` "no changed files" edge case — for `pull_request`, GitHub does not raise the event with zero changes, and the expression's `||` fallback already returns `'true'` if the filter output is empty.

## Verdict

APPROVED — the diff is correct, the chosen filter strategy is sound for this repo's actual build inputs, and the in-file comments document the picomatch footguns ("do not add `**`", "do not use `some` + negation") that motivated the change. The only follow-up is a documentation refresh of the PR body before merge.

Jerry-Xin (approved):
This PR is in scope: it modifies this repository’s CI workflow for PR path filtering, and the final negation-only `predicate-quantifier: every` logic should correctly skip docs/template-only PRs while running CI for source, config, root infra, and dotfile changes.

💬 Non-blocking
- 🔵 Suggestion: [.github/workflows/ci.yml:51](/private/tmp/pr-review-0E85qv/repo/.github/workflows/ci.yml:51) says the behavior was validated with picomatch v4, but the pinned `dorny/paths-filter` action depends on picomatch v2.3.1. The observed behavior is still correct, but the comment should ideally match the action’s actual dependency to avoid future confusion.
- 🔵 Suggestion: [.github/workflows/ci.yml:58](/private/tmp/pr-review-0E85qv/repo/.github/workflows/ci.yml:58) says `picomatch('**')` does not match dotfiles, but the pinned action passes `dot: true`, under which `**` does match dotfiles. This is comment-only and does not affect the current workflow behavior.

✅ Highlights
- Minimal workflow permissions are preserved.
- Push events still default to running CI.
- The filter now avoids the previous ambiguous `some` + broad include + negation behavior for pull requests.

PR #27, opened 2026-05-20
fix(feature-store): differentiate transient probe failures from 404

### LINKED ISSUE #19: useFeatureStore caches probe failure with same TTL as success — sticky-hides App Bot menu for 10 min after startup race
## Description

`useFeatureStore.probeAppBots()` caches **both successful and failed** probe results to `localStorage` (`dm-admin-features`) with the same 10-minute TTL. When the very first probe after a fresh login happens during a transient backend hiccup (octo-server still booting, brief 5xx, network flap), the failure is sticky for the full 10 minutes — the "应用 Bot" menu stays hidden even after the backend is healthy.

## Steps to Reproduce

1. Fresh k8s deploy: octo-server in CrashLoopBackOff for 30–60 s waiting for MySQL `sql-migrate`, then becomes Ready
2. Open `/admin/` in a browser **while** octo-server is still warming up (or right after a 5xx burst)
3. `useFeatureStore` `probeAppBots` runs once during mount, hits a 5xx or timeout, stores `{appBotsAvailable: false, appBotsCheckedAt: <now>}` into localStorage
4. Backend `/v1/app_bot/available` is healthy seconds later (verified: `curl /api/v1/app_bot/available` returns `200 {}`)
5. User reloads (Cmd+Shift+R) — **localStorage survives hard reload** — `appBotsAvailable === false && isFresh(checkedAt, 10min)` → `probeAppBots` short-circuits, menu stays hidden
6. Menu only reappears after 10 minutes (TTL expires and next mount re-probes) or after manually clearing `dm-admin-features` from DevTools → Application → Local Storage

## Expected Behavior

A transient probe failure during startup should not hide the menu for 10 minutes when the backend becomes healthy in seconds. Either:

- Don't cache the failed result, OR
- Use a much shorter TTL for failures than successes, OR
- Distinguish 404 ("route not registered, definitely unavailable") from 5xx/network ("transient, retry soon")

## Actual Behavior

Failure and success share the same write path:

```ts
// src/store/feature.ts
try {
  await api.get('/v1/app_bot/available')
  set({ appBotsAvailable: true, appBotsCheckedAt: Date.now() })
} catch {
  // 404 → 路由未注册 → 隐藏；其它错误（网络/5xx）保守同样隐藏
  set({ appBotsAvailable: false, appBotsCheckedAt: Date.now() })  // ← sticky for 10 min
}
```

The inline comment acknowledges the distinction ("404 vs 网络/5xx") but treats them identically. The "保守" (conservative) framing is right for 404 (route not registered = definitely unavailable), but for 5xx/network it's an over-cache: probe will succeed seconds later and the operator pays a 10-minute UX delay for an upgrade race.

## Version

`docker.io/mininglamposs/octo-admin:latest` (built by PR #18 `bb63762`, 2026-05-18) and the corresponding `main` HEAD of `src/store/feature.ts`.

## Affected Component

Web Frontend (`octo-admin` SPA, `src/store/feature.ts`).

## Severity

P2 — Non-core Feature Issue. App Bot management is hidden but other admin functions work; backend functionality unaffected. Annoying during first-deploy / upgrade races, easy to misdiagnose as "missing menu = backend bug".

## Reproducibility

Race condition — reproduces consistently if `probeAppBots` runs during the first 30 s after deploy when octo-server is still warming up; otherwise probe succeeds first time and the issue is invisible. Easy to force-reproduce by intentionally returning 5xx from `/v1/app_bot/available` once, then making it return 200, and observing the menu stays hidden for 10 min across hard reloads.

## Environment

- Helm-deployed octo-server (warm-up window where Pod is Ready but health-check race can return 5xx briefly)
- Browser: any (Chrome 137 tested, but localStorage behavior is universal)
- octo-admin commit: `main` HEAD as of 2026-05-18

## Logs / Screenshots

DevTools → Application → Local Storage → `<host>` → `dm-admin-features`:

```json
{
  "state": {
    "appBotsAvailable": false,
    "appBotsCheckedAt": 1779081234567
  },
  "version": 0
}
```

Reproduces every fresh deploy hit during the octo-server warm-up window. Fix is to either drop the failure cache or special-case 404.

## Suggested Fix

Minimal 3-line diff:

```diff
   try {
     await api.get('/v1/app_bot/available')
     set({ appBotsAvailable: true, appBotsCheckedAt: Date.now() })
   } catch {
-    // 404 → 路由未注册 → 隐藏；其它错误（网络/5xx）保守同样隐藏
-    // 401 已被 axios 拦截器处理为登出，此处不应到达
-    set({ appBotsAvailable: false, appBotsCheckedAt: Date.now() })
+    // 失败不写 checkedAt → 下次 probeAppBots 仍会重新尝试，避免一次启动 race
+    // 永远卡死 menu 10 分钟。401 已被 axios 拦截器处理为登出，此处不到达。
+    set({ appBotsAvailable: false })
   }
```

`appBotsCheckedAt` 保持 `0` 让 `isFresh()` 返回 `false`，下次组件 mount 时重新 probe。

或者更细：识别 404 才 cache 失败，5xx/网络错误不 cache：

```ts
} catch (e) {
  const status = (e as AxiosError)?.response?.status
  if (status === 404) {
    // 后端确定没注册路由
    set({ appBotsAvailable: false, appBotsCheckedAt: Date.now() })
  } else {
    // 5xx / 网络错 → 不 cache，下次重 probe
    set({ appBotsAvailable: false })
  }
}
```

后者更精确，但前者改动最小。

## Workaround

清浏览器 localStorage 的 `dm-admin-features` key，或等 10 分钟后刷新页面。

Happy to send a PR with the suggested fix if welcome.

an9xyz:
Closes #19

## Summary
- `useFeatureStore.probeAppBots()` cached every failure with the same 10-min TTL as success. A single 5xx/network error during a startup race (octo-server warm-up) sticky-hid the **应用 Bot** menu for 10 minutes even after the backend recovered in seconds — localStorage survives hard reload, so users had to clear `dm-admin-features` manually or wait out the TTL.
- Distinguish failure modes by HTTP status:
  - **200** → cache success, 10-min TTL (unchanged)
  - **404** → cache "unavailable", 10-min TTL (route not registered is a stable backend-version signal)
  - **5xx / network / timeout** → cache "unavailable", **30-sec transient TTL** (recover fast, but still bounded to avoid hammering on a sustained outage)
- `src/api/index.ts` response interceptor previously wrapped errors in `new Error(message)`, dropping `response.status` and making the distinction impossible. Introduced `ApiError` that preserves `status`.
- Introduced **vitest + jsdom** (no test framework existed); added 9 tests covering success caching, 404/5xx/network distinction, transient-TTL recovery, no-hammering within the transient window, force-bypass, and `resetFeatures`.

## Behavior change
| Scenario | Before | After |
|---|---|---|
| Backend warm-up race (5xx then 200) | Menu hidden 10 min | Menu reappears within 30 s |
| Old backend without route (404) | Menu hidden 10 min | Menu hidden 10 min (unchanged) |
| Sustained 5xx outage | 1 request per 10 min | 1 request per 30 s |
| Healthy backend | Caches once per 10 min | Caches once per 10 min (unchanged) |

## Test plan
- [x] `npm test` — 9/9 pass
- [x] `npm run build` — tsc + vite build clean
- [ ] Manual: set localStorage `dm-admin-features` to `{state:{appBotsAvailable:false,appBotsCheckedAt:<now>,appBotsCheckTtl:30000},version:0}`, wait 31 s, reload → menu re-probes
- [ ] Manual: simulate fresh deploy with octo-server returning 502 then 200 — verify menu reappears within 30 s of recovery

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,4 +23,5 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npx tsc --noEmit
+      - run: npm test
       - run: npm run build
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,13 @@
 {
-  "name": "dmwork-admin",
+  "name": "octo-admin",
   "version": "0.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "dmwork-admin",
+      "name": "octo-admin",
       "version": "0.1.0",
+      "license": "Apache-2.0",
       "dependencies": {
         "@ant-design/icons": "^5.6.1",
         "antd": "^5.24.6",
@@ -21,8 +22,11 @@
         "@types/react": "^18.3.20",
         "@types/react-dom": "^18.3.6",
         "@vitejs/plugin-react": "^4.4.1",
+        "@vitest/coverage-v8": "^4.1.6",
+        "jsdom": "^29.1.1",
         "typescript": "^5.8.2",
-        "vite": "^6.2.0"
+        "vite": "^6.2.0",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@ant-design/colors": {
@@ -121,6 +125,57 @@
       "peerDependencies": {
         "react": ">=16.9.0"
       }
+    },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "5.1.11",
+      "resolved": "https://registry.npmmirror.com/@asamuzakjp/css-color/-/css-color-5.1.11.tgz",
+      "integrity": "sha512-KVw6qIiCTUQhByfTd78h2yD1/00waTmm9uy/R7Ck/ctUyAPj+AEDLkQIdJW0T8+qGgj3j5bpNKK7Q3G+LedJWg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@csstools/css-calc": "^3.2.0",
+        "@csstools/css-color-parser": "^4.1.0",
+        "@csstools/css-parser-algorithms": "^4.0.0",
+        "@csstools/css-tokenizer": "^4.0.0"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/dom-selector": {
+      "version": "7.1.1",
+      "resolved": "https://registry.npmmirror.com/@asamuzakjp/dom-selector/-/dom-selector-7.1.1.tgz",
+      "integrity": "sha512-67RZDnYRc8H/8MLDgQCDE//zoqVFwajkepHZgmXrbwybzXOEwOWGPYGmALYl9J2DOLfFPPs6kKCqmbzV895hTQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@asamuzakjp/nwsapi": "^2.3.9",
+        "bidi-js": "^1.0.3",
+        "css-tree": "^3.2.1",
+        "is-potential-custom-element-name": "^1.0.1"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/generational-cache": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmmirror.com/@asamuzakjp/generational-cache/-/generational-cache-1.0.1.tgz",
+      "integrity": "sha512-wajfB8KqzMCN2KGNFdLkReeHncd0AslUSrvHVvvYWuU8ghncRJoA50kT3zP9MVL0+9g4/67H+cdvBskj9THPzg==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/nwsapi": {
+      "version": "2.3.9",
+      "resolved": "https://registry.npmmirror.com/@asamuzakjp/nwsapi/-/nwsapi-2.3.9.tgz",
+      "integrity": "sha512-n8GuYSrI9bF7FFZ/SjhwevlHc8xaVlb/7HmHelnc/PZXBD2ZR49NnN9sMMuDdEGPeeRQ5d0hqlSlEpgCX3Wl0Q==",
+      "dev": true,
+      "license": "MIT"
     },
     "node_modules/@babel/code-frame": {
       "version": "7.29.0",
@@ -309,9 +364,9 @@
       }
     },
     "node_modules/@babel/parser": {
-      "version": "7.29.2",
-      "resolved": "https://registry.npmmirror.com/@babel/parser/-/parser-7.29.2.tgz",
-      "integrity": "sha512-4GgRzy/+fsBa72/RZVJmGKPmZu9Byn8o4MoLpmNe1m8ZfYnz5emHLQz3U4gLud6Zwl0RZIcgiLD7Uq7ySFuDLA==",
+      "version": "7.29.3",
+      "resolved": "https://registry.npmmirror.com/@babel/parser/-/parser-7.29.3.tgz",
+      "integrity": "sha512-b3ctpQwp+PROvU/cttc4OYl4MzfJUWy6FZg+PMXfzmt/+39iHVF0sDfqay8TQM3JA2EUOyKcFZt75jWriQijsA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -411,6 +466,169 @@
       },
       "engines": {
         "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@bcoe/v8-coverage": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmmirror.com/@bcoe/v8-coverage/-/v8-coverage-1.0.2.tgz",
+      "integrity": "sha512-6zABk/ECA/QYSCQ1NGiVwwbQerUCZ+TQbp64Q3AgmfNvurHH0j8TtXa1qbShXA6qqkpAj4V5W8pP6mLe1mcMqA==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/@bramus/specificity": {
+      "version": "2.4.2",
+      "resolved": "https://registry.npmmirror.com/@bramus/specificity/-/specificity-2.4.2.tgz",
+      "integrity": "sha512-ctxtJ/eA+t+6q2++vj5j7FYX3nRu311q1wfYH3xjlLOsczhlhxAg2FWNUXhpGvAw3BWo1xBcvOV6/YLc2r5FJw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "css-tree": "^3.0.0"
+      },
+      "bin": {
+        "specificity": "bin/cli.js"
+      }
+    },
+    "node_modules/@csstools/color-helpers": {
+      "version": "6.0.2",
+      "resolved": "https://registry.npmmirror.com/@csstools/color-helpers/-/color-helpers-6.0.2.tgz",
+      "integrity": "sha512-LMGQLS9EuADloEFkcTBR3BwV/CGHV7zyDxVRtVDTwdI2Ca4it0CCVTT9wCkxSgokjE5Ho41hEPgb8OEUwoXr6Q==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT-0",
+      "engines": {
+        "node": ">=20.19.0"
+      }
+    },
+    "node_modules/@csstools/css-calc": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmmirror.com/@csstools/css-calc/-/css-calc-3.2.1.tgz",
+      "integrity": "sha512-DtdHlgXh5ZkA43cwBcAm+huzgJiwx3ZTWVjBs94kwz2xKqSimDA3lBgCjphYgwgVUMWatSM0pDd8TILB1yrVVg==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.19.0"
+      },
+      "peerDependencies": {
+        "@csstools/css-parser-algorithms": "^4.0.0",
+        "@csstools/css-tokenizer": "^4.0.0"
+      }
+    },
+    "node_modules/@csstools/css-color-parser": {
+      "version": "4.1.1",
+      "resolved": "https://registry.npmmirror.com/@csstools/css-color-parser/-/css-color-parser-4.1.1.tgz",
+      "integrity": "sha512-eZ5XOtyhK+mggRafYUWzA0tvaYOFgdY8AkgQiCJF9qNAePnUo/zmsqqYubBBb3sQ8uNUaSKTY9s9klfRaAXL0g==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "dependencies": {
+        "@csstools/color-helpers": "^6.0.2",
+        "@csstools/css-calc": "^3.2.1"
+      },
+      "engines": {
+        "node": ">=20.19.0"
+      },
+      "peerDependencies": {
+        "@csstools/css-parser-algorithms": "^4.0.0",
+        "@csstools/css-tokenizer": "^4.0.0"
+      }
+    },
+    "node_modules/@csstools/css-parser-algorithms": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmmirror.com/@csstools/css-parser-algorithms/-/css-parser-algorithms-4.0.0.tgz",
+      "integrity": "sha512-+B87qS7fIG3L5h3qwJ/IFbjoVoOe/bpOdh9hAjXbvx0o8ImEmUsGXN0inFOnk2ChCFgqkkGFQ+TpM5rbhkKe4w==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.19.0"
+      },
+      "peerDependencies": {
+        "@csstools/css-tokenizer": "^4.0.0"
+      }
+    },
+    "node_modules/@csstools/css-syntax-patches-for-csstree": {
+      "version": "1.1.4",
+      "resolved": "https://registry.npmmirror.com/@csstools/css-syntax-patches-for-csstree/-/css-syntax-patches-for-csstree-1.1.4.tgz",
+      "integrity": "sha512-wgsqt92b7C7tQhIdPNxj0n9zuUbQlvAuI1exyzeNrOKOi62SD7ren8zqszmpVREjAOqg8cD2FqYhQfAuKjk4sw==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT-0",
+      "peerDependencies": {
+        "css-tree": "^3.2.1"
+      },
+      "peerDependenciesMeta": {
+        "css-tree": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@csstools/css-tokenizer": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmmirror.com/@csstools/css-tokenizer/-/css-tokenizer-4.0.0.tgz",
+      "integrity": "sha512-QxULHAm7cNu72w97JUNCBFODFaXpbDg+dP8b/oWFAZ2MTRppA3U00Y2L1HqaS4J6yBqxwa/Y3nMBaxVKbB/NsA==",
+      "dev": true,
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/csstools"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/csstools"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.19.0"
       }
     },
     "node_modules/@emotion/hash": {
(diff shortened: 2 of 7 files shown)


an9xyz:
Re: dep-change-bot — confirming the `package.json` changes are intentional. All three are **devDependencies** (no impact on the production bundle):

- `vitest ^4.1.6` — test runner; project had no test framework before this PR
- `@vitest/coverage-v8 ^4.1.6` — coverage reporter, wired up via the new `test:coverage` script
- `jsdom ^29.1.1` — required by `vitest.environment: 'jsdom'` so the zustand `persist` middleware can hit `localStorage` in tests

Also added three npm scripts: `test`, `test:watch`, `test:coverage`.

No runtime dependencies changed.

- [x] Confirm dependency changes are intentional
- [x] Review package delta if lockfile changed

yujiawei (dismissed):
# Code Review — PR #27 (octo-admin)

## Verdict

**APPROVED.** The PR cleanly fixes a real, user-visible bug (10-minute "应用 Bot" menu blackout after a transient backend hiccup) with a minimal, well-tested change. No blocking issues found. A few non-blocking suggestions below.

## 1. Validation

| Item | Status | Evidence |
|---|---|---|
| Root cause matches PR description | ✅ | `src/store/feature.ts` (pre-PR) cached every `catch` with the 10-min TTL; the only signal that could distinguish "route missing" from "5xx" was `error.response.status`, and `src/api/index.ts:27` was throwing `new Error(message)` which discarded it. The PR addresses both ends. |
| Status preserved end-to-end | ✅ | `src/api/index.ts:4-11,36` defines `ApiError` and `src/api/index.ts:36` propagates `error.response?.status`. `src/store/feature.ts:43-44` reads it via `e instanceof ApiError ? e.status : undefined`. |
| 404 → 10-min TTL (unchanged) | ✅ | `src/store/feature.ts:45,49` keeps `APP_BOTS_TTL_MS` when `status === 404`. Covered by `feature.test.ts:51-58`. |
| 5xx / network → 30-s transient TTL | ✅ | `src/store/feature.ts:45,49` selects `APP_BOTS_TRANSIENT_TTL_MS` for any non-404 error. Covered by `feature.test.ts:60-74`. |
| Recovers within 30s after backend heals | ✅ | `feature.test.ts:76-87` advances fake timers past the transient TTL and verifies a second request fires and succeeds. |
| No hammering during sustained outage | ✅ | `feature.test.ts:89-97` verifies cache holds inside the 30s window. |
| `resetFeatures` resets TTL too | ✅ | `src/store/feature.ts:54-55` and `feature.test.ts:108-119`. |
| Zustand persist migration safety | ✅ | Default `merge` does `{...currentState, ...persistedState}`. Existing localStorage entries lack `appBotsCheckTtl`, so the new initial `appBotsCheckTtl: APP_BOTS_TTL_MS` is preserved — no `undefined` reads from `isFresh`. See note 1 below for a one-time-upgrade caveat. |
| `ApiError` does not break existing axios consumers | ✅ | `ApiError extends Error`, preserves `.message`, so `instanceof Error` and `e.message` keep working. The previous interceptor already wrapped errors in `new Error(message)` and discarded `error.response`, so no caller could have been relying on the raw axios error. Verified there are no other call sites importing `ApiError`. |
| Probe call sites still correct | ✅ | `src/layouts/MainLayout.tsx:68-72` and `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:69-74` both call `void probeAppBots()` in `useEffect`; signature unchanged. |
| `npm run build` claim | ✅ | tsc gate is preserved; nothing in the diff would break type-checking (new field added to interface, `ApiError` exported). |
| Test infra wiring | ✅ | `vite.config.ts:15-19` adds `test` block scoped to `src/**/*.test.{ts,tsx}`. `vite.config.ts:1` adds the `/// <reference types="vitest" />` triple-slash needed for the type augmentation under `defineConfig`. |

## 2. Findings

None at P0/P1.

## 3. Suggestions (non-blocking)

### S1. Persist `version` bump for instant heal on upgrade
`src/store/feature.ts:57-59` keeps the existing `persist` config without bumping `version`. Users who hit the bug *right before* this fix ships will still be inside the old 10-min TTL after upgrade — their persisted `{appBotsAvailable: false, appBotsCheckedAt: <recent>}` will hydrate, `appBotsCheckTtl` will default to 10 min via the merge, and they'll wait out the old timer once. Bumping `version: 1` (with the default `migrate` discarding the old shape) would clear that one-time tail.

Suggested:
```ts
{
  name: 'dm-admin-features',
  version: 1, // force re-probe so users sitting on the pre-fix 10-min cache heal immediately
}
```
Trade-off: an extra probe per upgraded session. Given the bug's blast radius (menu hidden for 10 min) this seems worth it, but reasonable to defer.

### S2. Treat only 5xx / network as transient; 4xx (non-404) should probably be sticky
`src/store/feature.ts:45` uses `const transient = status !== 404`, which means 400/401/403/etc. also get the 30-s short TTL. In practice:
- 401 is intercepted upstream (`src/api/index.ts:30-33`) and the user is redirected, so unreachable here as the comment notes.
- 403 ("forbidden for this user") is a *stable* signal, not transient; re-probing every 30s wastes requests for the user's entire session.
- 400 ("bad request") should not happen for a fixed-URL GET, but if it did, it's also not transient.

A slightly tighter rule:
```ts
const transient = status === undefined || status >= 500
```
That preserves the fast recovery for the actual transient classes (network, timeout, 5xx) and avoids hammering on a stable 4xx. Not a correctness bug today — just a minor efficiency / future-proofing note.

### S3. (Pre-existing, not introduced) Two layouts probe in parallel on fresh load
`src/layouts/MainLayout.tsx:70-72` and `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:72-74` each call `probeAppBots()` in `useEffect`. On a cold load they both fire before either has set state, so the de-dupe gate (`appBotsAvailable !== null`) doesn't trip and the backend gets two near-simultaneous requests. Pre-existing, not in scope for this PR, but worth a follow-up — an in-flight promise gate inside `probeAppBots` would solve it.

### S4. Test nit — drop `mockReset` calls duplicated by `vi.fn()` reuse
`feature.test.ts:21` calls `apiGet.mockReset()` in `beforeEach`, but the mock is defined once at module scope with `vi.fn()`. Reset is correct here — just noting that if you ever switch to `vi.restoreAllMocks` (recommended when you also stub real implementations) you'd want to be aware of the difference. No action required.

## 4. Additional notes

- The behavior table in the PR description matches the code. ✅
- Test coverage is appropriate for the surface changed (9 cases for ~15 lines of logic).
- The license addition (`Apache-2.0` in `package.json`) and the lockfile reformatting are noise but acceptable in the same PR.
- vitest 4.x and jsdom 29 are recent majors but both stable; no concerns for a frontend-only test runner.

lml2468 (dismissed):
[APPROVED]

**Verdict:** APPROVED — root cause correctly identified, fix is targeted and well-tested.

**CI Note:** CodeQL and build checks still IN_PROGRESS at review time; no failures observed.

---

**Root cause trace**

The original interceptor in `src/api/index.ts`:
```ts
return Promise.reject(new Error(message))
```
stripped `error.response?.status`, making every failure indistinguishable in the catch block. Introducing `ApiError` with an optional `status` field is the correct minimal fix — it threads the HTTP status through without touching callers that only need the message.

---

**Logic verification — `transient = status !== 404`**

| Error type | `status` | `transient` | TTL | Expected? |
|---|---|---|---|---|
| 200 | — | — | 10 min | ✅ |
| 404 | 404 | false | 10 min | ✅ |
| 5xx | 5xx | true | 30 s | ✅ |
| Network/timeout | undefined | true | 30 s | ✅ |

The logic is elegant and correct for the stated failure modes.

**Persistence correctness:** `appBotsCheckTtl` is included in the zustand-persist store (no `partialize` exclusion), which is intentional — the short TTL survives page reload, so the re-probe happens within 30 s of recovery, not 10 min. Success resets it to 10 min. ✅

**Test coverage:** 9 tests cover all distinct paths — success, 404, 5xx, network-no-status, transient-recovery, anti-hammering, force-bypass, and reset. `vi.useFakeTimers()` + `vi.setSystemTime()` is the right approach for TTL-dependent tests. ✅

---

🔵 **Suggestion — non-404 client errors get transient TTL**

`transient = status !== 404` means 403/400/422 all get the 30-second TTL. For a feature-probe endpoint a 403 (access denied) could be a stable condition, not transient. Low risk for this specific endpoint, but worth a comment or an explicit list:

```ts
// Treat only 404 as a stable "route not registered" signal; everything else
// (5xx, network, and even unexpected 4xx) is treated as transient.
const transient = status !== 404
```

Not a blocker — the current behavior is safe and the comment above explains the intent.

---

🔵 **Minor — test constants duplicated from source**

`APP_BOTS_TTL_MS` and `APP_BOTS_TRANSIENT_TTL_MS` are re-declared in `feature.test.ts`. If the production values change, the test assertions silently drift. Consider exporting them from `feature.ts` (or a constants file) so tests import the live values. Not blocking.

---

🔵 **Unrelated: `package-lock.json` name rename**

`dmwork-admin` → `octo-admin` in the lockfile is bundled here. Harmless, but worth noting it's unrelated to the fix.

Jerry-Xin (changes requested):
The feature-probe logic is relevant to octo-admin and mostly matches the intended behavior, but the new test stack breaks the repository’s existing Node 18 support.

🔴 Blocking

🔴 Critical: `npm test` does not run on the repo’s supported Node 18 environment. `package.json:32-36` adds `vitest@4.1.6` and `jsdom@29.1.1`; the lockfile records Node 20+ engine requirements for jsdom and Vitest (`package-lock.json:2711`, `package-lock.json:4120`). This conflicts with the project’s current Node 18 targets in `.github/workflows/ci.yml:17` and `Dockerfile:1`. I reproduced the failure with Node 18.20.8: Vitest fails before collecting tests with `ERR_REQUIRE_ESM` from the jsdom dependency chain. Please either use test dependencies that support Node 18, or deliberately migrate CI, Docker, and documented runtime expectations to Node 20+.

💬 Non-blocking

🟡 Warning: `package-lock.json` now updates the root metadata from `dmwork-admin` to `octo-admin` and adds `license` metadata. This appears consistent with `package.json`, but it is unrelated to the feature fix and worth keeping only if intentional.

✅ Highlights

The `ApiError` addition in `src/api/index.ts:4-10` preserves HTTP status cleanly enough for this use case.

The store behavior in `src/store/feature.ts:30-50` correctly distinguishes stable 404s from transient failures and resets the TTL on later success.

The test cases in `src/store/feature.test.ts:31-125` cover the important cache, TTL, recovery, force-refresh, and reset paths.

Verification run: `npm test` and `npm run build` pass under Node 24.10.0 after `npm ci`; `npm test` fails under Node 18.20.8 as described above.

lml2468 (commented):
[APPROVE] *(Submitting as COMMENT — gh CLI is authenticated as PR author; cannot self-approve.)*

Independent cross-review for PR #27.

## Verdict: APPROVED — correct fix, solid tests, no blocking issues.

## What I verified

### 1. Core logic: transient vs 404 differentiation ✅

The change correctly separates two fundamentally different failure modes:
- **404** = route not registered → definitive signal, cache 10 min
- **5xx / network** = transient outage → cache 30 sec, recover quickly

Traced through the full error path:
1. `axios` interceptor wraps **all** response errors as `ApiError(message, status?)` → `status` is `undefined` for network errors (no `error.response`)
2. In `probeAppBots`, `e instanceof ApiError` is always true for errors from this api instance ✅
3. `status !== 404` correctly routes: `undefined !== 404` → transient (true), `500 !== 404` → transient (true), `404 !== 404` → definitive (false) ✅

The 30-second window balances two competing concerns well: fast recovery from startup races vs not hammering a dying backend.

### 2. ApiError class ✅

Minimal, carries status, properly extends Error with `this.name = "ApiError"`. Clean integration — the interceptor already constructed it at the boundary, so consumers get typed errors automatically.

### 3. Test quality ✅

9 test cases cover all meaningful paths:
- Success caching + TTL skip
- 404 → full TTL
- 5xx → short TTL
- Network error (no status) → short TTL
- Transient TTL expiry → re-probe + recovery
- Sustained outage → no hammering within TTL
- `force=true` bypass
- `resetFeatures` cleanup

All use `vi.useFakeTimers()` for deterministic time assertions. Mock isolation is correct — `vi.mock` with `importActual` preserves `ApiError` while stubbing the default export.

### 4. Persist migration (localStorage schema change) ✅

`appBotsCheckTtl` is new. Zustand persist does a shallow merge on hydration — old localStorage (without `appBotsCheckTtl`) merges over the initial state, so `appBotsCheckTtl` defaults to `APP_BOTS_TTL_MS`. Worst case: a user who had a 5xx cached pre-upgrade gets the old 10-min TTL for that single cache entry, identical to previous behavior. No regression.

### 5. CI green ✅

All 9 checks pass: build (Node 18), build (Node 20), CodeQL, actionlint, labeler.

## On Allen's P0 finding (Node 18 compat)

I **disagree** with the P0 severity. Here is why:

1. **Node 18 went EOL in April 2025** — over a year ago. Continuing to support it is a choice, not a hard requirement.
2. **vitest/jsdom are devDependencies** — they affect testing, not production builds. The Node 18 build CI (`npm ci && tsc --noEmit && npm run build`) passes because `npm` does not enforce `engines` by default and the build pipeline never runs `npm test`.
3. **CI does not run tests today.** The CI workflow runs `tsc --noEmit` + `npm run build` only. There is no `npm test` step. So this PR does not break any existing green check.
4. The correct fix is a follow-up chore: either drop Node 18 from CI matrix (it is EOL) or add `npm test` only on the Node 20+ row. Neither should block this PR.

I would rate this **P2** — a CI matrix cleanup item, not a production or build blocker.

## Suggestions (non-blocking)

**P3** — Consider adding `npm run test` to CI (on Node 20+ only), so the new tests actually gate PRs. Without it, test regressions are invisible in CI.

**P3** — The persist config has no `version` field. If the store schema evolves further, adding `version: 1` + a `migrate` function now would make future changes smoother.

an9xyz:
Pushed `d8f6ec2` addressing all review feedback.

**@Jerry-Xin — blocking issue resolved**
- Downgraded `vitest` 4.1.6 → **2.1.9** and `jsdom` 29.1.1 → **24.1.3**; both explicitly support Node 18.
- Reproduced your `ERR_REQUIRE_ESM` failure locally on Node 18.16.0 before the fix; after the downgrade, `npm test` passes 9/9 on **both Node 18.16.0 and Node 20.18.1**, and `npm ci` no longer emits `EBADENGINE` warnings.
- Also added `- run: npm test` to `.github/workflows/ci.yml` so any future Node-18 incompatibility fails CI instead of silently shipping.

**@lml2468 — non-blocking suggestions adopted**
- `APP_BOTS_TTL_MS` / `APP_BOTS_TRANSIENT_TTL_MS` are now exported from `src/store/feature.ts`; the test imports them so future TTL changes can't silently drift the assertions.
- Expanded the comment around `transient = status !== 404` to make the policy explicit (404 = stable "route not registered", everything else including unexpected 4xx = transient 30 s TTL).

The unrelated `package-lock.json` rename / license note remains — happy to revert that hunk if preferred.

yujiawei (approved):
# Code Review — PR #27 (octo-admin)

## Summary

Targeted, well-scoped fix for the issue where a single transient probe failure (5xx / network / timeout) of `/v1/app_bot/available` would cache `appBotsAvailable=false` for 10 minutes in `localStorage`, sticky-hiding the **应用 Bot** menu long after the backend recovered. The change distinguishes failure modes by HTTP status (via a new `ApiError` carrying `response.status`), applies the original 10-min TTL only to success / 404 (stable signal), and uses a 30-second transient TTL for everything else. Tests, build, and CI gating are all in place.

I reproduced the change locally and checked callers — no regressions to the existing UX contract.

## Verification

- ✅ Bug → fix mapping. `src/store/feature.ts:46-54` reads `e.status`, sets `transient = status !== 404`, and chooses `APP_BOTS_TRANSIENT_TTL_MS` vs `APP_BOTS_TTL_MS` accordingly. The 5xx-during-startup-race scenario from the PR description is now bounded to ~30 s.
- ✅ Status is actually preserved end-to-end. `src/api/index.ts:28-37` wraps rejections in `ApiError(message, error.response?.status)` instead of a bare `Error(message)`, so `e instanceof ApiError ? e.status : undefined` in the store can read it. Without this change the fix would be inert.
- ✅ TTL field is honored in the freshness check. `src/store/feature.ts:32` switches `isFresh(appBotsCheckedAt, APP_BOTS_TTL_MS)` → `isFresh(appBotsCheckedAt, appBotsCheckTtl)`, so the new short TTL actually applies on read, not only on write.
- ✅ Callers unaffected. The only consumers — `src/App.tsx:36-38`, `src/layouts/MainLayout.tsx:67-79`, `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:68-81` — read `appBotsAvailable` (nullable boolean) and call `probeAppBots()`. Signatures unchanged, behavior is strictly more responsive.
- ✅ Tests cover the meaningful axes: success, 404 (stable TTL), 5xx (transient TTL), network-error with no status, recovery after transient TTL expires, no-hammering inside the transient window, force bypass, and `resetFeatures`. Constants are imported from the live module (`APP_BOTS_TTL_MS`, `APP_BOTS_TRANSIENT_TTL_MS`) so they cannot silently drift.
- ✅ Node 18 compatibility preserved. `vitest@^2.1.9` and `jsdom@^24.1.3` both keep Node 18 in their supported range, matching the CI matrix `[18, 20]` and `node:18-alpine` in `Dockerfile`. CI now runs `npm test` between `tsc --noEmit` and `npm run build` (`.github/workflows/ci.yml:26`), so any future bump that drops Node 18 will fail fast on the 18 leg.
- ✅ Persisted state is forward-compatible. The `persist` middleware's default shallow merge keeps existing fields from old localStorage entries while initializing the new `appBotsCheckTtl` from the store's defaults — no orphan-state crash on first load after deploy.

## Findings

### P2 / nit — `ApiError` is broader than `feature.ts`; check other catch sites

The interceptor change is a global contract change: every `api.*` rejection is now an `ApiError`, not a plain `Error`. I grepped for callers — none in this repo inspect the rejection beyond `.message`, so this is purely additive today. Worth keeping in mind when future code wants to handle errors by status: prefer `e instanceof ApiError ? e.status : undefined` (the pattern this PR already uses) so plain `Error` rejects (e.g. timeouts re-thrown from elsewhere) degrade gracefully.

### P2 / nit — comment on the 401 path is slightly misleading

`src/store/feature.ts:43-45` says "401 已被 axios 拦截器处理为登出，此处不到达". The interceptor at `src/api/index.ts:31-34` does log out and redirect on 401, but it does **not** short-circuit — it still calls `Promise.reject(new ApiError(...))`, so the `catch` block here would still observe a 401 in the brief window before the `window.location.href` navigation actually unloads the page. With `transient = status !== 404`, a 401 would be (harmlessly) cached as transient for 30 s. Not a real bug — the page is navigating away — but the comment promise ("此处不到达") is not strictly true. Either tweak the comment ("此处理论上不可达，但若到达按瞬态处理") or have the interceptor return `Promise.reject` early on 401 to make the comment factual.

### P2 / nit — interceptor still leaks `error` rather than wrapping it

Adjacent observation, not blocking: `error.message` for a network failure is something like `"Network Error"`, while many downstream messages would benefit from a richer string (`error.code`, request URL). Out of scope for this PR — leaving as a future improvement.

### P2 / nit — no test asserts the interceptor produces `ApiError`

The store tests mock `api.get` directly and inject `ApiError` themselves, which is fine for the store logic. But the contract "axios → `ApiError(message, status)`" is now load-bearing; a one-shot test that hits the real interceptor with a stubbed adapter (or `MockAdapter`) would lock that contract down. Optional.

## Recommendation

Approve. The fix is the minimal, correct change for the reported bug; the new `ApiError` type is the right abstraction for status-aware error handling; tests are focused and pass on both supported Node versions; CI gating prevents the next regression.

The nits above are maintainability suggestions, not merge blockers.

Jerry-Xin (approved):
Relevant PR; the implementation addresses the App Bots transient-probe caching bug, preserves API status for feature probing, and adds useful regression coverage.

💬 Non-blocking
- 🟡 Warning: `src/layouts/MainLayout.tsx:70` and `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:72` only call `probeAppBots()` on mount. The 30-second TTL is passive, so the menu reappears after a later probe, reload, or remount, not automatically 30 seconds after backend recovery. If automatic recovery is expected, add a delayed retry or visibility/focus-triggered re-probe after transient failures.
- 🟡 Warning: `package.json:32-36` introduces Vitest 2.1.9; current `npm audit` reports dev-only advisories through the Vitest/Vite stack. Not a production blocker, but worth upgrading in a follow-up if compatible.

✅ Highlights
- `src/api/index.ts:4` preserves HTTP status via `ApiError`, which fits the feature-store need without disrupting existing callers that read `.message`.
- `src/store/feature.ts:31-52` correctly distinguishes 404 from transient failures and persists the chosen TTL.
- `src/store/feature.test.ts` covers the main cache paths, retry window, force bypass, and reset behavior.
- Verified locally: `npm test` passed, and `npm run build` completed successfully.

lml2468 (commented):
[APPROVE] *(Re-review after d8f6ec2 — gh CLI auth as PR author, submitting as COMMENT.)*

Incremental re-review. All previous findings from Jerry-Xin, lml2468, and my own P3 suggestions are resolved in one clean commit.

## Changes verified

### 1. Node 18 compat restored ✅
`vitest` 4.1.6 → 2.1.9, `jsdom` 29.1.1 → 24.1.3, `@vitest/coverage-v8` 4.1.6 → 2.1.9. Both downgraded versions support Node 18. CI `build (18)` + `build (20)` green — **including `npm test`** which now runs in the pipeline.

### 2. CI now runs tests ✅
`npm test` added to the matrix, executes on both Node 18 and 20. My previous P3 suggestion is now a reality. Future test regressions or Node compat breaks will fail fast.

### 3. TTL constants exported ✅
`APP_BOTS_TTL_MS` and `APP_BOTS_TRANSIENT_TTL_MS` are now `export const` in `feature.ts`. Test imports the live values instead of redeclaring — eliminates silent drift risk.

### 4. Comment expanded ✅
Three-line comment on `transient = status !== 404` makes the design intent explicit: only 404 is stable, everything else (5xx, network, timeout, unexpected 4xx) is bounded by 30s transient TTL.

## CI ✅
All 9 checks green at d8f6ec2, including `npm test` on Node 18 and 20.

No remaining issues. Approve and merge.

PR #29, opened 2026-05-20
Render system settings dynamically

an9xyz:
## Summary
- Render system settings by category from API items instead of hard-coded register/support fields
- Add generic controls for bool, string, int, and encrypted setting types
- Preserve encrypted settings when the field is left blank during save

## Testing
- npm run build

### DIFF
--- a/src/pages/SystemSetting/index.tsx
+++ b/src/pages/SystemSetting/index.tsx
@@ -9,17 +9,13 @@ import {
   Modal,
   Row,
   Select,
-  Space,
-  Tag,
   Tooltip,
   message,
 } from 'antd'
 import {
-  MailOutlined,
   ReloadOutlined,
   SaveOutlined,
   SendOutlined,
-  UserAddOutlined,
 } from '@ant-design/icons'
 import {
   SECRET_MASK,
@@ -31,88 +27,21 @@ import {
 } from '../../api/system-setting'
 
 type BoolFormValue = '' | '0' | '1'
+type SettingFormValue = string | number | null | undefined
 
-interface SystemSettingFormValues {
-  register_off: BoolFormValue
-  register_only_china: BoolFormValue
-  register_username_on: BoolFormValue
-  register_email_on: BoolFormValue
-  support_email: string
-  support_email_smtp: string
-  support_email_pwd: string
-}
+type SystemSettingFormValues = Record<string, SettingFormValue>
 
 interface TestEmailFormValues {
   to: string
 }
 
-interface BoolSettingDef {
-  name: keyof SystemSettingFormValues
-  category: 'register'
-  key: 'off' | 'only_china' | 'username_on' | 'email_on'
-  label: string
-  description: string
-  onLabel: string
-  offLabel: string
-}
-
-const boolSettings: BoolSettingDef[] = [
-  {
-    name: 'register_off',
-    category: 'register',
-    key: 'off',
-    label: '注册入口',
-    description: '是否关闭注册',
-    onLabel: '关闭注册',
-    offLabel: '允许注册',
-  },
-  {
-    name: 'register_only_china',
-    category: 'register',
-    key: 'only_china',
-    label: '手机号范围',
-    description: '仅中国手机号可以注册',
-    onLabel: '仅中国手机号',
-    offLabel: '不限制地区',
-  },
-  {
-    name: 'register_username_on',
-    category: 'register',
-    key: 'username_on',
-    label: '用户名注册',
-    description: '是否开启用户名注册',
-    onLabel: '开启',
-    offLabel: '关闭',
-  },
-  {
-    name: 'register_email_on',
-    category: 'register',
-    key: 'email_on',
-    label: '邮箱注册/登录',
-    description: '是否开启邮箱注册/登录',
-    onLabel: '开启',
-    offLabel: '关闭',
-  },
-]
-
 const settingMapKey = (category: string, key: string) => `${category}.${key}`
+const settingFormName = (category: string, key: string) => settingMapKey(category, key)
 
-const initialValues: SystemSettingFormValues = {
-  register_off: '',
-  register_only_china: '',
-  register_username_on: '',
-  register_email_on: '',
-  support_email: '',
-  support_email_smtp: '',
-  support_email_pwd: '',
-}
-
-function getValue(items: SystemSettingItem[], category: string, key: string) {
-  return items.find((item) => item.category === category && item.key === key)?.value ?? ''
-}
-
-function getSetting(items: SystemSettingItem[], category: string, key: string) {
-  return items.find((item) => item.category === category && item.key === key)
+const categoryTitles: Record<string, string> = {
+  login: '登录配置',
+  register: '注册配置',
+  support: '邮件服务',
 }
 
 function normaliseBoolValue(value: string): BoolFormValue {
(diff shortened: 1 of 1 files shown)


lml2468 (approved):
[APPROVED]

**Verdict:** APPROVED — architecturally sound refactor; core logic and encrypted-value preservation are correct.

**CI Note:** CodeQL and build checks still IN_PROGRESS at review time.

---

**Encrypted-value preservation — trace**

```ts
const keepEncryptedValue =
  item.value_type === 'encrypted' && !value &&
  (item.configured || item.value === SECRET_MASK)
```

| State | `item.configured` | `item.value` | blank form | result |
|---|---|---|---|---|
| DB has a value | true | `SECRET_MASK` | ✓ | send `SECRET_MASK` (preserve) ✅ |
| Default has a value | false | `SECRET_MASK` | ✓ | send `SECRET_MASK` (preserve) ✅ |
| No value anywhere | false | `''` / `null` | ✓ | send `''` (clear is fine) ✅ |

`valuesFromSettings` always initialises encrypted fields to `''`, consistent with the "blank = preserve existing" contract. ✅

**`settingGroups` useMemo** groups items by category with insertion-order preserved via `Map` — correct, renders tabs in the order the API returns categories. ✅

**`valuesFromSettings` bool path** — first assigns `item.value ?? ''` then immediately normalises via `normaliseBoolValue`. The double-assignment reads awkward but is correct. ✅

---

🟡 **`genericSettingExtra` returns raw key when no effective value**

```ts
if (!item.effective_value) return identity  // → e.g. "register.off"
```

When a setting has never been configured, the only help text shown to the operator is the technical key (`register.off`), not the human-readable `item.description`. The description is already used as the field label via `settingLabel()`, so repeating it in `extra` would be redundant — but `undefined` (no extra text) would be cleaner than a raw API key:

```ts
if (!item.effective_value) return undefined
```

Not blocking; admin-only page and the key is still parseable.

---

🟡 **Bool `onLabel`/`offLabel` lost — now always '是' / '否'**

Old code had per-field meaningful labels ('关闭注册' / '允许注册', '仅中国手机号' / '不限制地区'). The dynamic approach can't carry per-field labels without the API providing them. If the backend can extend `SystemSettingItem` with `on_label` / `off_label` fields, the Select options and `boolText()` could be made richer. For now '是' / '否' is acceptable but loses context.

---

🔵 **`int` type — `<Input type="number">` vs `<InputNumber>`**

```tsx
return <Input allowClear type={item.value_type === 'int' ? 'number' : undefined} />
```

Native `type="number"` on an HTML input does not prevent decimal entry in all browsers and skips Ant Design's validation UX. `<InputNumber precision={0} />` would give integer enforcement. Low risk for an admin page.

---

🔵 **No way to intentionally clear a configured encrypted field**

`allowClear` on the password input clears the form field to `''`, but `keepEncryptedValue` will be `true` for a configured field — so the clear action silently preserves the old value. The placeholder text documents this correctly (`留空保留现有值`), but there is no escape hatch to remove a stored encrypted setting via this UI. Worth a follow-up issue or a dedicated "clear" affordance if that use case exists.

---

🔵 **`schema` field from API response is unused**

`SystemSettingResponse.schema` (`SystemSettingSchemaItem[]`) is fetched but never consumed. If it carries ordering or validation metadata, it could replace the hardcoded `categoryTitles` map.

yujiawei (approved):
# Code Review — PR #29 (octo-admin)

## Verdict

Approving. The refactor is a clear net win: ~110 lines of hard-coded field plumbing replaced by a generic renderer driven by the `SystemSettingItem.value_type` discriminator. No P0/P1 issues — no correctness, security, build, or data-loss blockers. Comments below are P2 suggestions worth folding into a follow-up, not merge blockers.

## What I checked

- Diff against `main` for `src/pages/SystemSetting/index.tsx` (the only changed file).
- `src/api/system-setting.ts` — type definitions for `SystemSettingItem`, `SystemSettingUpdateItem`, `SystemSettingValueType`, and `SECRET_MASK`. The four supported `value_type`s (`bool | string | int | encrypted`) are all handled by `renderSettingInput`.
- AntD 5.24 / rc-field-form `Form.Item name` semantics: a string `name` is a literal field key (only arrays denote nested paths), so `"register.off"` is safe — no accidental nesting from the dot.
- Encrypted-field preservation path through `valuesToPayload` to confirm the SMTP password "leave blank to keep" behaviour survives.

## Findings

### P2 — Email validation regression on `support.email`

`src/pages/SystemSetting/index.tsx:138` renders every non-bool, non-encrypted, non-int setting as a plain `<Input allowClear />`. The prior code attached `rules={[{ type: 'email', message: '请输入有效的邮箱地址' }]}` to the support email field; that rule is now gone, so the form will happily POST `not-an-email` to the backend. Backend likely re-validates, but the local UX guardrail is lost.

If you want to keep generic rendering, one option is to extend `SystemSettingItem` with an optional `format` hint (`'email' | 'host_port' | ...`) and have `renderSettingInput` translate it into AntD rules. Not blocking, but worth a follow-up ticket.

### P2 — `extra` text now shows the raw identifier instead of the description

`genericSettingExtra` (`src/pages/SystemSetting/index.tsx:97-105`) uses `settingMapKey(item.category, item.key)` (e.g. `support.email`) as the base hint string. Meanwhile `settingLabel` (`:111-113`) already renders `item.description || item.key` as the field label. So:

- Label: human-readable description.
- Extra: backend identifier + effective value.

Pre-PR, the `extra` carried the human description plus the effective-value annotation, which is more useful to an admin user. The current shape exposes implementation identifiers as the primary hint text. Consider:

```ts
function genericSettingExtra(item: SystemSettingItem) {
  const base = item.description || settingMapKey(item.category, item.key)
  if (!item.effective_value) return base
  // ... effective-value branch unchanged
}
```

### P2 — Group order is non-deterministic w.r.t. server payload

`settingGroups` (`:151-160`) preserves first-seen order from `items`. If the server reshuffles `items[]` between requests (e.g. after adding a new setting), card order shifts. Recommend sorting by a stable key — either the `categoryTitles` declaration order with unknowns appended, or alphabetical category name:

```ts
return Array.from(groups.entries()).sort(([a], [b]) => {
  const order = ['login', 'register', 'support']
  return (order.indexOf(a) + 1 || 99) - (order.indexOf(b) + 1 || 99) || a.localeCompare(b)
})
```

Field order within a group has the same caveat.

### P2 — `valuesToPayload` now POSTs every server-returned item on each save

`valuesToPayload` (`:65-76`) iterates over the full `items` array, not just fields the user touched. For non-encrypted fields whose current value is `''` (i.e. unconfigured, falling through to default), this submits `value: ""` for every save. Whether that flips `configured` from `false` → `true` in the backend depends on the server's update semantics — worth a quick check that "saving without changing anything" does not mark previously-defaulted settings as DB-configured-empty. The encrypted branch has explicit `keepEncryptedValue` guarding; non-encrypted does not.

If preserving the previous "send only known fields" behaviour matters, you can diff `values` against `valuesFromSettings(items)` and only emit changed entries.

### P2 — Encrypted placeholder/save logic inconsistency

`renderSettingInput` (`:128-135`) chooses the placeholder using only `item.configured`. But `valuesToPayload` (`:68`) preserves the existing value when `item.configured || item.value === SECRET_MASK`. So if the API returns `configured: false, value: SECRET_MASK` (e.g. an encrypted default from env), the placeholder says "留空跟随默认配置" while the save path actually preserves `SECRET_MASK`. Edge case, but the two conditions should agree.

### Nits

- Lost icons on the cards (`UserAddOutlined`, `MailOutlined`) and the DB-configured `Tag` next to the SMTP password label remove some visual cues. The genericisation reason is reasonable, just calling it out.
- Previous placeholders (`support@example.com`, `smtp.example.com:587`) were friendlier than empty. If/when you add the `format` hint above, you could attach placeholders to it.
- `initialValues={{}}` is fine but redundant; the prop can be dropped.
- `valuesFromSettings` writes `values[fieldName]` twice for bool items (once raw, once normalised). Minor — fold into a single ternary if you touch this again.

## Out-of-scope but observed

- `SystemSettingResponse.schema` is fetched from the API but never consumed. Now that rendering is item-driven, the schema field is effectively dead weight in the client. Consider removing it from the type or wiring it in to render rows for settings that exist in the schema but have no item yet.

Jerry-Xin (approved):
The PR is in scope for `octo-admin` and the dynamic settings rendering approach looks mergeable; I found no blocking correctness or security issues.

💬 Non-blocking

🟡 Warning: `src/pages/SystemSetting/index.tsx:245` now renders all string settings generically, which removes the previous email validation rule for the support sender email. If the API does not validate this server-side, admins can save an invalid sender address and only discover it later when email sending fails. Consider adding schema-driven validation or preserving key-specific validation for known fields like `support.email`.

🔵 Suggestion: `src/pages/SystemSetting/index.tsx:38-39` uses `${category}.${key}` as both display identity and form field name. This is fine for today’s known keys, but it can collide if future dynamic settings contain dots in category/key names. Using an Ant Design `NamePath` such as `[category, key]`, or an escaped internal ID, would make the dynamic form more robust.

✅ Highlights

The new payload path at `src/pages/SystemSetting/index.tsx:65-75` correctly avoids resubmitting masked encrypted values as literal user input when the field is left blank, and the UI now naturally follows whatever settings the API returns instead of being limited to the old hard-coded register/support fields.

Verification note: I attempted `npm run build`, but this checkout does not have dependencies installed, so the command failed with `tsc: command not found`.

lml2468 (commented):
[APPROVE] *(gh CLI auth as PR author — submitting as COMMENT.)*

Independent cross-review for PR #29.

## Verdict: APPROVED — sound refactor, no blocking issues.

The PR eliminates ~110 lines of hardcoded field definitions and replaces them with a generic renderer driven by `SystemSettingItem.value_type`. Any new setting category/key added on the backend automatically appears in the admin UI without frontend changes. Net win.

## What I verified independently

### 1. Encrypted-value preservation ✅

Traced `valuesToPayload` logic:
```
keepEncryptedValue = value_type=encrypted && form_empty && (configured || value===MASK)
```
- DB has value → send SECRET_MASK (preserve) ✅
- Default has value → send SECRET_MASK (preserve) ✅
- Nothing stored → send empty string (correct) ✅

`valuesFromSettings` initializes encrypted fields to `""` — consistent with "blank = keep existing" contract shown in placeholder.

### 2. Form timing correctness ✅

`setSettings(items)` and `form.setFieldsValue(valuesFromSettings(items))` are called synchronously. Ant Design Form stores values independently of rendered Form.Items — items read from the store when they mount after re-render. No race condition.

`onValuesChange` is only triggered by user interaction, not `setFieldsValue`, so `dirty` flag stays clean after fetch.

### 3. Dynamic grouping ✅

`settingGroups` uses `Map` → insertion order preserved from API response. If backend returns `[login, register, support]`, that is the card order. Correct for an API-driven UI.

### 4. Bool normalization path ✅

`valuesFromSettings` normalizes bools: `item.value → formValueToString → normaliseBoolValue`. Handles `"true"/"TRUE"/"1"` → `"1"`, `"false"/"FALSE"/"0"` → `"0"`, anything else → `""`. Roundtrip through `valuesToPayload` sends back `""`, `"1"`, or `"0"`. Correct.

### 5. CI ✅

All checks green: build (18+20), CodeQL, actionlint.

## Non-blocking observations

### P2 — Bool labels regression: "是/否" loses domain context

Old: `register_off` → "关闭注册" / "允许注册"
New: `register_off` → "是" / "否"

For a field named "是否关闭注册", "是" means "yes, close registration" — a double-negative that is cognitively harder for operators than "关闭注册." If the API can be extended with `on_label`/`off_label` fields on `SystemSettingItem`, `renderSettingInput` can use them; otherwise this is acceptable for an admin-only page. lml2468 noted this too.

### P2 — Client-side email validation removed

Old: `support_email` had `rules={[{ type: "email" }]}`.
New: generic `<Input allowClear />` with no validation rules.

If the backend does not validate email format, admins can save garbage and only discover it when sending fails. A lightweight fix: if `item.key` contains `"email"` and `value_type === "string"`, add a conditional email rule. Or better — extend the API schema to declare validation hints.

### P3 — `genericSettingExtra` shows raw key when no `effective_value`

```ts
if (!item.effective_value) return identity  // → "register.off"
```

An operator who has never configured a setting sees `register.off` as the help text. Since `item.description` is already used as the field label, returning `undefined` (no extra text) would be cleaner than showing an API key.

### P3 — API `schema` field is fetched but unused

`SystemSettingResponse.schema` carries `SystemSettingSchemaItem[]` with `type` and `description`. If it provides ordering, grouping, or validation metadata, consuming it could replace the hardcoded `categoryTitles` and enable schema-driven validation.

## Agreement with other reviewers

- lml2468's encrypted-value trace: confirmed independently ✅
- Allen's email validation warning: agree, P2 ✅
- Allen's dot-in-key collision concern: theoretically valid but practically low risk since category/key are API-controlled identifiers

Overall: clean refactor, merge-ready. UX polish can come as follow-ups.

PR #30, opened 2026-05-20
Fix app bot bind guide flow

an9xyz:
## Summary
- Show the OpenClaw bind guide directly in the app bot creation success modal alongside the one-time token.
- Share bind guide generation between the creation modal and detail drawer.
- Allow the detail drawer's copy guide action to reveal the token when needed before copying the full guide.

## Testing
- npm run build
- git diff --check

### DIFF
--- a/src/pages/AppBots/CreateModal.tsx
+++ b/src/pages/AppBots/CreateModal.tsx
@@ -1,13 +1,20 @@
 import { useState } from 'react'
-import { Modal, Form, Input, Typography, message } from 'antd'
+import { Button, Form, Input, Modal, Space, Typography, message } from 'antd'
 import { CopyOutlined } from '@ant-design/icons'
 import { createAppBot, createSpaceAppBot, type AppBotCreateReq } from '../../api/app-bot'
+import { buildConnectGuide } from './connectGuide'
 
 interface Props {
   open: boolean
   spaceId?: string
   onClose: () => void
   onSuccess: () => void
+}
+
+interface CreatedBotInfo {
+  id: string
+  displayName: string
+  token: string
 }
 
 /** Copy text to clipboard with fallback */
@@ -29,12 +36,12 @@ async function copyToClipboard(text: string): Promise<void> {
 export default function CreateModal({ open, spaceId, onClose, onSuccess }: Props) {
   const [form] = Form.useForm<AppBotCreateReq>()
   const [loading, setLoading] = useState(false)
-  const [createdToken, setCreatedToken] = useState<string | null>(null)
+  const [createdBot, setCreatedBot] = useState<CreatedBotInfo | null>(null)
 
   const handleOk = async () => {
     // If token is showing, this is the "I've saved it" confirmation
-    if (createdToken) {
-      setCreatedToken(null)
+    if (createdBot) {
+      setCreatedBot(null)
       form.resetFields()
       onSuccess()
       return
@@ -45,8 +52,12 @@ export default function CreateModal({ open, spaceId, onClose, onSuccess }: Props
       const resp = spaceId
         ? await createSpaceAppBot(spaceId, values)
         : await createAppBot(values)
-      // Show token — user must explicitly dismiss
-      setCreatedToken(resp.token)
+      // Show token and bind guide together; user must explicitly dismiss.
+      setCreatedBot({
+        id: resp.id || values.id,
+        displayName: values.display_name,
+        token: resp.token,
+      })
     } catch (err) {
       if (err instanceof Error) message.error(err.message)
     } finally {
@@ -55,9 +66,9 @@ export default function CreateModal({ open, spaceId, onClose, onSuccess }: Props
   }
 
   const handleCancel = () => {
-    if (createdToken) {
+    if (createdBot) {
       // Already created, treat cancel as confirm
-      setCreatedToken(null)
+      setCreatedBot(null)
       form.resetFields()
       onSuccess()
       return
@@ -66,34 +77,62 @@ export default function CreateModal({ open, spaceId, onClose, onSuccess }: Props
   }
 
   const handleCopyToken = async () => {
-    if (!createdToken) return
+    if (!createdBot) return
     try {
-      await copyToClipboard(createdToken)
+      await copyToClipboard(createdBot.token)
       message.success('Token copied')
     } catch {
       message.error('复制失败')
     }
   }
 
+  const handleCopyGuide = async () => {
+    if (!createdBot) return
+    try {
+      await copyToClipboard(
+        buildConnectGuide({
+          displayName: createdBot.displayName,
+          botId: createdBot.id,
+          token: createdBot.token,
+        }),
+      )
+      message.success('连接指南已复制')
+    } catch {
+      message.error('复制失败')
+    }
+  }
+
+  const connectGuide = createdBot
+    ? buildConnectGuide({
+      displayName: createdBot.displayName,
+      botId: createdBot.id,
+      token: createdBot.token,
+    })
+    : ''
+
   return (
     <Modal
-      title={createdToken ? '✅ 创建成功 — 请保存 Token' : '创建应用 Bot'}
+      title={createdBot ? '创建成功 - 请保存 Token 并完成绑定' : '创建应用 Bot'}
       open={open}
       onOk={handleOk}
       onCancel={handleCancel}
-      okText={createdToken ? '已保存，关闭' : '确定'}
+      okText={createdBot ? '已保存，关闭' : '确定'}
       confirmLoading={loading}
       destroyOnClose
+      width={createdBot ? 680 : undefined}
     >
-      {createdToken ? (
+      {createdBot ? (
         <div>
           <Typography.Paragraph type="warning" style={{ marginBottom: 12 }}>
-            ℹ️ Token 仅在创建时显示一次，关闭后无法再次查看。请立即复制并安全保存。
+            Token 仅在创建时完整显示一次。请复制 Token，并将连接指南发给 OpenClaw 执行完成绑定。
           </Typography.Paragraph>
+          <Typography.Title level={5} style={{ marginTop: 0 }}>
+            API Token
+          </Typography.Title>
           <div
             style={{
               padding: '12px 16px',
-              background: '#f5f5f5',
+              background: 'var(--a-bg-tertiary, #f5f5f5)',
               borderRadius: 8,
               fontFamily: 'monospace',
               fontSize: 13,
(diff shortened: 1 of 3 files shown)


lml2468 (approved):
[APPROVED]

**Verdict:** APPROVED — clean extraction, correct token-reveal flow, good UX improvement in DetailDrawer.

**CI Note:** CodeQL and build checks still IN_PROGRESS at review time.

---

**`connectGuide.ts` — extraction is correct**

`getBotApiUrl()` + `buildConnectGuide()` cleanly shared between `CreateModal` and `DetailDrawer`. No duplication of the guide template. Default `apiUrl = getBotApiUrl()` is evaluated at call time (correct — not a stale module-level capture). ✅

**`DetailDrawer` — auto-reveal-on-copy flow**

```ts
const token = tokenMasked ? await revealToken() : bot.token
```

`revealToken()` returns `null` only when `botId` is null, guarded by `if (!bot) return` above. The `if (!token) return` inside `finally` works correctly since `setCopyingGuide(false)` still runs. ✅

`isTokenMasked` check (`startsWith('****') || startsWith('••••')`) matches both masking styles. ✅

**`AppBotCreateResp.id`** is typed as non-optional `string`, so `resp.id || values.id` is always `resp.id`. Not a bug — defensive, but worth simplifying.

---

🔵 **`resp.id || values.id` — fallback is unreachable**

`AppBotCreateResp` types `id` as `string` (non-optional), so `resp.id` is always truthy. The `|| values.id` branch is dead code. Can simplify to `id: resp.id`.

---

🔵 **`handleCopyGuide` in `CreateModal` recomputes the guide**

The component already computes `const connectGuide = createdBot ? buildConnectGuide(...) : ''`. `handleCopyGuide` then calls `buildConnectGuide(...)` a second time with identical params. Consider:

```ts
await copyToClipboard(connectGuide)
```

---

🔵 **`buildConnectCommand` parameter over-specification**

`buildConnectCommand` accepts `BuildConnectGuideParams` but only destructures `{ botId, token, apiUrl }`. `displayName` is unused. Consider `Pick<BuildConnectGuideParams, 'botId' | 'token' | 'apiUrl'>` or just document that the full params object is accepted for convenience.

---

🔵 **Auto-reveal side-effect on "复制指南"**

When `tokenMasked` is true and the user clicks "复制指南", `revealToken()` calls `setTokenVisible(true)` — the token becomes visible in the UI as a side effect of copying the guide. This seems intentional (the guide contains the real token, so revealing it is logical), but it's an implicit UX consequence. A comment or user-facing message (`'Token 已显示并复制指南'`) would make the intent clear.

---

🔵 **`copyToClipboard` still duplicated**

Both `CreateModal.tsx` and `DetailDrawer.tsx` define their own `copyToClipboard` with identical clipboard/textarea fallback logic. Since `connectGuide.ts` is now a shared module, this helper could live there too as a follow-up.

yujiawei (approved):
# Code Review — PR #30 (octo-admin)

## Summary

This PR (a) extracts the OpenClaw bind-guide string into a shared helper (`connectGuide.ts`), (b) surfaces the bind guide alongside the one-time token in the creation success modal, and (c) makes the detail-drawer "Copy guide" button auto-reveal the masked token instead of being disabled.

Scope is small (3 files, +187/-69), `tsc --noEmit` is clean on `233dfcc`, and the refactor is genuinely a deduplication. The flagged concern is around the behavioural change in step (c), which is intentional but increases the implicit privilege of one button — worth a human eye since this is marked `security_sensitive`.

## Findings

### P2 — Detail-drawer "Copy guide" silently reveals the token in the DOM (security UX)

`src/pages/AppBots/DetailDrawer.tsx:108-176`

Before this PR, the "复制指南" button was `disabled={tokenMasked}` (old line 364): the operator first had to click "显示", which made the reveal action explicit, then could copy the guide. After this PR, the button is always enabled and the new `handleCopyGuide` flow does:

```ts
const handleCopyGuide = async () => {
  if (!bot) return
  setCopyingGuide(true)
  try {
    const token = tokenMasked ? await revealToken() : bot.token
    ...
```

and `revealToken()` (lines 108-116) both fetches the value **and** flips DOM visibility:

```ts
const revealToken = async (): Promise<string | null> => {
  ...
  setBot((prev) => (prev ? { ...prev, token: resp.token } : prev))
  setTokenVisible(true)   // <-- exposes token in the API-Token block above
  return resp.token
}
```

Net effect of one click on "复制指南" against a masked bot:
1. Network call to `revealAppBotToken` / `revealSpaceAppBotToken` (an audit-logged action).
2. `tokenVisible = true`, so the API-Token block above the guide flips from `••••••••••••••••••••` to the literal token.
3. The 12-line connect-guide block renders with the literal token.
4. Clipboard now contains the literal token.

The operator already has reveal privilege, and the 30s `TOKEN_AUTO_HIDE_MS` timer eventually re-masks step 2 — so this is not a privilege escalation. The concern is that the visible "reveal" affordance ("显示") becomes a side effect of a button whose label is "复制". An operator screen-sharing, recording, or sitting next to a colleague would not expect "复制指南" to also flash the raw token in two on-screen blocks. This is the kind of subtle behavioural change a `security_sensitive` label is there to surface.

Suggested fix (small, optional): factor `setTokenVisible(true)` out of `revealToken` and let each caller decide. `handleRevealToken` keeps it true, `handleCopyGuide` passes false. Token is still in clipboard (which is the user's intent) but it does not also light up the DOM.

```ts
const revealToken = async (opts: { showInDom?: boolean } = {}): Promise<string | null> => {
  if (!botId) return null
  const resp = spaceId ? await revealSpaceAppBotToken(spaceId, botId) : await revealAppBotToken(botId)
  setBot((prev) => (prev ? { ...prev, token: resp.token } : prev))
  if (opts.showInDom) setTokenVisible(true)
  return resp.token
}

// handleRevealToken: await revealToken({ showInDom: true })
// handleCopyGuide:   const token = tokenMasked ? await revealToken({ showInDom: false }) : bot.token
```

Not blocking — the design is explicitly stated in the PR description ("Allow the detail drawer's copy guide action to reveal the token when needed before copying the full guide"), and a human reviewer with security context should make the call.

### P2 — Bot token printed as a shell argument (pre-existing, worth re-confirming)

`src/pages/AppBots/connectGuide.ts:13-19`

```ts
return `npx -y openclaw-channel-dmwork bind --bot-token ${token} ...`
```

Passing the token as a positional CLI flag means it ends up in shell history (`~/.bash_history`, `~/.zsh_history`), in `ps`/`/proc/$pid/cmdline` while the process runs, and in any shell that echoes the pasted block. This is not new — the same string existed inline in `DetailDrawer.tsx` before — but since the bind guide is now also surfaced inside the creation modal (so more operators will copy it), it is worth confirming the OpenClaw CLI does not offer a `--bot-token-file` / `BOT_TOKEN` env-var alternative that could replace this in the printed guide. If it does, switching the guide to that form is a meaningful hardening for the same operator workflow.

### P2 — Token lives in `CreatedBotInfo` state until the modal closes

`src/pages/AppBots/CreateModal.tsx:39, 56-60, 105-111`

The created token is held in `createdBot` state and embedded into a memoised `connectGuide` string. The token is therefore present in the React fiber, in the rendered DOM twice (the API-Token block and the guide block), and in the closed-over `handleCopyGuide` / `handleCopyToken` callbacks until the operator clicks "已保存，关闭" or "取消". This matches the documented "Token 仅在创建时完整显示一次" contract — operator must see it to save it — but unlike `DetailDrawer.tsx` there is no auto-hide timer. Acceptable for a one-shot creation flow, just flagging for the human reviewer that the duplication is intentional.

### Nit — Dead export

`src/pages/AppBots/connectGuide.ts:13`

`buildConnectCommand` is exported but no caller imports it (only `buildConnectGuide` is consumed). If it is meant to be a public seam for tests or future reuse, fine; otherwise drop the `export` to keep the surface tight. Trivial.

### Nit — Inconsistent emoji cleanup across the two surfaces

`CreateModal.tsx:115, 127` vs `DetailDrawer.tsx:324, 353`

CreateModal drops the `✅` / `ℹ️` / `💡` glyphs from titles and paragraphs; DetailDrawer keeps `🔌 连接指南` and only drops `💡` from the trailing tip. Both files now show the same bind-guide block, so visual treatment diverging between them is mildly jarring. Not a correctness issue — pick a direction and apply it to both, if you care.

## What I checked

- `tsc --noEmit` on the head SHA (`233dfcc`) — clean.
- Verified `connectGuide.ts` preserves the exact bind command emitted previously (`npx -y openclaw-channel-dmwork bind --bot-token … --api-url … --account-id … --agent <agent标识>`) and the `VITE_BOT_API_URL || window.location.origin` resolution semantics.
- Confirmed `tokenMasked` is computed against `bot?.token` so the closure used in `handleCopyGuide` re-reads it per render (no stale-closure bug between successive copies).
- Confirmed no test suite exists in this repo today, so absence of new tests is not a regression.
- Confirmed no other callers of the removed `getApiUrl` helper exist outside `DetailDrawer.tsx` (no broken imports elsewhere).

## What I'd ask a human to manually verify (security_sensitive)

1. The intentional "reveal on copy" UX in the detail drawer is acceptable for your threat model (see P2 #1).
2. The OpenClaw CLI does not support a non-argv way to receive the bot token; if it does, prefer that in the printed guide (see P2 #2).
3. Auto-hide timer (30 s) is still considered sufficient now that more flows can trigger a reveal.

None of these are correctness bugs — all are policy calls.

## Verdict

No P0/P1 blockers. Refactor is clean, types compile, behaviour change is documented in the PR description. Findings are security-UX hardening suggestions, not merge blockers. Approving; human reviewer on the `needs-human-review` queue can decide whether to act on the P2s in this PR or follow-up.

Jerry-Xin (approved):
The PR is in scope for `octo-admin` and the App Bot bind-guide flow; I found no blocking issues.

💬 Non-blocking
- 🟡 Warning: [src/pages/AppBots/connectGuide.ts:18](/private/tmp/pr-review-NjAQbQ/repo/src/pages/AppBots/connectGuide.ts:18) builds a shell command by directly interpolating token, API URL, and bot ID. If any value ever contains shell metacharacters, the pasted command can fail or expand unexpectedly. Consider shell-quoting command arguments in the shared helper.
- 🔵 Suggestion: Add a small test around [src/pages/AppBots/connectGuide.ts](/private/tmp/pr-review-NjAQbQ/repo/src/pages/AppBots/connectGuide.ts:1), since it is now shared by both creation and detail flows and controls the copied user-facing command.

✅ Highlights
- The shared guide helper removes duplicated guide text between the creation modal and detail drawer.
- [src/pages/AppBots/DetailDrawer.tsx:156](/private/tmp/pr-review-NjAQbQ/repo/src/pages/AppBots/DetailDrawer.tsx:156) correctly reveals the token before copying the full guide when the drawer only has a masked token.

Verification: `git diff --check` passed. I could not independently run `npm run build` in this checkout because `tsc` is not available, likely due to missing local dependencies.

lml2468 (commented):
[APPROVE] *(gh CLI auth as PR author — submitting as COMMENT.)*

Independent cross-review for PR #30.

## Verdict: APPROVED — clean extraction, correct token flow, good UX.

This PR improves the bot creation/management UX in two ways:
1. **CreateModal**: shows the connect guide immediately after creation alongside the token, so the admin can copy both in one step.
2. **DetailDrawer**: "复制指南" now auto-reveals the token when masked, then copies the guide with the real token — eliminates the "reveal first, then copy" friction.

The `connectGuide.ts` extraction consolidates the guide text template and `getBotApiUrl()` into a shared module. No duplication of the guide string.

## What I verified independently

### 1. Auto-reveal-on-copy flow (DetailDrawer) ✅

```ts
const token = tokenMasked ? await revealToken() : bot.token
```

Traced:
- `revealToken()` calls the reveal API, updates `bot.token` via `setBot`, sets `tokenVisible = true`, returns the token string
- Returns `null` only when `botId` is null — impossible here because `if (!bot) return` guards above
- `setCopyingGuide(false)` in `finally` always runs ✅
- `isTokenMasked` checks both masking patterns (`****` and `••••`) ✅

### 2. CreateModal state management ✅

`createdBot` replaces `createdToken` — carries `{id, displayName, token}`. All branches (`handleOk` confirm, `handleCancel` dismiss) properly reset to `null` and call `form.resetFields()` + `onSuccess()`. No state leak.

### 3. `buildConnectGuide` correctness ✅

- `getBotApiUrl()` evaluated at call time (not module-level) — correct
- `buildConnectCommand` and `buildConnectGuide` cleanly separated — callers can use just the command if needed
- Template string matches the old inline version exactly (diffed mentally)

### 4. `AppBotCreateResp.id` type ✅

`id: string` is non-optional in the type. `resp.id || values.id` always resolves to `resp.id`. The fallback is defensive but dead code.

### 5. CI ✅

All checks green: build (18+20), CodeQL, actionlint.

## Non-blocking observations

### P3 — Shell interpolation in `buildConnectCommand`

Agree with Allen. The command interpolates `token`, `apiUrl`, `botId` directly:
```
npx -y openclaw-channel-dmwork bind --bot-token ${token} --api-url ${apiUrl} ...
```
- `botId`: regex-validated (`^[a-z0-9][a-z0-9_-]{0,29}$`) — safe ✅
- `token`: API-generated, likely alphanum — but not contractually guaranteed
- `apiUrl`: URL, can contain `&` or other shell-meaningful chars

Risk is low (user manually pastes, not auto-executed), but shell-quoting values would be more robust:
```ts
`--bot-token '${token}' --api-url '${apiUrl}'`
```

### P3 — `copyToClipboard` duplicated

Both `CreateModal.tsx` (line 21) and `DetailDrawer.tsx` (line 39) define identical clipboard fallback functions. Good candidate for export from a shared utils module alongside `connectGuide.ts`.

### P3 — `handleCopyGuide` in CreateModal recomputes the guide

`connectGuide` is already computed as a component variable. `handleCopyGuide` calls `buildConnectGuide(...)` again with identical params. Can simplify to `await copyToClipboard(connectGuide)`.

### P3 — `resp.id || values.id` — unreachable fallback

Per `AppBotCreateResp`, `id` is non-optional `string`. Simplify to `id: resp.id`.

## Agreement with other reviewers

- Allen's shell injection warning: agree, P3 ✅
- Allen's `copyToClipboard` duplication: agree ✅
- lml2468's auto-reveal side-effect note: agree the behavior is logically correct — the guide contains the real token, so revealing it in UI is consistent. A brief toast like "Token 已显示" would make the implicit consequence explicit.

Clean PR, good UX improvement. Merge-ready.

PR #54, opened 2026-05-26
feat(space): inline-edit space profile fields in admin UI

an9xyz:
## Summary

Wires up the new partial-update endpoints (`PUT /v1/manager/spaces/:id` from server PR #158, `PUT /v1/space/:id` from server PR #164) and exposes them as **inline editing** in the admin UI — no separate edit modal.

## API

- `src/api/space.ts` — `updateSpaceProfile(spaceId, req)` for `name` / `description` / `logo` / `join_mode` / `max_users`.
- `src/api/space-user.ts` — `updateSpaceUserProfile(spaceId, req)` for `name` / `description` / `logo` / `join_mode` / `preset_group_ids` (no `max_users`).

## UI

- **`InlineEditField`** — reusable per-field editor (text / textarea / number / select).
  - Always-visible pencil affordance.
  - Esc cancels; Enter commits (Ctrl/⌘+Enter for textarea).
  - Custom rune counter for textarea to avoid overlap with action buttons; antd `showCount` is intentionally not used.
  - No native `maxLength` — the browser attribute counts UTF-16 code units and truncates surrogate-pair input (emoji, etc.) before the server-side rune limit. Length is enforced solely via `validate` using `Array.from(...).length` to match the server's `utf8.RuneCountInString`.
- **`SpaceInfoPanel`** — card layout for the super-admin Drawer with status pill + copyable Space ID, Logo thumbnail (with fallback initial), combined member-count / max-users display, and field grouping.
- **`SpaceAdminLayout`** — collapsible \"空间信息\" panel for owner/admin self-service editing via the user endpoint. After a successful `name` / `logo` save, also patches the matching `mySpaces` entry in the auth store so `SpaceSwitcher` reflects the change immediately instead of waiting for the next `/space/my` refetch.

## Validation parity with server

- Trim + rune count for `name` (≤100), `description` (≤500), `logo` (≤200).
- `join_mode` ∈ {0, 1}.
- `max_users` ≥ 0; if > 0, client also rejects values below current `member_count` (matches the server's best-effort check, fails fast in the UI).
- `preset_group_ids` not yet exposed in the user UI — it isn't part of the detail GET response, so there's no value to round-trip. Can be added later if a dedicated preset-groups page is built.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run build` clean
- [ ] Super-admin flow: edit each field on an active space; verify Drawer + list both refresh; banned/disbanded spaces hide edit affordance
- [ ] Owner / admin flow: edit fields via `/space/:id`; verify SpaceSwitcher name updates immediately
- [ ] Idempotent save (same value) skips the request
- [ ] Validation: empty name, oversized name (including emoji-heavy input), `max_users` below member count
- [ ] Keyboard: Esc cancels, Enter / Ctrl+Enter commits

### DIFF
--- a/src/api/space-user.ts
+++ b/src/api/space-user.ts
@@ -73,6 +73,19 @@ export const getUser = (uid: string) => unwrap(api.get<UserInfo>(`/v1/users/${ui
 export const getSpaceUserDetail = (spaceId: string) =>
   unwrap(api.get<SpaceUserDetail>(`/v1/space/${spaceId}`))
 
+export interface SpaceUserProfileUpdateReq {
+  name?: string
+  description?: string
+  logo?: string
+  join_mode?: 0 | 1
+  preset_group_ids?: string
+}
+
+export const updateSpaceUserProfile = (
+  spaceId: string,
+  data: SpaceUserProfileUpdateReq,
+) => api.put(`/v1/space/${spaceId}`, data)
+
 export const listSpaceUserMembers = (
   spaceId: string,
   params: { page?: number; limit?: number } = {},
--- a/src/api/space.ts
+++ b/src/api/space.ts
@@ -100,6 +100,17 @@ export const dissolveSpace = (spaceId: string) =>
 export const updateSpaceStatus = (spaceId: string, status: 1 | 2) =>
   api.put(`/v1/manager/spaces/${spaceId}/status/${status}`)
 
+export interface SpaceProfileUpdateReq {
+  name?: string
+  description?: string
+  logo?: string
+  join_mode?: SpaceJoinMode
+  max_users?: number
+}
+
+export const updateSpaceProfile = (spaceId: string, data: SpaceProfileUpdateReq) =>
+  api.put(`/v1/manager/spaces/${spaceId}`, data)
+
 export interface SpaceMemberListParams extends PageParams {}
 
 export const listSpaceMembers = (spaceId: string, params: SpaceMemberListParams = {}) =>
--- a/src/api/space.ts
+++ b/src/api/space.ts
@@ -108,6 +108,10 @@ export interface SpaceProfileUpdateReq {
   max_users?: number
 }
 
+// 成员上限的产品规则：上界 50000（业务侧限制，避免极端值）；0 表示不限。
+// 创建与编辑必须使用同一上限，否则会出现 "可创建但无法再编辑回去" 的不一致。
+export const MAX_USERS_HARD_CAP = 50000
+
 export const updateSpaceProfile = (spaceId: string, data: SpaceProfileUpdateReq) =>
   api.put(`/v1/manager/spaces/${spaceId}`, data)
 
--- a/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
+++ b/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
@@ -3,6 +3,8 @@ import { Outlet, useNavigate, useParams, useLocation } from 'react-router-dom'
 import {
   Layout,
   Avatar,
+  Collapse,
+  Descriptions,
   Dropdown,
   Tooltip,
   Tabs,
@@ -19,6 +21,8 @@ import {
   DesktopOutlined,
   ArrowLeftOutlined,
 } from '@ant-design/icons'
+import InlineEditField from '../Spaces/InlineEditField'
+import { updateSpaceUserProfile } from '../../api/space-user'
 import { useAuthStore } from '../../store/auth'
 import { useFeatureStore } from '../../store/feature'
 import { useTheme, type Theme } from '../../hooks/useTheme'
@@ -82,6 +86,24 @@ export default function SpaceAdminLayout() {
   )
   const [detail, setDetail] = useState<SpaceUserDetail | null>(null)
   const [loading, setLoading] = useState(false)
+
+  // 仅 owner(2) / admin(1) 且空间正常状态(1) 时可编辑
+  const canEdit = !!detail && detail.role >= 1 && detail.status === 1
+  // 与服务端 modules/space/api_manager.go 中的字符上限保持一致。
+  const NAME_MAX = 100
+  const DESC_MAX = 500
+  const LOGO_MAX = 200
+  const runeCount = (s: string) => Array.from(s).length
+
+  const saveDetailField = async (
+    patch: Parameters<typeof updateSpaceUserProfile>[1],
+    apply: (d: SpaceUserDetail) => SpaceUserDetail,
+  ) => {
+    if (!detail) return
+    await updateSpaceUserProfile(detail.space_id, patch)
+    setDetail(apply(detail))
+    message.success('已保存')
+  }
 
   // 每次挂载/spaceId 变化时强拉 /space/my,校验权限,防止 mySpaces 过期显示已退出的空间
   useEffect(() => {
@@ -306,6 +328,115 @@ export default function SpaceAdminLayout() {
               )}
             </div>
 
+            <Collapse
+              ghost
+              size="small"
+              style={{ marginTop: 8, marginBottom: 12 }}
+              items={[
+                {
+                  key: 'info',
+                  label: <span style={{ color: 'var(--a-text-tertiary)' }}>空间信息</span>,
+                  children: (
+                    <Descriptions
+                      size="small"
+                      column={2}
+                      colon={false}
+                      labelStyle={{
+                        color: 'var(--a-text-tertiary)',
+                        fontSize: 12,
+                        fontWeight: 500,
+                        width: 80,
+                        padding: '6px 0',
+                      }}
+                      contentStyle={{ fontSize: 13, padding: '6px 0' }}
+                    >
+                      <Descriptions.Item label="名称">
+                        <InlineEditField
+                          kind="text"
+                          value={detail.name}
+                          readOnly={!canEdit}
+                          maxLength={NAME_MAX}
+                          validate={(v) => {
+                            const t = String(v).trim()
+                            if (!t) return '空间名称不能为空'
+                            if (runeCount(t) > NAME_MAX) return `空间名称不能超过 ${NAME_MAX} 个字符`
+                            return null
+                          }}
+                          onSave={(v) =>
+                            saveDetailField({ name: String(v) }, (d) => ({ ...d, name: String(v) }))
+                          }
+                        />
+                      </Descriptions.Item>
+                      <Descriptions.Item label="加入方式">
+                        <InlineEditField
+                          kind="select"
+                          value={detail.join_mode}
+                          readOnly={!canEdit}
+                          display={
+                            detail.join_mode === 0 ? (
+                              <span className="pill-outline neutral">直接加入</span>
+                            ) : (
+                              <span className="pill-outline warning">需审批</span>
+                            )
+                          }
+                          options={[
+                            { value: 0, label: '直接加入' },
+                            { value: 1, label: '需审批' },
+                          ]}
+                          onSave={(v) => {
+                            const jm = (Number(v) === 1 ? 1 : 0) as 0 | 1
+                            return saveDetailField({ join_mode: jm }, (d) => ({
+                              ...d,
+                              join_mode: jm,
+                            }))
+                          }}
+                        />
+                      </Descriptions.Item>
+                      <Descriptions.Item label="Logo" span={2}>
+                        <InlineEditField
+                          kind="text"
+                          value={detail.logo}
+                          readOnly={!canEdit}
+                          maxLength={LOGO_MAX}
+                          placeholder="https://..."
+                          emptyText="未设置"
+                          validate={(v) =>
+                            runeCount(String(v)) > LOGO_MAX
+                              ? `Logo 不能超过 ${LOGO_MAX} 个字符`
+                              : null
+                          }
+                          onSave={(v) =>
+                            saveDetailField({ logo: String(v) }, (d) => ({ ...d, logo: String(v) }))
+                          }
+                        />
+                      </Descriptions.Item>
+                      <Descriptions.Item label="简介" span={2}>
+                        <InlineEditField
+                          kind="textarea"
+                          value={detail.description}
+                          readOnly={!canEdit}
+                          maxLength={DESC_MAX}
+                          rows={3}
+                          emptyText="未填写"
+                          validate={(v) =>
+                            runeCount(String(v).trim()) > DESC_MAX
+                              ? `空间描述不能超过 ${DESC_MAX} 个字符`
+                              : null
+                          }
+                          onSave={(v) =>
+                            saveDetailField({ description: String(v) }, (d) => ({
+                              ...d,
+                              description: String(v),
+                            }))
+                          }
+                        />
+                      </Descriptions.Item>
+                    </Descriptions>
+                  ),
+                },
+              ]}
+            />
+
             <Tabs
               activeKey={activeTab}
               onChange={(k) => navigate(`/space/${spaceId}/${k}`)}
@@ -317,6 +448,7 @@ export default function SpaceAdminLayout() {
           </>
         )}
       </Content>
+
     </Layout>
   )
 }
--- a/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
+++ b/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
@@ -101,7 +101,18 @@ export default function SpaceAdminLayout() {
   ) => {
     if (!detail) return
     await updateSpaceUserProfile(detail.space_id, patch)
-    setDetail(apply(detail))
+    const next = apply(detail)
+    setDetail(next)
+    // 同步 mySpaces：name / logo 影响 SpaceSwitcher 显示，不同步则继续渲染旧值直到下次拉取 /space/my。
+    if (patch.name !== undefined || patch.logo !== undefined) {
+      setMySpaces(
+        mySpaces.map((s) =>
+          s.space_id === next.space_id
+            ? { ...s, name: next.name, logo: next.logo }
+            : s,
+        ),
+      )
+    }
     message.success('已保存')
   }
 
--- a/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
+++ b/src/pages/SpaceAdmin/SpaceAdminLayout.tsx
@@ -103,12 +103,17 @@ export default function SpaceAdminLayout() {
     await updateSpaceUserProfile(detail.space_id, patch)
     const next = apply(detail)
     setDetail(next)
-    // 同步 mySpaces：name / logo 影响 SpaceSwitcher 显示，不同步则继续渲染旧值直到下次拉取 /space/my。
-    if (patch.name !== undefined || patch.logo !== undefined) {
+    // 同步 mySpaces：name / logo 影响 SpaceSwitcher 显示；join_mode 当前不渲染但属于 MySpace
+    // 类型字段，一并同步避免下游消费者拿到陈旧值。不同步则继续渲染旧值直到下次拉取 /space/my。
+    if (
+      patch.name !== undefined ||
+      patch.logo !== undefined ||
+      patch.join_mode !== undefined
+    ) {
       setMySpaces(
         mySpaces.map((s) =>
           s.space_id === next.space_id
-            ? { ...s, name: next.name, logo: next.logo }
+            ? { ...s, name: next.name, logo: next.logo, join_mode: next.join_mode }
             : s,
         ),
       )
--- a/src/pages/Spaces/InlineEditField.tsx
+++ b/src/pages/Spaces/InlineEditField.tsx
@@ -184,7 +184,10 @@ export default function InlineEditField(props: InlineEditFieldProps) {
           onChange={(e) => setDraft(e.target.value)}
           onKeyDown={onKeyDown}
           placeholder={props.placeholder}
-          maxLength={props.maxLength}
+          // 故意不传 maxLength：浏览器按 UTF-16 code unit 计数，
+          // 会在 emoji / 罕见 CJK 等代理对字符上提前截断 (e.g. 50 emoji ≈ 100 code units)，
+          // 与服务端 utf8.RuneCountInString 的 rune 上限不一致。
+          // 计数与超限拦截统一交给 validate（rune 计数）处理。
           rows={props.rows ?? 3}
           disabled={saving}
           style={{ width: '100%' }}
@@ -223,7 +226,8 @@ export default function InlineEditField(props: InlineEditFieldProps) {
         onPressEnter={commit}
         onKeyDown={onKeyDown}
         placeholder={props.placeholder}
-        maxLength={props.maxLength}
+        // 不传 maxLength：浏览器按 UTF-16 code unit 计数，会在 emoji 等代理对字符上提前截断；
+        // 长度上限统一由 validate（rune 计数）兜底，与服务端 utf8.RuneCountInString 对齐。
         disabled={saving}
         size="small"
         style={{ minWidth: 200 }}
--- a/src/pages/Spaces/InlineEditField.tsx
+++ b/src/pages/Spaces/InlineEditField.tsx
@@ -1,0 +1,273 @@
+import { useEffect, useRef, useState, type KeyboardEvent, type ReactNode } from 'react'
+import { Input, InputNumber, Radio, Button, Space, Tooltip, message } from 'antd'
+import { EditOutlined, CheckOutlined, CloseOutlined } from '@ant-design/icons'
+
+export type InlineFieldKind = 'text' | 'textarea' | 'number' | 'select'
+
+export interface InlineFieldOption {
+  value: number | string
+  label: ReactNode
+}
+
+interface BaseProps {
+  /** 当前持久值；非编辑态下可由 display 覆盖渲染 */
+  value: string | number | null | undefined
+  /** 非编辑态展示的内容；不传则直接展示 value */
+  display?: ReactNode
+  /** 编辑态占位符 */
+  placeholder?: string
+  /** 字段层级的本地校验：返回非空字符串表示错误信息 */
+  validate?: (next: string | number) => string | null
+  /** 保存回调；抛错时弹窗内保留输入并显示错误 */
+  onSave: (next: string | number) => Promise<void>
+  /** 只读时不渲染编辑入口 */
+  readOnly?: boolean
+  /** value 为空时的占位文本 */
+  emptyText?: string
+}
+
+interface TextProps extends BaseProps {
+  kind: 'text'
+  maxLength?: number
+}
+
+interface TextareaProps extends BaseProps {
+  kind: 'textarea'
+  maxLength?: number
+  rows?: number
+}
+
+interface NumberProps extends BaseProps {
+  kind: 'number'
+  min?: number
+  max?: number
+}
+
+interface SelectProps extends BaseProps {
+  kind: 'select'
+  options: InlineFieldOption[]
+}
+
+export type InlineEditFieldProps = TextProps | TextareaProps | NumberProps | SelectProps
+
+// 与服务端 utf8.RuneCountInString 一致：按 code point 数。
+function runeLen(s: string): number {
+  return Array.from(s).length
+}
+
+export default function InlineEditField(props: InlineEditFieldProps) {
+  const { value, display, readOnly, emptyText = '—', validate, onSave } = props
+  const [editing, setEditing] = useState(false)
+  const [draft, setDraft] = useState<string | number>(value ?? '')
+  const [saving, setSaving] = useState(false)
+  const inputRef = useRef<HTMLElement | null>(null)
+
+  useEffect(() => {
+    if (editing) setDraft(value ?? '')
+  }, [editing, value])
+
+  // 自动聚焦编辑控件
+  useEffect(() => {
+    if (editing && inputRef.current && typeof inputRef.current.focus === 'function') {
+      // antd InputNumber/Input refs 类型不一致，统一兜底
+      ;(inputRef.current as HTMLInputElement).focus?.()
+    }
+  }, [editing])
+
+  const cancel = () => {
+    if (saving) return
+    setEditing(false)
+  }
+
+  const commit = async () => {
+    let next = draft
+    if (props.kind === 'text' || props.kind === 'textarea') {
+      next = typeof next === 'string' ? next.trim() : String(next ?? '')
+    }
+    // 与持久值相同则不发请求
+    const current =
+      props.kind === 'text' || props.kind === 'textarea'
+        ? typeof value === 'string'
+          ? value
+          : value == null
+            ? ''
+            : String(value)
+        : value
+    if (next === current || (next === '' && (current == null || current === ''))) {
+      setEditing(false)
+      return
+    }
+    if (validate) {
+      const err = validate(next)
+      if (err) {
+        message.error(err)
+        return
+      }
+    }
+    setSaving(true)
+    try {
+      await onSave(next)
+      setEditing(false)
+    } catch (error) {
+      message.error((error as Error).message)
+    } finally {
+      setSaving(false)
+    }
+  }
+
+  if (!editing) {
+    const hasValue = value !== undefined && value !== null && value !== ''
+    return (
+      <span className="inline-edit-display">
+        <span>{display ?? (hasValue ? value : <span style={{ color: 'var(--a-text-quaternary)' }}>{emptyText}</span>)}</span>
+        {!readOnly && (
+          <Tooltip title="编辑" mouseEnterDelay={0.3}>
+            <button
+              type="button"
+              aria-label="编辑"
+              className="inline-edit-trigger"
+              onClick={() => setEditing(true)}
+            >
+              <EditOutlined />
+            </button>
+          </Tooltip>
+        )}
+      </span>
+    )
+  }
+
+  // 通用键盘交互：Esc 取消；非 textarea 时 Enter 提交（已由 onPressEnter 处理）；
+  // textarea 用 Ctrl/Cmd+Enter 提交，避免与正常换行冲突。
+  const onKeyDown = (e: KeyboardEvent<HTMLElement>) => {
+    if (e.key === 'Escape') {
+      e.preventDefault()
+      cancel()
+      return
+    }
+    if (props.kind === 'textarea' && e.key === 'Enter' && (e.metaKey || e.ctrlKey)) {
+      e.preventDefault()
+      void commit()
+    }
+  }
+
+  const actions = (
+    <Space.Compact className="inline-edit-actions">
+      <Button
+        size="small"
+        type="primary"
+        icon={<CheckOutlined />}
+        loading={saving}
+        onClick={commit}
+      />
+      <Button
+        size="small"
+        icon={<CloseOutlined />}
+        disabled={saving}
+        onClick={cancel}
+      />
+    </Space.Compact>
+  )
+
+  if (props.kind === 'textarea') {
+    // textarea 走垂直布局：actions 落到右下角，更适合多行编辑。
+    // 自渲染字符数 —— antd showCount 会占用右下角与 actions 重叠，禁用之。
+    const text = (draft as string) ?? ''
+    const count = runeLen(text)
+    return (
+      <div className="inline-edit-textarea-wrap">
+        <Input.TextArea
+          ref={(el) => {
+            const ta = el as unknown as { resizableTextArea?: { textArea: HTMLElement } } | null
+            inputRef.current = ta?.resizableTextArea?.textArea ?? null
+          }}
+          value={text}
+          onChange={(e) => setDraft(e.target.value)}
+          onKeyDown={onKeyDown}
+          placeholder={props.placeholder}
+          maxLength={props.maxLength}
+          rows={props.rows ?? 3}
+          disabled={saving}
+          style={{ width: '100%' }}
+        />
+        <div className="inline-edit-textarea-footer">
+          {props.maxLength ? (
+            <span
+              className="inline-edit-textarea-count"
+              style={{
+                color:
+                  count > props.maxLength
+                    ? 'var(--a-danger)'
+                    : 'var(--a-text-quaternary)',
+              }}
+            >
+              {count} / {props.maxLength}
+            </span>
+          ) : (
+            <span />
+          )}
+          <Tooltip title="Ctrl/⌘+Enter 提交">{actions}</Tooltip>
+        </div>
+      </div>
+    )
+  }
+
+  let control: ReactNode = null
+  if (props.kind === 'text') {
+    control = (
+      <Input
+        ref={(el) => {
+          inputRef.current = el?.input ?? null
+        }}
+        value={(draft as string) ?? ''}
+        onChange={(e) => setDraft(e.target.value)}
+        onPressEnter={commit}
+        onKeyDown={onKeyDown}
+        placeholder={props.placeholder}
+        maxLength={props.maxLength}
+        disabled={saving}
+        size="small"
+        style={{ minWidth: 200 }}
+      />
+    )
+  } else if (props.kind === 'number') {
+    control = (
+      <InputNumber
+        ref={(el) => {
+          inputRef.current = (el as unknown as { input: HTMLElement | null })?.input ?? null
+        }}
+        value={typeof draft === 'number' ? draft : Number(draft) || 0}
+        onChange={(v) => setDraft(typeof v === 'number' ? v : 0)}
+        onPressEnter={commit}
+        onKeyDown={onKeyDown}
+        min={props.min}
+        max={props.max}
+        disabled={saving}
+        size="small"
+        style={{ width: 140 }}
+      />
+    )
+  } else if (props.kind === 'select') {
+    control = (
+      <span onKeyDown={onKeyDown} style={{ display: 'inline-flex' }}>
+        <Radio.Group
+          value={draft}
+          onChange={(e) => setDraft(e.target.value)}
+          disabled={saving}
+        >
+          {props.options.map((opt) => (
+            <Radio key={String(opt.value)} value={opt.value}>
+              {opt.label}
+            </Radio>
+          ))}
+        </Radio.Group>
+      </span>
+    )
+  }
+
+  return (
+    <Space align="center" size={8} wrap>
+      {control}
+      {actions}
+    </Space>
+  )
+}
--- a/src/pages/Spaces/InlineEditField.tsx
+++ b/src/pages/Spaces/InlineEditField.tsx
@@ -235,10 +235,37 @@ export default function InlineEditField(props: InlineEditFieldProps) {
       />
     )
   } else if (props.kind === 'number') {
-    // 非负整数（min >= 0）时连 '-' 一起剥离，杜绝非数字输入；
-    // 允许负值的场景才保留前导 '-'。小数点、字母、空格等始终拒。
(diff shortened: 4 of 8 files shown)


Jerry-Xin (dismissed):
The PR is in scope for `octo-admin` and cleanly wires inline editing into the existing space admin flows; I found no blocking correctness or security issues.

💬 Non-blocking

🟡 Warning — `src/pages/Spaces/SpaceInfoPanel.tsx:178` / `src/pages/Spaces/SpaceInfoPanel.tsx:186`: `max_users` accepts decimal values through `InputNumber`, but the API field is effectively an integer count. Consider adding an integer check or `precision={0}` in `InlineEditField` for number fields to avoid avoidable server-side validation failures.

🔵 Suggestion — `src/pages/Spaces/InlineEditField.tsx:82`: `commit` does not early-return when `saving` is already true. Ant Design’s loading state reduces the risk, but a rapid double-submit or repeated Enter key can still race before React applies state. A simple `if (saving) return` would make saves idempotent at the component level.

🔵 Suggestion — `src/pages/Spaces/SpaceInfoPanel.tsx:51`: the super-admin save path closes the editor but does not show success feedback, while the space-admin path does. Consider making the feedback consistent.

🔵 Suggestion — There are no focused tests for `InlineEditField` keyboard behavior, idempotent save skipping, rune-count validation, or integer validation. These are good candidates for small component/unit tests.

✅ Highlights

- Good scope alignment: the PR extends existing space profile management UI and API clients.
- Validation mostly mirrors the server constraints, including code-point length checks for emoji-heavy input.
- Banned/disbanded spaces correctly hide edit affordances through read-only state.
- `npm run build` passes after installing locked dependencies with `npm ci`.

yujiawei (dismissed):
# Code Review — PR #54 (octo-admin)

## Summary

Wires up the new partial-update endpoints (`PUT /v1/manager/spaces/:id`, `PUT /v1/space/:id`) and surfaces them as inline editing in the admin UI via a new reusable `InlineEditField` plus two consumers (`SpaceInfoPanel` for super-admin, `SpaceAdminLayout` for owner/admin). No new authority surface — these endpoints already exist server-side and the UI only exposes affordances that the server is already permission-gated for. The change is well-scoped and self-contained; the security-sensitive label is conservative rather than reflecting a real new boundary.

## Verdict

No blocking issues found. A few low-priority maintainability/UX nits below — all non-blocking.

## What was reviewed

- New API clients: `src/api/space.ts` (`updateSpaceProfile`), `src/api/space-user.ts` (`updateSpaceUserProfile`).
- New components: `src/pages/Spaces/InlineEditField.tsx`, `src/pages/Spaces/SpaceInfoPanel.tsx`.
- Consumer wiring: `src/pages/Spaces/SpaceDetailDrawer.tsx`, `src/pages/Spaces/index.tsx`, `src/pages/SpaceAdmin/SpaceAdminLayout.tsx`.
- New CSS: `src/styles/admin.css` additions for inline edit + space info card.

## Strengths

- **Validation parity with server**: `runeLen = Array.from(s).length` correctly mirrors Go's `utf8.RuneCountInString`, and the deliberate avoidance of the browser `maxLength` attribute is well-documented in `InlineEditField.tsx:187-190` and `:229-230`. This is the right call — `maxLength` on `<input>` counts UTF-16 code units and would silently truncate emoji / surrogate-pair input before the server ever sees it, producing values that pass client and fail server (or vice versa). The inline comments explain *why* clearly.
- **Idempotent skip**: `InlineEditField.tsx:88-99` short-circuits the request when the trimmed draft equals the persisted value, avoiding gratuitous PUTs on accidental focus-and-commit. Good.
- **Edit gating**: `SpaceInfoPanel.tsx:48` (`editable = space.status === 1`) and `SpaceAdminLayout.tsx:91` (`canEdit = role >= 1 && status === 1`) both check the active-status precondition, so banned/disbanded spaces correctly hide the edit affordances. The server is still authoritative; the UI is honest about what it claims is possible.
- **Endpoint surface parity**: super-admin `SpaceProfileUpdateReq` includes `max_users`; owner/admin `SpaceUserProfileUpdateReq` does not — matches the server's permission split. `preset_group_ids` is exposed in the type but intentionally not surfaced in the UI, with a reasonable justification in the PR body.
- **`max_users` client pre-check**: `SpaceInfoPanel.tsx:178-185` fails fast when the requested cap would be below `member_count`, with the correct framing ("best-effort, server is authoritative"). Good UX without pretending the client is the source of truth.
- **`mySpaces` sync after save**: `SpaceAdminLayout.tsx:107-115` patches the matching entry in the auth store so `SpaceSwitcher` doesn't display a stale name/logo until the next `/space/my` refetch. Currently only `name` is actually visible in `SpaceSwitcher`, but syncing `logo` too is harmless and defensive.
- **Keyboard ergonomics**: Esc cancels, Enter commits (Ctrl/⌘+Enter for textarea to avoid clashing with newlines). The custom rune counter avoids overlap with the action buttons that antd's `showCount` would cause — the trade-off is called out and reasonable.
- **No XSS surface introduced**: no `dangerouslySetInnerHTML`, no string-templated URLs, no `console.*`/`debugger` leftovers. The user-controlled `logo` field is only rendered via `` — modern browsers do not execute scripts in images (including SVG loaded via ``), and `javascript:` URLs in `` are no-ops, so this is safe.

## Security review (per `security_sensitive` routing)

I looked specifically for new auth/middleware/credential surface, since that's what triggered the routing:

- **No new auth/credential code introduced.** This PR consumes two already-existing server endpoints; it does not move any auth, middleware, secret, or token logic.
- **Authorization is server-enforced**, not client-trusted. The UI's `editable` / `canEdit` flags are purely cosmetic — the server still validates the caller's role against the target space on every PUT.
- **`logo` URL is user-controlled** but rendered only via ``. As above, this is not an XSS vector under any current browser. If the server stores it without sanitization (e.g. `javascript:foo` text in the field), that would render as a broken image rather than an exploit. Worth a one-line server-side scheme check (`http`/`https`/`data:image`) if not already present, but that's a server concern, not this PR's.
- **`max_users` lowering**: the client refuses to PUT a value below current `member_count`, and the PR body notes the server does the same "best-effort" check. This is defense in depth; the server check is the one that matters.
- **No CSRF concern introduced**: the project's existing auth header strategy (Bearer token via axios interceptor) carries over unchanged.

Nothing here that I'd want a human to manually verify beyond the standard "exercise the flows in staging" pass listed in the PR's Test plan checkboxes that are still unchecked.

## Findings

### P2 — CSS rules not scoped under `.admin-shell`

`src/styles/admin.css:935-1091` adds the new `.inline-edit-*` and `.space-info-*` rule blocks at the top level, unscoped. The file header explicitly states the contract: *"Every rule is scoped under .admin-shell so /changelog is untouched."* These new selectors break that contract.

There is no observable bleed today, since the components are only used inside the admin shell. But the constraint exists to keep the changelog page (and any future non-admin consumer of the same stylesheet) immune to accidental leak. Consider prefixing the new selectors, e.g. `.admin-shell .inline-edit-display { ... }`.

### P2 — Auto-focus is a no-op for `number` and `select` kinds

`InlineEditField.tsx:70-75` sets up an effect to focus `inputRef.current` when editing flips on, but:

- **`number` (InputNumber)**: `inputRef.current = (el as unknown as { input: HTMLElement | null })?.input ?? null` (`:240`). Antd's `InputNumberRef` does not expose `.input` directly — it exposes `focus()`, `blur()`, `nativeElement`. So `inputRef.current` is null and the focus effect is a no-op.
- **`select` (Radio.Group)**: no `ref` is attached at all (`:253-268`), so `inputRef.current` stays at whatever the previous kind left it (or null on first edit), and focus does nothing.

Net effect: opening edit on a name/textarea field auto-focuses (good), but opening edit on `max_users` or `join_mode` requires an extra click into the control. Low-priority UX improvement.

A minimal fix would be calling `el?.focus()` directly on the antd ref for `number`, and either attaching a ref to the first `<Radio>` or making focus a no-op explicitly for `select` (since Radio.Group focusing is already keyboard-reachable via Tab).

### P2 — In-flight saves on unmount

`InlineEditField.tsx:107-115`: if the user closes the Drawer (`destroyOnClose`) while a `PUT` is in flight, the `.finally()` setState calls will fire on an unmounted component. React logs a dev warning but it isn't a functional bug — the PUT still completes server-side, which is the correct outcome (no silent data loss).

Optional: track `mounted` via a ref or `AbortController` to silence the warning. Non-blocking.

### Nit — empty-name commit can skip validation

`InlineEditField.tsx:96`: the idempotent-skip check fires when `next === ''` and the persisted `value` is also empty (`current == null || current === ''`). If a space somehow already has an empty name (server should reject this, so it shouldn't happen in practice), pressing commit with an empty draft would silently no-op rather than surface the "空间名称不能为空" error from `validate`. The current ordering means validate is bypassed when the skip condition wins.

Since the server rejects empty names on create/update, this is an edge case that should never occur with real data. Mentioning for completeness — happy to leave as-is.

### Nit — stale `mySpaces.join_mode` after edit

`SpaceAdminLayout.tsx:107` only patches `mySpaces` when `patch.name !== undefined || patch.logo !== undefined`. `MySpace.join_mode` is part of the type and would go stale after a `join_mode` change until the next `/space/my` refetch. Today `SpaceSwitcher` doesn't render `join_mode`, so this is invisible — but if `SpaceSwitcher` (or any other consumer of `mySpaces`) starts displaying it later, this becomes a latent bug.

Either sync `join_mode` too, or add a comment explaining the intentional omission.

## Build / type-check

- `npx tsc --noEmit` is clean for the files in this PR. There is one pre-existing TS error in `src/store/feature.test.ts` (`Cannot find module 'vitest'`) that has been on `main` since `e570a94` and is unrelated to this change.

## Test plan status

The PR's test plan correctly identifies the manual flows that need exercising before merge:
- Super-admin field edits across active / banned / disbanded spaces.
- Owner / admin edits with `SpaceSwitcher` reflecting name updates immediately.
- Idempotent same-value save.
- Validation: empty name, oversized name (emoji-heavy), `max_users` below member count.
- Keyboard: Esc / Enter / Ctrl+Enter.

Those checkboxes are still unchecked in the PR body; please tick them off (or call out which couldn't be exercised) before landing.

lml2468 (dismissed):
The PR looks sound for the profile-editing flow and I did not find a blocking correctness or security issue in the changed call paths.

Blocking items: None.

Non-blocking notes:
1. `src/styles/admin.css` adds the new `.inline-edit-*` and `.space-info-*` rules without the existing `.admin-shell` prefix used by the rest of this stylesheet. The current class names are specific enough for this page, but scoping these selectors would keep the file's original isolation contract intact.
2. The inline editors allow multiple fields to be edited/saved at the same time, while `SpaceInfoPanel.save` and `SpaceAdminLayout.saveDetailField` both apply updates from the current render snapshot. If two saves overlap, the backend patch calls should still persist, but the local drawer/header state can briefly show the older value from the later-resolving snapshot until the next refresh. Not worth blocking this PR, but a functional updater or post-save refetch would make the state model more robust.

Highlights:
1. The new API wrappers are only used from the intended manager and space-scoped admin surfaces, and the changed components have no dangling references.
2. Validation covers the user-editable text limits and uses code-point counting to match the documented backend behavior instead of browser UTF-16 `maxLength`.
3. Failed saves stay in edit mode and surface the API error rather than silently reverting or pretending the update succeeded.

lml2468 (dismissed):
Approved: the incremental changes do not introduce a new blocking bug or security issue.

Previous findings status:
- Resolved: the new inline-edit and space-info CSS selectors are now scoped under `.admin-shell`, preserving the admin stylesheet isolation contract.
- Still open, non-blocking: `SpaceInfoPanel.save()` still calls `onSpaceChange({ ...space, [field]: value })` after awaiting the request, so overlapping saves across different fields can still apply over a stale `space` prop snapshot. This remains the same P2 risk from the prior round and is not a merge blocker by itself.

New issues:
- None found in the incremental diff.

Non-blocking notes:
- The new commit also addresses previously raised reviewer concerns around integer-only `InputNumber`, duplicate-save guarding, and super-admin save success feedback.
- Existing review-thread concerns about test coverage are not repeated here.

yujiawei (dismissed):
# Code Review — PR #54 (octo-admin)

## Summary

Solid, well-scoped frontend wiring of the new partial-update endpoints. The split between `updateSpaceProfile` (manager) and `updateSpaceUserProfile` (user) cleanly mirrors the server, the rune-count validation is deliberately matched to `utf8.RuneCountInString` (Go) via `Array.from(s).length` (UTF-16 surrogate-aware), and the idempotent-save short-circuit keeps the UX quiet for no-op edits. No P0/P1 blockers found. Several P2 nits and one item I'd flag for a human verifier given the security-sensitive label.

**Verdict: APPROVED** — comments below are nits / suggestions, not gating.

## 1. Verification

| Item | Status | Evidence |
| --- | --- | --- |
| Manager vs user endpoints split correctly | ✅ | `src/api/space.ts:111` calls `/v1/manager/spaces/${id}`; `src/api/space-user.ts:84-87` calls `/v1/space/${id}` |
| `max_users` only in manager payload | ✅ | `SpaceProfileUpdateReq` (space.ts:103-109) includes `max_users`; `SpaceUserProfileUpdateReq` (space-user.ts:76-82) intentionally omits it, matching server PR #164 contract per PR description |
| Length caps match server (name=100, desc=500, logo=200) | ✅ | Constants in `SpaceInfoPanel.tsx:12-14` and `SpaceAdminLayout.tsx:93-95` align; rune count via `Array.from(s).length` matches Go `utf8.RuneCountInString` |
| Status-gated editability | ✅ | Manager view: `editable = space.status === 1` (`SpaceInfoPanel.tsx:48`). User view: `canEdit = detail.role >= 1 && detail.status === 1` (`SpaceAdminLayout.tsx:91`) — banned/disbanded spaces hide the pencil affordance |
| `mySpaces` sync after self-service edit | ✅ | `SpaceAdminLayout.tsx:108-120` patches name/logo/join_mode in the auth store so `SpaceSwitcher` reflects immediately |
| Idempotent save | ✅ | `InlineEditField.tsx:90-101` short-circuits when trimmed draft equals current (string) or numeric draft equals current value, including the empty/null equivalence case |
| `tsc --noEmit` / `npm run build` | ⚠️ Trusted | PR description claims clean; couldn't run locally (no workspace checkout for this repo), but the discriminated-union props on `InlineEditField` and the explicit `SpaceJoinMode` cast read as type-safe |

## 2. Findings

### P2 — `setLogoBroken(false)` runs before the network call resolves (UX nit)

`src/pages/Spaces/SpaceInfoPanel.tsx:104-107`

```ts
onSave={(v) => {
  setLogoBroken(false)
  return save('logo', String(v))
}}
```

If the PUT rejects (e.g. server validation), the broken-image fallback is already cleared and the `` is re-attempted with the same broken URL the user just tried to save. Minor visual flicker on failure, but functionally harmless because the `` `onError` handler will flip it back. Cleaner: reset `logoBroken` only after `save(...)` resolves, e.g.

```ts
onSave={async (v) => {
  await save('logo', String(v))
  setLogoBroken(false)
}}
```

### P2 — Idempotent-save check normalizes whitespace silently

`src/pages/Spaces/InlineEditField.tsx:85-101`

For text/textarea kinds, the draft is `.trim()`ed before being compared to `value`. If the persisted value happens to have leading/trailing whitespace (legacy data, server didn't trim on a previous version), entering edit mode and pressing Enter without changes will fail the equality check (`'foo' !== 'foo '`) and silently fire a PUT that just trims. Not incorrect — arguably a desirable normalization — but worth being aware of: a user who clicks edit and immediately commits is no longer guaranteed a no-op. If you'd rather avoid the surprise PUT, compare the trimmed draft against the trimmed current value.

### P2 — `validate` on `max_users` doesn't catch `NaN`

`src/pages/Spaces/SpaceInfoPanel.tsx:179-186`

```ts
validate={(v) => {
  const n = Number(v)
  if (n < 0) return '成员上限不能为负'
  if (n > 0 && n < space.member_count) { ... }
  return null
}}
```

`Number('abc')` is `NaN`; `NaN < 0` is `false` and `NaN > 0` is `false`, so a `NaN` would slip through to `save('max_users', NaN)` and reach the server. In practice antd `InputNumber` plus the `setDraft(typeof v === 'number' ? v : 0)` fallback in `InlineEditField.tsx:244` makes this unreachable through normal interaction, so this is defense-in-depth, not a real bug. Adding `if (!Number.isFinite(n)) return '请输入有效数字'` at the top is cheap insurance.

### P2 — Clearing the number input snaps to `0`, not the original value

`src/pages/Spaces/InlineEditField.tsx:243-244`

```ts
value={typeof draft === 'number' ? draft : Number(draft) || 0}
onChange={(v) => setDraft(typeof v === 'number' ? v : 0)}
```

When the user backspaces the InputNumber empty, `draft` becomes `0`. Combined with the existing semantics (`max_users === 0` means "不限"), pressing Enter at that point silently flips a capped space to "unlimited". The validator allows this (it's a legal value). I'd consider one of:

- Keep `draft` as the previous value when the input is cleared (only commit on explicit user edit), or
- Add a `confirmZero` prompt for the specific `max_users` field, or at minimum
- Document that 0 = unlimited near the input.

Not a correctness bug; just a footgun for an admin field.

### P2 — In-progress edit can be silently overwritten by parent re-renders

`src/pages/Spaces/InlineEditField.tsx:66-68`

```ts
useEffect(() => {
  if (editing) setDraft(value ?? '')
}, [editing, value])
```

If the parent re-renders with a new `value` while the user is mid-edit (e.g. another field's save resolves and the parent rebuilds), the in-progress draft is replaced. Not exploitable through this PR's flows (the user can only edit one field at a time and `disabled={saving}` blocks input during the save itself), but worth a one-line guard:

```ts
useEffect(() => {
  if (editing && !saving) setDraft(value ?? '')
}, [editing])  // intentionally drop `value` from deps once editing started
```

### P2 — Logo `` lacks `referrerPolicy` (security-sensitive flag)

`src/pages/Spaces/SpaceInfoPanel.tsx:80-85`

```tsx
<img src={space.logo} alt="" className="space-info-logo-thumb" onError={...} />
```

A space owner can set `logo` to any URL. When the super-admin opens the drawer, the browser sends a `Referer: https://<admin-host>/...` header to that arbitrary host, leaking the admin domain (and potentially path) to attacker-controlled image servers. Same goes for the user-side layout if/when a logo thumbnail is rendered there. Cheap mitigation:

```tsx
<img src={space.logo} alt="" referrerPolicy="no-referrer" loading="lazy" ... />
```

This is the kind of item I'd want a human reviewer to weigh given the `needs-human-review` label — not a blocker, but an easy win for an admin surface.

### P2 — Server-side authorization is the only real gate (security-sensitive flag)

`canEdit` / `editable` are purely UX. `PUT /v1/manager/spaces/:id` and `PUT /v1/space/:id` (server PRs #158/#164) MUST enforce that the caller is super-admin / owner-or-admin respectively, and that the space is in status=1 before allowing mutating fields. Worth a human verifier double-checking the corresponding handler tests on the server side, since this PR has no test coverage of its own.

### P2 — `Tag` import in `SpaceAdminLayout.tsx` is now redundant for the inline-edit panel but still used by the meta header — kept on purpose; flagging only because the file's import surface grew

No action.

### P3 — Stray empty line introduced

`src/pages/SpaceAdmin/SpaceAdminLayout.tsx:467` adds a blank line between `</Content>` and `</Layout>`. Cosmetic.

## 3. Suggestions

1. (P2) Reset `logoBroken` after `save(...)` resolves, not before.
2. (P2) Add `referrerPolicy="no-referrer"` (and ideally `loading="lazy"`) to the logo ``.
3. (P2) Guard the `max_users` validator against `NaN` and consider warning when an admin clears the field to `0` (= unlimited).
4. (P2) Skip the auto-PUT-on-no-op when the only delta is trim-only whitespace, or compare trimmed-vs-trimmed.
5. (P2) Drop `value` from the `useEffect` deps inside `InlineEditField` once `editing === true`, so concurrent parent re-renders don't blow away an in-progress draft.
6. (P3) Consider one canonical `runeCount` helper exported from a util file rather than copy-pasted into `InlineEditField`, `SpaceInfoPanel`, and `SpaceAdminLayout`.
7. (P3) Add at least one unit test around `commit()` covering: trim → idempotent skip; rune-count over-limit on emoji-heavy input; numeric `0` vs non-zero diff. Easy regression net for the next time these fields grow constraints.

## 4. Additional observations

- The `setMySpaces(...)` patch in `SpaceAdminLayout.tsx:113-119` overwrites all three of `name`/`logo`/`join_mode` on every patched field, not just the changed one. Because `next` is built from `apply(detail)` (which only mutates the patched field), the other two values round-trip unchanged — this is correct, just non-obvious.
- `Tag` color `gold` for "需审批" in the meta header (`SpaceAdminLayout.tsx:316-319`) and the `pill-outline warning` for the same state inside the inline-edit panel (`SpaceAdminLayout.tsx:393-396`) are visually slightly different. Worth a quick consistency pass at some point but not in this PR.
- The CSS file's `.admin-shell`-scoped rules continue the established isolation contract — good.

Jerry-Xin (dismissed):
# Code Review — PR #54 (re-review of 1ab914b)

## Summary

The new commit addresses all substantive non-blocking suggestions from the initial reviews. Code is clean and well-structured.

## Previous Suggestions — Status

| Suggestion | Status |
|---|---|
| `max_users` InputNumber needs `precision={0}` | ✅ Resolved — `precision={0}` + `step={1}` set on the number input |
| `commit()` missing double-submit guard | ✅ Resolved — early `if (saving) return` at top of commit() |
| Super-admin save path missing success feedback | ✅ Resolved — `message.success('已保存')` in both SpaceInfoPanel and SpaceAdminLayout save paths |
| CSS selectors should be scoped under `.admin-shell` | ✅ Resolved — all `.inline-edit-*` and `.space-info-*` now nested under `.admin-shell` |
| Unit tests for InlineEditField | Deferred — reasonable given the manual test plan and component maturity |

## ✅ Highlights

- Clean `InlineEditField` abstraction: keyboard a11y (Esc/Enter/Ctrl+Enter), focus management, saving-guard, idempotent-skip
- Rune-count validation correctly uses `Array.from(...).length` to match server-side `utf8.RuneCountInString`
- Intentionally avoids native `maxLength` (UTF-16 code-unit mismatch with server rune limit) — well-documented in comments
- SpaceSwitcher store sync after name/logo save prevents stale UI
- CSS scoped under `.admin-shell` maintains isolation contract

No blocking issues found.

lml2468 (approved):
Approved: the incremental changes address the numeric input validation risk without introducing a new blocking issue.

Previous findings status:
- Still open, non-blocking: `SpaceInfoPanel.save` still applies updates over the captured `space` prop snapshot with `{ ...space, [field]: value }`, so overlapping field saves can still overwrite a newer local field value.
- Resolved: saving a new logo now resets `logoBroken` before applying the save.
- Still open, non-blocking: text/textarea idempotent-save comparison still compares the trimmed draft against the raw persisted value, so values that only differ by surrounding whitespace can still send an unnecessary PUT.
- Resolved: `max_users` now avoids the NaN path through the `InputNumber` parser/onChange fallback and adds the hard-cap validation.

New issues: none found in the incremental diff.

Non-blocking notes: Node 18/20 builds and CodeQL are passing on the current head; the queued `code-review` workflow was still pending when I checked.

lml2468 (dismissed):
Approved for current head `68212e5e23523670d6dfffc00a3e4d705fb4c052`: I reviewed both the requested `1ab914b...47e3498` delta and the extra `47e3498...68212e5` delta that landed while the review was in progress, and found no new blocking issues.

Previous findings status:
- Still open, non-blocking: `SpaceInfoPanel.save` still applies updates over the captured `space` prop snapshot with `{ ...space, [field]: value }`, so overlapping field saves can still overwrite a newer local field value.
- Resolved: saving a new logo now resets `logoBroken` before applying the save.
- Still open, non-blocking: text/textarea idempotent-save comparison still compares the trimmed draft against the raw persisted value, so values that only differ by surrounding whitespace can still send an unnecessary PUT.
- Resolved: `max_users` now avoids the NaN path through the `InputNumber` parser/onChange fallback, adds hard-cap validation, and the latest commit clamps pasted/system input to the configured min/max.

New issues: none found in the incremental diffs.

Non-blocking notes: Node 18/20 builds, CodeQL, changed-path detection, check-sprint, label, and code-review checks are passing on the current head.

Jerry-Xin (dismissed):
This PR is in scope for `octo-admin` and the inline editing implementation looks mergeable; I found only non-blocking polish and consistency issues.

**💬 Non-Blocking**
- 🟡 Warning: [src/pages/Spaces/SpaceInfoPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-_u5yh8sn/repo/src/pages/Spaces/SpaceInfoPanel.tsx:15) adds a client-only `max_users` hard cap of `50000`, but the create form does not enforce the same cap and the PR’s validation-parity notes only mention `>= 0` plus member-count checks. If the server allows larger values, super-admins can create values they later cannot edit back to another valid large value.
- 🔵 Suggestion: [src/pages/Spaces/SpaceInfoPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-_u5yh8sn/repo/src/pages/Spaces/SpaceInfoPanel.tsx:53) updates only the edited field locally, so displayed metadata such as `updated_at` remains stale after a successful save. Consider refetching the drawer detail or accepting the server response if the endpoint returns the updated object.
- 🔵 Suggestion: [src/pages/SpaceAdmin/SpaceAdminLayout.tsx](/Users/mlamp/.openclaw/tmp/pr-review-_u5yh8sn/repo/src/pages/SpaceAdmin/SpaceAdminLayout.tsx:98) has the same local-only update behavior for the owner/admin view, so metadata can also remain stale there.

**✅ Highlights**
- The PR correctly hides edit affordances for non-active spaces and uses the user endpoint separately from the manager endpoint.
- Text length validation intentionally avoids native `maxLength`, which matches the server-side rune-count behavior better for emoji and other surrogate-pair input.
- `npm run build` passed.
- `npm test` passed: 1 test file, 9 tests.

yujiawei (dismissed):
# Code Review — PR #54 (octo-admin)

Scope: 8 files / +908 / -68. Wires up `PUT /v1/manager/spaces/:id` (manager) and `PUT /v1/space/:id` (owner/admin self-service) and exposes them as inline edit on the super-admin Drawer and the Space-admin layout. Build (`tsc --noEmit` + `vite build`) and existing test suite are green.

## Summary verdict

No blocking correctness or security issues found in the diff. The structural choices are sound — split between the manager-side `SpaceInfoPanel` and the owner/admin-side info panel embedded in `SpaceAdminLayout`, separate API surfaces matching server roles, idempotent saves, an explicit reentrancy guard, and rune-count parity with `utf8.RuneCountInString`. A small set of P2 nits and one item I'd ask a human to verify on the server side are listed below.

## What looks right

- **Privilege separation by endpoint.** `updateSpaceProfile` (manager) accepts `max_users`; `updateSpaceUserProfile` (owner/admin self-service) intentionally does not — `SpaceUserProfileUpdateReq` is a strict subset (`name | description | logo | join_mode | preset_group_ids`). This matches the principle that raising a space's member cap is a manager-level action. (`src/api/space.ts:103-112`, `src/api/space-user.ts:76-87`)
- **Client gating respects the server contract, not the other way around.** `editable = space.status === 1` in `SpaceInfoPanel.tsx:50` and `canEdit = !!detail && detail.role >= 1 && detail.status === 1` in `SpaceAdminLayout.tsx:91` only hide UI; the server remains authoritative. The role check is duplicated in the user layout (mySpaces filter at `SpaceAdminLayout.tsx:136-142`), which is good defense-in-depth.
- **UTF-8 rune-count parity.** Frontend deliberately omits the native `maxLength` (which counts UTF-16 code units and truncates surrogate pairs early), enforcing length only via `Array.from(s).length` to match the server's `utf8.RuneCountInString`. The comments at `InlineEditField.tsx:188-191` and `:230-231` capture exactly the right reasoning.
- **Idempotent commit.** `InlineEditField.commit()` short-circuits when the new value equals the current value, and when both are empty (`InlineEditField.tsx:90-101`). Prevents redundant requests when users open edit just to look.
- **Reentrancy guard.** The `saving` flag at `InlineEditField.tsx:62` blocks both Enter-spamming and concurrent commits.
- **`mySpaces` cache sync after save.** `saveDetailField` in `SpaceAdminLayout.tsx:98-122` patches the local `mySpaces` entry for `name | logo | join_mode` so `SpaceSwitcher` reflects the change immediately, without waiting for the next `/space/my` refetch. Correctly conditional — only fires when one of those keys is in the patch.
- **Cleanup of duplicated state in the manager Drawer.** Moving the previous in-Drawer `Descriptions` block into a dedicated `SpaceInfoPanel` and routing list refresh through `onUpdated → fetchData()` (`SpaceDetailDrawer.tsx`, `Spaces/index.tsx:427`) keeps the source of truth single.

## P2 — nits / follow-ups

1. **`commit()`: `current` is not normalized before equality check** (`InlineEditField.tsx:82-101`). For text/textarea, `next` is `draft.trim()` but `current` is the raw stored `value`. If the persisted value happens to have leading/trailing whitespace (legacy data, or a future server-side bug), opening edit and pressing Enter without changes will trip the diff and fire a spurious PUT. Cheap fix: normalize `current` the same way you normalize `next` before comparing.
2. **`useEffect([editing, value])` overwrites the draft mid-edit** (`InlineEditField.tsx:66-68`). If a parallel write lands while the user is typing (drawer reopened, parent state pushed an updated `space`, etc.), the draft is silently replaced. Not exploitable, but a quiet UX loss. Consider only resetting on the `editing: false → true` edge.
3. **`save<K extends keyof Space>` is too permissive at the type level** (`SpaceInfoPanel.tsx:53-60`). The cast to `Parameters<typeof updateSpaceProfile>[1]` lets you call `save('member_count', 5)` at compile time even though the server PUT has no such field. Tightening `K` to `keyof SpaceProfileUpdateReq` would catch any future caller that strays.
4. **Latent off-by-when-negative in the keystroke clamp** (`InlineEditField.tsx:258-266`). `projected.replace(/[^\d]/g, '')` strips the leading `-` before the `> maxCap` compare, so when `allowNegative === true` the projected magnitude is what's checked, not the signed value. The current single caller (`max_users`) uses `min: 0` so this branch isn't exercised, but it'll bite the first caller with a negative range.
5. **Inconsistent `join_mode` clamp.** `SpaceAdminLayout.tsx:402-407` clamps `Number(v) === 1 ? 1 : 0`, while `SpaceInfoPanel.tsx:148` uses `Number(v) as SpaceJoinMode` with no runtime clamp. Both rely on `Radio.Group` emitting only the option values, which is true today. Same defensive cast in both places would be cleaner.
6. **Duplicate display on the owner/admin layout.** `SpaceAdminLayout.tsx:311-345` (the `meta-inline` strip) still shows `name`, `join_mode` tag, member count, and description, and the new "空间信息" Collapse below shows the same fields again. Functional, but the page now reads as two parallel headers. Worth a future polish pass.
7. **Trailing blank line before `</Layout>`** at `SpaceAdminLayout.tsx:467`. Cosmetic.

## P1 — please verify on the server side (security_sensitive ask)

Frontend can't prove these on its own; flagging for human review:

- `MAX_USERS_HARD_CAP = 50000` is a frontend-only constant (`SpaceInfoPanel.tsx:16`). Confirm the server-side validator on `PUT /v1/manager/spaces/:id` enforces a matching (or lower) ceiling. If the server cap differs, the UI either rejects valid values (too low) or sends invalid ones the server bounces (too high).
- Confirm `PUT /v1/space/:id` server-side rejects `max_users` (and any other manager-only key) when the body includes it — frontend doesn't include it, but a hand-crafted request from a user with `role ∈ {1, 2}` should not be able to escalate via that endpoint.
- `space.logo` is rendered as `` (`SpaceInfoPanel.tsx:82-87`). React auto-escapes the URL, and modern browsers don't execute `javascript:` in ``, but the field is unvalidated for scheme on either side. Recommend the server reject anything that's not `http(s)://` or empty so that admin-page Referer leakage and weird `data:` payloads are excluded at the source.
- Confirm the server enforces the same rune limits (name ≤ 100, description ≤ 500, logo ≤ 200) the frontend asserts; the comment in `InlineEditField.tsx:53-56` claims parity but the truth is server-side.

## Test plan coverage

The PR description's manual test plan covers the right flows (super-admin edit, owner/admin edit + `SpaceSwitcher` refresh, idempotent save, oversized/empty/emoji inputs, keyboard). No automated tests are added for `InlineEditField`; given how much keyboard / clamp / parser logic now lives there, a small Vitest suite would be valuable as a follow-up — not a merge blocker.

## Files reviewed

- `src/api/space.ts` — new `SpaceProfileUpdateReq` + `updateSpaceProfile` (manager PUT)
- `src/api/space-user.ts` — new `SpaceUserProfileUpdateReq` + `updateSpaceUserProfile` (owner/admin PUT)
- `src/pages/Spaces/InlineEditField.tsx` — new shared inline editor
- `src/pages/Spaces/SpaceInfoPanel.tsx` — new manager-side info card
- `src/pages/Spaces/SpaceDetailDrawer.tsx` — replaces in-Drawer `Descriptions` with `SpaceInfoPanel`
- `src/pages/Spaces/index.tsx` — wires `onUpdated` to refresh the list
- `src/pages/SpaceAdmin/SpaceAdminLayout.tsx` — collapsible "空间信息" panel for owner/admin self-service
- `src/styles/admin.css` — scoped under `.admin-shell`, follows the existing isolation contract

Jerry-Xin (approved):
Project relevance gate passed: this PR directly extends octo-admin space profile management with inline editing, and I found no blocking correctness or security issues.

**Non-Blocking**
- 🟡 Warning: `src/pages/Spaces/InlineEditField.tsx:82` relies on React state (`saving`) to prevent duplicate submits, but two rapid submit events can run before the state update is observed by the current closure. Consider a `useRef` guard if duplicate `PUT`s matter for noisy audit logs or repeated success messages.
- 🔵 Suggestion: `src/pages/Spaces/SpaceInfoPanel.tsx:52` and `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:103` patch local state from the submitted value rather than refetching or consuming a server response. That is fine for these fields, but `updated_at` remains stale in the open drawer/layout after save, and any future server normalization would not show until a refresh.

**Highlights**
- ✅ Inline edit is scoped to active spaces and preserves server-side authorization as the source of truth.
- ✅ Rune-count validation avoids UTF-16 truncation issues for emoji and other surrogate-pair input.
- ✅ Super-admin list refresh and space-admin `mySpaces` patching cover the main visible stale-state paths.

I could not verify the build locally because dependencies are not installed in this workspace: `npm run build` failed with `tsc: command not found`.

lml2468 (approved):
Incremental re-review for f2089d2468efdd3b00ce92ae0fe8a07695f3407c.

No blocking findings in the new changes.

Verified:
- `MAX_USERS_HARD_CAP` is now exported from `src/api/space.ts` and used by both the create form and inline edit path, so the create/edit max_users cap mismatch is addressed.
- The create form now enforces a non-negative integer value and the shared 50000 cap before submitting.

Remaining non-blocking notes from prior rounds are still present, but I do not consider them blockers for this PR:
- `SpaceInfoPanel.tsx:52-56` still applies successful saves over the captured `space` prop snapshot, so overlapping saves can drop another just-saved field locally.
- `InlineEditField.tsx:86-98` trims the next text value but compares it against the raw current value, so raw values with surrounding whitespace can still produce an idempotent PUT.
- `InlineEditField.tsx:66-68` still resets the draft when `value` changes while editing.
- `SpaceDetailDrawer.tsx:61-64` plus `index.tsx:423-429` still refresh only the parent list after save, not the drawer detail metadata such as `updated_at`.

yujiawei (approved):
# Code Review — PR #54 (octo-admin)

Independent review of the inline-edit space profile work wiring `PUT /v1/manager/spaces/:id` and `PUT /v1/space/:id`.

## Verdict

Approving with a couple of low-priority polish suggestions and security notes for the human reviewer.

## Summary

The change cleanly splits the two server endpoints (manager vs. owner/admin self-service), introduces a reusable `InlineEditField` with sane keyboard / validation / dedupe semantics, and reuses a single `MAX_USERS_HARD_CAP` constant across the create and edit paths. Validation parity with the server (rune count vs. UTF-16 code units) is intentional and the rationale is documented inline. CSS is correctly scoped under `.admin-shell` so it cannot leak out of the admin shell.

## Strengths

- **Endpoint separation is correct.** `updateSpaceProfile` (manager path) carries `max_users`; `updateSpaceUserProfile` (user path) deliberately omits it. Field whitelists match the description's contract.
- **Validation parity.** Skipping native `maxLength` and counting via `Array.from(s).length` matches `utf8.RuneCountInString` and avoids surrogate-pair truncation. The comments explaining *why* are exactly the kind of context that prevents future regressions.
- **Save dedupe + idempotency** — `commit()` short-circuits when `next === current` or both are empty, so a no-op edit doesn't hit the API.
- **Re-entrancy guard.** `if (saving) return` inside `commit()` plus `loading={saving}` on the confirm button protects against double-fire on rapid Enter / double-click.
- **Numeric input has both keystroke filtering and `onChange` clamping**, plus a `digitOnlyParser` for paste paths. Defense in depth here is appropriate, since `InputNumber`'s parser only fires on commit.
- **`MAX_USERS_HARD_CAP` is reused in both create and edit** — this directly closes the "creatable but not editable back" inconsistency the comment calls out.
- **mySpaces sync after a user-side save** (`name` / `logo` / `join_mode`) is the right call — it keeps `SpaceSwitcher` from going stale until the next `/space/my` poll.
- **Edit affordance is gated** on `role >= 1 && status === 1`, hiding the pencil for members and non-active spaces. This is UX gating only; the actual authorization must (and presumably does) live on the server.

## Findings

### P2 — Draft can be clobbered if `value` prop changes mid-edit

`InlineEditField.tsx`:

```ts
useEffect(() => {
  if (editing) setDraft(value ?? '')
}, [editing, value])
```

Because `value` is in the dep array, any prop change during an active edit resets `draft` to the new prop value, silently discarding what the user has typed. In the current call sites this is benign (parents only mutate `value` after a successful save, by which point `editing` is about to flip to `false`), but it's fragile — e.g. a future caller that adds a polling refresh, or an `onUpdated` callback that triggers a refetch into the same panel, would start dropping keystrokes.

Suggested fix: depend only on `[editing]` and seed the draft when entering edit mode, or capture the initial `value` into a ref on transition.

### P2 — `SpaceUserProfileUpdateReq.preset_group_ids` typing

`src/api/space-user.ts`: `preset_group_ids?: string`. The matching create-side type uses `string | null`. The user UI doesn't expose this field today, so it's a non-issue at runtime, but if/when a preset-groups page is added, an explicit `null` for "clear" should be representable. Worth aligning with the create signature pre-emptively.

### P2 — Optimistic update can briefly diverge from server-normalized state

In `SpaceInfoPanel.save` and `SpaceAdminLayout.saveDetailField`, after `await updateSpace*Profile(...)` the local state is patched with the value the client *sent*, not the value the server *returned*. If the server normalizes (e.g. trims, lowercases, defaults), the panel will show the client's version until the next refetch. The drawer flow does call `onUpdated → fetchData()` for the list, but the drawer's own `space` state isn't refetched. Low-impact since the client already trims, but if either endpoint returns the updated entity, consuming that response would eliminate the divergence window.

### Nit — Stray empty line in `SpaceAdminLayout.tsx`

A blank line was added inside the `</Content></Layout>` block (around the new closing of the JSX); cosmetic.

### Nit — `meta-inline` header duplicates fields shown in the collapsed "空间信息" panel

`detail.name`, `join_mode`, `member_count`, and `description` are visible in both the always-on header strip and the editable panel. After an inline edit, both refresh from the same `detail` state, so it's coherent — but the duplication makes the page heavier and may be a UX consideration once owners/admins have a real editing surface.

## Items I'd flag for human verification (label: needs-human-review)

1. **Server-side authorization for `PUT /v1/space/:id`.** The frontend gates with `role >= 1 && status === 1`, which is UX only. Please confirm server PR #164 enforces (a) caller is a member of the space, (b) caller has admin/owner role, and (c) targeted `space_id` matches the caller's session-scoped space.
2. **Field whitelist enforcement on the server.** The user-side request type *intentionally* excludes `max_users`. If a client crafts a request body with `max_users` against `/v1/space/:id`, the server should reject or ignore it — not silently apply the manager-tier field. Worth a quick spot-check.
3. **`MAX_USERS_HARD_CAP = 50000` parity.** The constant lives client-side. Confirm server PR #158 enforces the same upper bound (and that 0-means-unlimited semantics match) so the UI's "best-effort" check is genuinely cosmetic.
4. **`logo` URL handling.** `` accepts arbitrary URLs (including `data:` and `javascript:`). Modern browsers don't execute `javascript:` in `img.src`, but `data:` SVGs can still pose XSS risk in older or embedded contexts. Existing behavior (creation has always accepted any URL), so this PR is not a regression — but if there is no server-side scheme allowlist for `logo`, please consider adding one.
5. **Member-count vs `max_users` race.** The validate rule rejects `max_users < member_count` only at client commit time. Concurrent joins between save click and server commit could still drop the cap below the live count. Server enforcement is acknowledged as best-effort in the PR description; just flagging that the UI hint is exactly that — a hint.

## Test coverage

The PR description's test plan is appropriate; the unchecked manual items (super-admin, owner/admin, idempotent save, validation, keyboard) should be exercised before merge given the security-sensitive label and the absence of unit tests in the diff.

lml2468 (approved):
## ✅ Re-APPROVED (1ff2651f9cc2)

Trivial merge commit — only brings in PR #55's workflow permissions fix (`.github/workflows/pr-contributor-welcome.yml +2 -3`). Zero changes to feature code.

All Space inline-edit logic remains identical to the previously-approved `f2089d2468ef`. No conflicts, no regressions.

Jerry-Xin (approved):
The PR is in scope for `octo-admin`: it extends existing space management/admin UI and wires matching space profile update APIs.

💬 Non-blocking

🟡 Warning — Concurrent inline saves can temporarily revert local UI state.  
In [SpaceInfoPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-pr49qdgi/repo/src/pages/Spaces/SpaceInfoPanel.tsx:52), `save()` updates with `{ ...space, [field]: value }` from the render-time `space` snapshot. Since each `InlineEditField` has independent edit state, two fields can be edited/saved at the same time; the later response can overwrite the first field in the drawer with stale local data. The same pattern exists in [SpaceAdminLayout.tsx](/Users/mlamp/.openclaw/tmp/pr-review-pr49qdgi/repo/src/pages/SpaceAdmin/SpaceAdminLayout.tsx:98) via the captured `detail` and `mySpaces`. This does not appear to lose server data because each API request is partial, but it can leave the UI stale until refetch/navigation. Prefer functional state updates or refetching the detail after save.

🟡 Warning — Drawer detail is only partially refreshed after super-admin saves.  
[SpaceInfoPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-pr49qdgi/repo/src/pages/Spaces/SpaceInfoPanel.tsx:56) optimistically patches only the edited field, while [Spaces/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-pr49qdgi/repo/src/pages/Spaces/index.tsx:428) refreshes the list. Fields like `updated_at`, and any server-normalized values, remain stale in the open drawer. If the intended behavior is “drawer + list both refresh,” consider consuming the update response or refetching `getSpace()` after save.

🔵 Suggestion — Add focused tests around `InlineEditField`.  
The new reusable editor in [InlineEditField.tsx](/Users/mlamp/.openclaw/tmp/pr-review-pr49qdgi/repo/src/pages/Spaces/InlineEditField.tsx:82) owns important behavior: idempotent saves, trim behavior, rune-count validation handoff, Escape cancel, Enter/Ctrl+Enter commit, and number clamping. A small component test suite would reduce regression risk.

✅ Highlights

The PR keeps the new feature aligned with existing API/client patterns, scopes edit affordances to active spaces, and correctly avoids native `maxLength` where UTF-16 counting would diverge from server rune-count validation.

Verification note: I could not run `npm run build` or `npm test` locally because this workspace does not have dependencies installed; `tsc` and `vitest` were missing from `node_modules/.bin`.

PR #69, opened 2026-06-03
fix(app-bots): fetch latest detail when opening edit modal

### LINKED ISSUE #68: App Bot edit modal does not fetch latest detail via GET /v1/admin/app_bot/:bot_id
## Problem

When editing an App Bot, the edit modal (`EditModal.tsx`) pre-fills the form **directly from the list row snapshot** instead of fetching the latest detail through `GET /v1/admin/app_bot/:bot_id` (or `GET /v1/space/:spaceId/app_bot/:bot_id`).

```ts
// src/pages/AppBots/EditModal.tsx
useEffect(() => {
  if (open && bot) {
    form.setFieldsValue({
      display_name: bot.display_name,
      description: bot.description,
      welcome_msg: bot.welcome_msg,
    })
    setAvatarVersion(Date.now())
  }
}, [open, bot, form])
```

The detail endpoints `getAppBot(id)` / `getSpaceAppBot(spaceId, id)` already exist in `src/api/app-bot.ts` and are used by `DetailDrawer.tsx`, but the edit flow never calls them.

## Impact

- The form shows the data captured when the list was loaded, not the value at the moment of editing.
- If the bot was modified elsewhere (another admin / another session) after the list loaded, the edit modal shows stale values, which can silently overwrite newer data on save.

This is not a "fields are blank" bug — the three editable fields (`display_name`, `description`, `welcome_msg`) are returned by the list API, so basic echo works. The issue is freshness/consistency.

## Expected

On opening the edit modal, fetch the latest detail via `GET /v1/admin/app_bot/:bot_id` (platform) or the space-scoped equivalent, then pre-fill the form with the response. Show a loading state while fetching, consistent with `DetailDrawer`.

## Scope

- `src/pages/AppBots/EditModal.tsx`

an9xyz:
## Summary

The App Bot edit modal pre-filled its form from the **stale list-row snapshot** instead of fetching the current value, so it never called `GET /v1/admin/app_bot/:bot_id` (or the space-scoped equivalent). If the bot was modified elsewhere after the list loaded, the edit form showed outdated values and could silently overwrite newer data on save.

This change makes the edit modal fetch the latest detail on open, consistent with how `DetailDrawer` already works.

## Changes

- `EditModal` now calls `getAppBot(id)` / `getSpaceAppBot(spaceId, id)` when it opens and pre-fills the form with the response.
- Pre-fills from the row data first to avoid a blank flash, then overwrites with the freshly fetched detail.
- Shows a `Modal` loading skeleton while fetching and disables the save button until the latest detail is loaded.
- Uses a `stale` guard to avoid races when switching bots quickly; falls back to the row data with an error toast if the fetch fails.

## Test plan

- [x] `tsc --noEmit` type-check passes
- [x] `vitest run` passes (9/9)
- [ ] Manual: open the edit modal and confirm a request to `GET /v1/admin/app_bot/:bot_id` fires and the form reflects the latest values
- [ ] Manual: verify the loading skeleton shows and the save button is disabled until the detail loads

Closes #68

### DIFF
--- a/src/pages/AppBots/EditModal.tsx
+++ b/src/pages/AppBots/EditModal.tsx
@@ -3,6 +3,8 @@ import { Modal, Form, Input, Avatar, Upload, message } from 'antd'
 import { CameraOutlined, LoadingOutlined, RobotOutlined } from '@ant-design/icons'
 import { useTranslation } from 'react-i18next'
 import {
+  getAppBot,
+  getSpaceAppBot,
   updateAppBot,
   updateSpaceAppBot,
   uploadAppBotAvatar,
@@ -24,19 +26,45 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
   const { t } = useTranslation('appBots')
   const [form] = Form.useForm<AppBotUpdateReq>()
   const [loading, setLoading] = useState(false)
+  const [fetching, setFetching] = useState(false)
   const [uploading, setUploading] = useState(false)
   const [avatarVersion, setAvatarVersion] = useState(Date.now)
 
+  // On open, fetch the latest detail so the form reflects the current value
+  // rather than a (possibly stale) list-row snapshot.
   useEffect(() => {
-    if (open && bot) {
+    if (!open || !bot) return
+
+    setAvatarVersion(Date.now())
+
+    const fill = (data: Pick<AppBot, 'display_name' | 'description' | 'welcome_msg'>) => {
       form.setFieldsValue({
-        display_name: bot.display_name,
-        description: bot.description,
-        welcome_msg: bot.welcome_msg,
+        display_name: data.display_name,
+        description: data.description,
+        welcome_msg: data.welcome_msg,
       })
-      setAvatarVersion(Date.now())
     }
-  }, [open, bot, form])
+
+    // Pre-fill from the row data immediately to avoid a blank flash, then
+    // overwrite with the freshly fetched detail.
+    fill(bot)
+
+    let stale = false
+    setFetching(true)
+    const fetchBot = spaceId ? getSpaceAppBot(spaceId, bot.id) : getAppBot(bot.id)
+    fetchBot
+      .then((data) => {
+        if (!stale) fill(data)
+      })
+      .catch((err: Error) => {
+        if (!stale) message.error(err.message)
+      })
+      .finally(() => {
+        if (!stale) setFetching(false)
+      })
+
+    return () => { stale = true }
+  }, [open, bot, spaceId, form])
 
   const handleOk = async () => {
     if (!bot) return
@@ -78,6 +106,8 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
       onOk={handleOk}
       onCancel={onClose}
       confirmLoading={loading}
+      loading={fetching}
+      okButtonProps={{ disabled: fetching }}
       destroyOnClose
     >
       {/* Avatar upload */}
--- a/src/pages/AppBots/EditModal.tsx
+++ b/src/pages/AppBots/EditModal.tsx
@@ -26,7 +26,11 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
   const { t } = useTranslation('appBots')
   const [form] = Form.useForm<AppBotUpdateReq>()
   const [loading, setLoading] = useState(false)
-  const [fetching, setFetching] = useState(false)
+  // The latest fetched (or row-fallback) detail. Drives the modal title and the
+  // "loaded" gate below. Deriving the gate from `detail.id === bot.id` (rather
+  // than a flag flipped inside the effect) avoids the one-frame window where the
+  // OK button could render enabled before the fetch starts.
+  const [detail, setDetail] = useState<AppBot | null>(null)
   const [uploading, setUploading] = useState(false)
   const [avatarVersion, setAvatarVersion] = useState(Date.now)
 
@@ -36,6 +40,7 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
     if (!open || !bot) return
 
     setAvatarVersion(Date.now())
+    setDetail(null)
 
     const fill = (data: Pick<AppBot, 'display_name' | 'description' | 'welcome_msg'>) => {
       form.setFieldsValue({
@@ -50,21 +55,26 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
     fill(bot)
 
     let stale = false
-    setFetching(true)
     const fetchBot = spaceId ? getSpaceAppBot(spaceId, bot.id) : getAppBot(bot.id)
     fetchBot
       .then((data) => {
-        if (!stale) fill(data)
+        if (stale) return
+        fill(data)
+        setDetail(data)
       })
       .catch((err: Error) => {
-        if (!stale) message.error(err.message)
-      })
-      .finally(() => {
-        if (!stale) setFetching(false)
+        if (stale) return
+        message.error(err.message)
+        // Fall back to the row snapshot as the editable baseline so the user
+        // is not locked out when the detail fetch fails.
+        setDetail(bot)
       })
 
     return () => { stale = true }
   }, [open, bot, spaceId, form])
+
+  // True only once the detail for the currently-open bot has settled.
+  const detailReady = !!detail && detail.id === bot?.id
 
   const handleOk = async () => {
     if (!bot) return
@@ -101,13 +111,13 @@ export default function EditModal({ bot, spaceId, open, onClose, onSuccess, onAv
 
   return (
     <Modal
-      title={t('edit.title', { name: bot?.display_name || '' })}
+      title={t('edit.title', { name: detail?.display_name ?? bot?.display_name ?? '' })}
       open={open}
       onOk={handleOk}
       onCancel={onClose}
       confirmLoading={loading}
-      loading={fetching}
-      okButtonProps={{ disabled: fetching }}
+      loading={!detailReady}
+      okButtonProps={{ disabled: !detailReady }}
       destroyOnClose
     >
       {/* Avatar upload */}


Jerry-Xin (dismissed):
Summary: This PR is in scope for `octo-admin` and correctly updates the App Bot edit modal to fetch current detail data before normal edits.

💬 Non-blocking
- 🟡 Warning: `fetching` is set inside `useEffect`, so the modal may briefly render with the OK button enabled before the effect starts the fetch (`src/pages/AppBots/EditModal.tsx:35`, `src/pages/AppBots/EditModal.tsx:109`). This is a narrow UI race, but it slightly weakens the “disabled until latest detail is loaded” guarantee. Consider tracking a `detailLoaded` state or deriving the disabled state from `open && bot && !detailLoaded`.
- 🔵 Suggestion: There is no component test coverage for this new behavior. A focused test for “opens edit modal, calls `getAppBot`/`getSpaceAppBot`, applies fetched values, and disables OK while fetching” would protect against regressions.

✅ Highlights
- ✅ Uses the correct platform and space-scoped detail APIs (`getAppBot` / `getSpaceAppBot`).
- ✅ Keeps the existing row prefill behavior while replacing it with fresh detail data.
- ✅ Includes a stale guard to avoid applying late responses after switching/closing.

yujiawei (approved):
# Code Review — PR #69 (octo-admin)

## Summary
The PR fixes a real lost-update bug: previously the edit modal pre-filled from a (potentially stale) list-row snapshot, so saving could silently overwrite newer server-side data. The fix fetches the latest detail on open and disables Save while fetching. Verdict: the targeted gap is closed correctly. One residual lost-update window remains but is a pre-existing, server-side limitation out of scope for this frontend change.

## What the fix does (verified against `src/pages/AppBots/EditModal.tsx`)
- On open it pre-fills from the row object (`fill(bot)`, L50) to avoid a blank flash, then fetches fresh detail and overwrites the form (`fetchBot.then(... fill(data))`, L55-58).
- During the fetch, `fetching` is true → `Modal loading={fetching}` renders the antd 5 skeleton and `okButtonProps={{ disabled: fetching }}` (L109-110) disables Save. The skeleton means the form inputs aren't presented, and antd's disabled OK button also blocks the Enter-to-submit path, so the user cannot commit the stale pre-fill before fresh data lands.
- Stale-guard cleanup (`stale = true` in the effect teardown, L66) prevents a late response from a previously-opened bot clobbering the current form — matching the accepted pattern in `DetailDrawer.tsx` (L87-102).

Net effect: the user edits and saves on top of freshly fetched data, not the old list snapshot. The list→save overwrite window is closed.

## Findings

### Remaining lost-update window (P2, pre-existing / out of scope)
Fetch-on-open establishes a baseline at modal-open time (T1). The user may edit for some time and save at T2. The update calls (`updateAppBot` / `updateSpaceAppBot`, `src/api/app-bot.ts` L74-75, L103-104) send only `{ display_name, description, welcome_msg }` (`AppBotUpdateReq`, L42-46) with **no version / etag / If-Match / updated_at precondition**. If another actor mutates the bot between T1 and T2, this save still silently overwrites it.

This is the unavoidable residual that fetch-on-open cannot close — it only narrows the window from "list-load → save" to "modal-open → save". Closing it fully needs server-side optimistic locking. Note `AppBot.updated_at` already exists (L21) but is not part of the update request, so the primitive isn't wired through end to end. This is a pre-existing API-level gap, not introduced by this PR, and reasonably out of scope for a frontend-only fix. Flagging for awareness/follow-up, not as a blocker.

### No new injection / leak / PII (verified)
- Fetched fields flow into `form.setFieldsValue` (React-escaped) — no XSS sink.
- Error paths use `message.error(err.message)` (text, escaped) — no injection.
- The edit modal does not handle the bot token (token reveal/rotate lives only in `DetailDrawer`), so this change adds no secret-exposure surface. No new PII is rendered — the detail fetch returns the same fields already editable.

### No unit test for the new behavior (nit)
`tsc` + 10 vitest tests pass, but no test covers fetch-on-open / Save-disabled-while-fetching, and the manual test-plan checkboxes are unchecked. A small test asserting `okButtonProps.disabled` during the pending fetch and that `fill(data)` wins over `fill(bot)` would lock in the regression fix. Non-blocking.

## Verdict
Approve. The fix does what it claims and introduces no security/data-integrity regressions. The residual lost-update window is pre-existing and requires backend optimistic-locking support to close.

yujiawei (dismissed):
# Code Review — PR #69 (octo-admin)

## Verdict: APPROVED

This is a small, well-scoped fix that does exactly what it claims: the App Bot edit modal now fetches the latest detail on open instead of pre-filling solely from the (possibly stale) list-row snapshot. The implementation mirrors the already-accepted `DetailDrawer` fetch pattern, so it stays consistent with the codebase. No P0/P1 blockers found.

## What's correct

- **Closes the targeted gap.** On open the form pre-fills from the row to avoid a blank flash, then overwrites with `getAppBot(id)` / `getSpaceAppBot(spaceId, id)`. Save is gated by `Modal loading={fetching}` + `okButtonProps={{ disabled: fetching }}`, so the user can't commit the stale pre-fill before fresh data lands (this also blocks the antd Enter-to-submit path).
- **Race handling is sound.** The `let stale` flag with the effect-cleanup teardown correctly prevents a previous bot's late response from clobbering the form when switching bots quickly — structurally identical to `DetailDrawer`.
- **Effect deps are correct.** `[open, bot, spaceId, form]` is right; the parent (`AppBots/index.tsx`) holds `editBot` as independent state and only swaps `data` on list refresh, so the `bot` prop identity is stable for the open→edit→close cycle (no spurious refetch / mid-edit clobber).
- **No new security surface.** Fetched fields flow only into `form.setFieldsValue` (React-escaped); the edit modal never touches the bot token (reveal/rotate live only in `DetailDrawer`); error text via `message.error` is escaped. No XSS/leak/PII regression.

## Non-blocking suggestions (nits — optional, do not block merge)

1. **Error-path UX.** When the detail fetch rejects, the catch only toasts the error; the form keeps the row pre-fill and `finally` re-enables Save. The user can then save the row snapshot. This is not a regression (it matches the component's pre-PR behavior, and the update payload is a fixed 3-field whitelist — `display_name` / `description` / `welcome_msg` — so it can't clobber token/status/avatar), but you could keep Save disabled or show a "showing cached values" banner on fetch failure to fully honor the PR's intent.
2. **Stale modal title.** `t('edit.title', { name: bot?.display_name || '' })` reads the row prop, which the detail fetch never writes back into. In the exact stale-row scenario this PR targets, the title bar can show the old name while the field shows the fresh value. Purely cosmetic; the submitted content is correct. Drive the title from fetched state if you want full consistency.
3. **Test coverage.** No unit test exercises the new prefill→fetch→overwrite + stale-guard sequence. Consistent with the repo's current convention (no component-test harness; `DetailDrawer` shipped the same pattern untested), so this is a nice-to-have, not a requirement. The manual test-plan checkboxes in the PR description are also still unchecked.

## Follow-up (out of scope for this PR)

- **Residual lost-update window.** Fetch-on-open narrows the stale window from list-load→save down to open→save, but a concurrent server-side change between modal-open and save is still silently overwritten — the update sends no `version`/`If-Match` precondition (`AppBot.updated_at` exists but isn't plumbed into `AppBotUpdateReq`). This is a pre-existing API-level limitation that a frontend-only change can't close; worth a separate backend ticket if true optimistic concurrency is desired.

Nice, focused fix — approving.

Jerry-Xin (approved):
The PR is in scope for `octo-admin` and the main change correctly aligns `EditModal` with the existing detail-fetching pattern used by `DetailDrawer`.

💬 Non-blocking

🟡 Warning — [src/pages/AppBots/EditModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-zojlzdsk/repo/src/pages/AppBots/EditModal.tsx:39): `detail` is not cleared when the modal closes. If a user closes the modal and later reopens the same bot, the previous `detail` still matches `bot.id`, so `detailReady` can be `true` for the first render before the new fetch effect resets it. That briefly shows stale content and enables OK, which weakens the “disabled until latest detail is loaded” guarantee. Consider clearing `detail` in the `!open || !bot` branch.

🟡 Warning — There are no component tests covering the new fetch-on-open behavior, the failure fallback, or the stale-request guard. Given this change specifically protects against stale overwrites, a focused test around opening the modal and ensuring `getAppBot`/`getSpaceAppBot` refreshes the form would reduce regression risk.

✅ Highlights

🔵 Suggestion — The `stale` guard is a good fit for quick bot switching and matches the surrounding drawer pattern.

🔵 Suggestion — Prefilling from row data and then replacing it with fetched detail is a reasonable UX compromise, especially with the OK button gated while the detail request settles.

Note: I could not run the project checks locally because dependencies are not installed in this checkout: `vitest` and `tsc` were both missing from `node_modules/.bin`.

yujiawei (approved):
# Code Review — PR #69 (octo-admin)

## Verdict: APPROVED

This is a clean, well-reasoned bug fix. The edit modal previously pre-filled from the stale list-row snapshot and never fetched current state, so concurrent edits elsewhere could be silently overwritten on save. The fix makes the modal fetch the latest detail on open, mirroring the existing `DetailDrawer` pattern. No blocking issues found.

## What was verified

| Item | Status | Evidence |
|---|---|---|
| Fetches latest detail on open | ✅ | `EditModal.tsx:58` — `spaceId ? getSpaceAppBot(spaceId, bot.id) : getAppBot(bot.id)` |
| Space vs platform routing correct | ✅ | Matches the existing `updateSpaceAppBot`/`updateAppBot` split in `handleOk` (`EditModal.tsx:84-86`) and the API layer (`api/app-bot.ts:68,97`) |
| Race guard on rapid bot switching | ✅ | `let stale = false` + cleanup `return () => { stale = true }` (`EditModal.tsx:57,73`); all `.then`/`.catch` branches check `if (stale) return`. Identical to `DetailDrawer.tsx:87-102` |
| Save blocked until detail loads | ✅ | `loading={!detailReady}` + `okButtonProps={{ disabled: !detailReady }}` (`EditModal.tsx:119-120`) |
| Load gate derives from settled detail, not a flag | ✅ | `detailReady = !!detail && detail.id === bot?.id` (`EditModal.tsx:77`) avoids the one-frame window where OK could render enabled before the fetch starts |
| Error path does not lock out the user | ✅ | `.catch` falls back to row snapshot via `setDetail(bot)` after a toast (`EditModal.tsx:65-70`) |
| `Modal loading` prop availability | ✅ | Added in antd 5.18.0; project pins `antd ^5.24.6` (`package.json`) |
| Form reset across opens | ✅ | `destroyOnClose` (`EditModal.tsx:121`) unmounts the form on close, preventing stale field carryover |

The double-gating (`loading` skeleton **and** `okButtonProps.disabled`) is correct and not redundant: antd's `Modal loading` replaces only the body with a skeleton while the footer stays interactive, and antd `Form` preserves values set via `setFieldsValue` even when fields are unmounted — so without the disabled OK button a user could submit the row-snapshot values during the fetch. Disabling OK closes that hole.

## Findings

### P2 / Nits (non-blocking)

1. **`fetch(bot)` pre-fill vs the loading skeleton — doc nuance, not a bug.** The PR description and the inline comment (`EditModal.tsx:53-54`) state the row pre-fill exists "to avoid a blank flash." In the normal path the `loading={!detailReady}` skeleton actually covers the body during the fetch, so the pre-filled form is not visible then. The `fill(bot)` call is still genuinely needed — it seeds the editable baseline for the **error path** (`setDetail(bot)` flips `detailReady` true and the form must already hold the row values). Consider tightening the comment to reflect that the pre-fill is the error-path fallback rather than the blank-flash mitigation. Cosmetic only.

2. **`catch((err: Error) => …)` assumes the rejection is an `Error`.** `EditModal.tsx:65` casts the rejection to `Error` and reads `err.message` directly; a non-`Error` rejection would surface `undefined`. This is consistent with the rest of the file (`handleAvatarUpload`) and with `DetailDrawer.tsx:95`, so it is a codebase-wide pattern, not a regression introduced here. No action required for this PR.

## Summary

The fix is correct, minimal (single file, +47/-7), follows the established `DetailDrawer` fetch/race pattern, and properly guards the save action until fresh data is loaded while degrading gracefully on fetch failure. Manual test items in the PR (request fires, skeleton shows, OK disabled) are the right things to confirm, but the code path supporting them is sound. Approving.

PR #85, opened 2026-06-09
feat: add opanalytics dashboard data views

an9xyz:
## Summary
- add opanalytics dashboard API client wrappers and query serialization tests
- rebuild Dashboard with overview KPIs, derived charts, trend and conversation-type charts, Space/global direct-chat lazy tables, and Space channel drawer
- handle opanalytics error envelope fields from the API response

## Validation
- npm test
- npm run build
- verified dashboard overview, message composition, and day/week trend rendering against the current API

## Notes
- Space list and global direct chats are lazy-loaded by section visibility
- week trend labels use start_date/end_date rather than bucket so partial weeks display the actual range


Jerry-Xin (dismissed):
Scope check passed: this PR extends octo-admin’s existing dashboard/admin analytics functionality.

One-line summary: The implementation builds and tests pass; I found non-blocking dashboard correctness/polish issues around filter consistency and stale drawer rows.

**💬 Non-blocking**
- 🟡 Warning: The top Spaces chart ignores the selected Space filter. Overview and trend requests include `space_ids`, but `fetchChartSpaces` only sends date/ranking params, so the page can show a “spaces selected” scope while the “Top Spaces by Messages” chart remains global. See [src/pages/Dashboard/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-begnbupj/repo/src/pages/Dashboard/index.tsx:612) and the scoped overview call at [src/pages/Dashboard/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-begnbupj/repo/src/pages/Dashboard/index.tsx:560).
- 🟡 Warning: Switching the channel drawer between Spaces can temporarily show the previous Space’s channel rows under the new title while the next request is loading, because `openChannelDrawer` resets filters but does not clear `channels`/`channelsTotal`. See [src/pages/Dashboard/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-begnbupj/repo/src/pages/Dashboard/index.tsx:857) and the table using existing `channels` at [src/pages/Dashboard/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-begnbupj/repo/src/pages/Dashboard/index.tsx:1520).
- 🔵 Suggestion: Add focused component or hook-level tests for the dashboard filter behavior and drawer state reset. Current tests cover query serialization well, but not the higher-risk UI state transitions.

**✅ Highlights**
- Query serialization for repeated `space_ids` is covered by tests.
- Async request sequence guards reduce stale response races across the main data fetches.
- `npm test` passed: 23 tests.
- `npm run build` passed; Vite emitted only the existing large chunk warning.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# octo-admin PR#85 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-admin/pull/85
**Head SHA:** `8a6d7adadf17bc7a995ab9cbdcd5b3f7fd87306f`
**Scope:** +2304 / -160 across 7 files — feat: add opanalytics dashboard data views

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| API client types & serialization | ✅ | `src/api/dashboard.ts` — well-typed interfaces, `buildDashboardSearchParams` correctly handles arrays/empty values, covered by 3 unit tests |
| Error envelope backward compat | ✅ | `src/api/index.ts:34-40` — new `error.code`/`error.http_status` extraction falls back to old `msg` path |
| Race condition protection | ✅ | All 7 async fetchers use `useRef` sequence counters to discard stale responses |
| Lazy section loading | ✅ | `useLazySection` with IntersectionObserver; direct section correctly gated on `spacesLoadedOnce` |
| i18n completeness | ✅ | Both en-US and zh-CN dashboard.json have matching key sets (90 keys each) |
| Space filter → data propagation | ✅ | `selectedSpaceIds` correctly flows to overview/trend; direct chats hidden when filter active |
| Channel drawer sort | ❌ | `last_active_at` dataIndex vs `last_active` sort key mismatch — see P1 below |

## 2. Findings

### P1 — Channel drawer "Last active" sort is non-functional

**File:** `src/pages/Dashboard/index.tsx`, `channelColumns` definition + `handleChannelTableChange`

**Diff-scope:** **New** — introduced by this PR.

**Root cause (data-flow trace):**
1. The "Last active" column in `channelColumns` has `dataIndex: 'last_active_at'` and `key: 'last_active'`.
2. When user clicks sort, antd Table passes `sorter.field = 'last_active_at'` (from dataIndex) to `handleChannelTableChange`.
3. `sorterKey()` returns `'last_active_at'` (it checks `sorter.field` first).
4. `isChannelSortBy('last_active_at')` checks against `CHANNEL_SORT_VALUES = ['last_active', 'human_msg_count', 'agent_msg_count', 'total_msg', 'member_count']` — `'last_active_at'` is NOT in this list → returns `false`.
5. The sort handler silently no-ops. The column header's controlled `sortOrder` stays at its previous value. The API request never changes sort parameters.

**User impact:** Clicking the "Last active" column header in the channel drawer shows no sort change. The sort indicator may visually glitch (stays at previous direction). Data remains in the original order regardless of user action.

**Fix direction:** Either:
- (a) Change `dataIndex` to `'last_active'` and add a `render` that reads `record.last_active_at`, OR
- (b) Add `'last_active_at'` to `CHANNEL_SORT_VALUES` and map it to the API's expected `sort_by` value in `handleChannelTableChange`, OR
- (c) Set `sorterKey` fallback to use `columnKey` before `field` for this column.

### P2 — TrendLineChart `loading` prop is dead code

**File:** `src/pages/Dashboard/index.tsx`, `TrendLineChart` component

**Diff-scope:** New.

```tsx
if (maxValue <= 0) {
    return <EmptyChart title={emptyTitle} hint={loading ? emptyHint : emptyHint} />
}
```

Both branches of `loading ? emptyHint : emptyHint` produce the same value. The `loading` prop has zero effect on rendering. Likely the author intended different hint text for loading vs. empty states (e.g., show "Loading…" when `loading` is true).

### P2 — Space filter `onSearch` fires API on every keystroke

**File:** `src/pages/Dashboard/index.tsx`, Space filter `<Select>`

```tsx
onSearch={(value) => void fetchSpaceOptions(value)}
```

No debounce. Each keystroke triggers `listDashboardSpaces` API call. The sequence counter prevents stale results from landing, but this generates unnecessary network traffic and server load. A 200–300ms debounce is standard practice.

### P2 — `convTypeMessageRows` hardcodes conv_type enum `[1,2,3,4]`

**File:** `src/pages/Dashboard/index.tsx`

If the backend adds new `conv_type` values, they won't appear in the conversation-type bar chart. Acceptable for a known fixed enum, but worth a comment or deriving from the API response.

### P2 — `applySpaceSearch` / `refreshAll` not memoized

Minor inconsistency: these handlers are recreated every render while similar handlers use `useCallback`. No functional impact.

## 3. Suggestions

1. **P1 fix (required):** Align the channel `last_active_at` dataIndex with the sort key. Simplest: change `dataIndex: 'last_active_at'` to `dataIndex: 'last_active'` won't work since the API field IS `last_active_at`. Best approach: add a mapping in `handleChannelTableChange` — if `sorterKey` returns `'last_active_at'`, treat it as `'last_active'`.
2. **TrendLineChart:** Differentiate the loading hint, e.g. `hint={loading ? t('charts.loading') : emptyHint}`.
3. **Debounce space search:** Wrap `fetchSpaceOptions` call in `onSearch` with a debounce utility or a simple `setTimeout`/`clearTimeout` ref pattern.

## 4. Additional Observations

- **ETL button** has no sequence guard, but the backend returns 409 for concurrent runs — acceptable.
- **`formatTime`** assumes Unix seconds. If backend ever returns milliseconds, dates will display as year ~53000. Consider documenting this assumption.
- **Test coverage** is limited to `buildDashboardSearchParams` serialization. No component tests for the Dashboard, but this is consistent with the project's current testing approach (no existing component tests).
- **CSS** additions (372 lines) are well-scoped under `.admin-shell .dashboard-*` and don't leak globally.

## 5. Data-Flow Trace Summary

| Consumed Data | Upstream Source | Reaches Consumer? |
|---|---|---|
| `currentOverview.*` (metric cards) | `getDashboardOverview()` API → `overview` state → fallback `EMPTY_OVERVIEW` | ✅ Yes |
| `messageComposition` (donut) | `currentOverview.human_msg_count` / `agent_msg_count` | ✅ Yes |
| `memberComposition` (donut) | `currentOverview.human_member_total` / `agent_total` | ✅ Yes |
| `trendRows` (line chart) | `getDashboardTrend()` API, gated by `trendFeatureReady` | ✅ Yes |
| `topSpaceMessageRows` (bars) | `chartSpaces` from `listDashboardSpaces(sort_by=total_msg, limit=6)` | ✅ Yes |
| `convTypeMessageRows` (bars) | `currentOverview.message_composition` → Map lookup by conv_type | ✅ Yes |
| `spaces` (table) | `listDashboardSpaces()` API, lazy via IntersectionObserver | ✅ Yes |
| `directChats` (table) | `listDashboardDirectChats()` API, hidden when space filter active | ✅ Yes |
| `channels` (drawer table) | `listDashboardChannels(space_id)` API, gated on `drawerOpen` | ✅ Yes |
| `spaceOptions` (filter select) | Merged from `fetchSpaceOptions` + `mergeSpaceOptions` from list fetches | ✅ Yes |

---

[Octo-Q] verdict: REQUEST_CHANGES — P1 channel drawer sort dysfunction is user-visible and new; must fix before landing.

lml2468 (changes requested):
### REQUEST_CHANGES — Need i18n on error toasts, debounce on Space search, fix dead-code in TrendLineChart

Reviewed `src/pages/Dashboard/index.tsx` and the new `src/api/dashboard.ts` / `dashboard.test.ts` / `src/api/index.ts` changes at head `8a6d7ad`. Big picture: the page is well structured, the per-fetch sequence-guard pattern (`*Seq.current`) is consistent and correctly prevents stale-response races on rapid filter changes, error envelope unwrap in `src/api/index.ts` is clean, type discipline is tight (no `any` / `as unknown as` / `@ts-ignore` in the new code), and `ApiError` is surfaced everywhere — no silent fallback to fake data. That foundation is solid.

But these need to be fixed before merge.

---

#### P0 — Blockers

**1. Dead-code branch hides real loading state in `TrendLineChart` (line 372)**
```ts
return <EmptyChart title={emptyTitle} hint={loading ? emptyHint : emptyHint} />
```
Both ternary branches are the same — `loading` is silently ignored. The parent passes `loading={trendLoading}` precisely so the user can distinguish "still fetching" from "no data". As written, a 30-day query against a sparse Space will show "No data in range" the instant the request starts, indistinguishable from a true empty result. Fix: use a separate loading hint (or render the skeleton path while loading).

**2. Error toasts bypass i18n — 8 sites**
Every catch surfaces `message.error((error as Error).message)` (lines 548, 566, 590, 624, 648, 670, 693, 890) and `message.warning(error.message)` (line 888). `error.message` is whatever the backend or axios produced ("Network Error", "timeout of 0ms exceeded", a Chinese-only backend string from `/v1/manager/dashboard/etl/run` 409). These appear verbatim in both zh-CN and en-US UIs. Org rule "all user-visible strings via t()" is violated for the entire error surface of this page. Introduce a `dashboardError(t, error)` helper that maps `ApiError.code` (now available on `ApiError`, that's the point of the envelope change in `src/api/index.ts`) to a translation key with the raw message as a `defaultValue` fallback, then use it at all nine sites.

**3. `onSearch` on the Space multi-select fires one API call per keystroke (line 1271)**
```ts
onSearch={(value) => void fetchSpaceOptions(value)}
```
No debounce. Typing five characters fires five `GET /v1/manager/dashboard/spaces`. The sequence guard correctly ignores stale responses (good), but the server is still hit and the user pays the last keystroke's latency. On a large tenant this is a real DoS-the-manager-API path. Wrap with a 200–300ms debounce (`useMemo` + lodash `debounce`, or a small `useDebouncedCallback`), and remember to cancel on unmount.

---

#### P1 — Should fix

**4. Hardcoded `'ID'` column titles (lines 1115, 1214)** — two literal strings escape `t()`. Even "language-neutral" copy goes through `t('common:column.id')` per project convention.

**5. Clickable space name uses bare `<a onClick>` without `href` (line 996)** — not keyboard-focusable, not announced as interactive by screen readers. This is the primary drawer-open surface in the Spaces table. Use `<Typography.Link>` / `<Button type="link">`, or at minimum add `role="button"` + `tabIndex={0}` + Enter/Space `onKeyDown`. Violates the keyboard-nav-on-tables checklist.

**6. ETL trigger has no confirm step (lines 880–894, 1240–1241)** — `etlLoading` blocks a true double-click in-flight, but a second click after the API returns 202-accepted will fire another run. Manager-only operational trigger should ask "Run ETL now?" via `<Popconfirm>`. Bump to P0 if your backend cannot tolerate duplicate triggers within seconds.

**7. `fetchSpaceOptions` does not refresh when the date range changes (lines 532–559)** — `rangeParams` is in the `useCallback` deps so the function updates, but it's only invoked from `onOpenChange` / `onSearch`. The Space dropdown can keep showing "active in previous range" entries until the user reopens. Either refetch inside the RangePicker `onChange`, or add `useEffect(() => { void fetchSpaceOptions(spaceSearch) }, [rangeParams])`.

---

#### P2 — Nits

- **L246 `DonutChart` `<svg … aria-hidden>`** — chart is fully hidden from assistive tech. Consider `role="img" aria-label={title}` with a short text summary like `EmptyChart`'s SVG does on line 408.
- **L736–742 `useEffect` deps use `selectedSpaceIds.length`** rather than the array — fine today because `listDashboardDirectChats` does not accept `space_ids`, but it's a stale-deps trap if anyone later adds space filtering. Add a comment, or depend on the full array.

---

#### Verified vs sampled

- Verified by grep at head `8a6d7ad`: lines 372, 548, 566, 590, 624, 648, 670, 693, 888, 890, 996, 1115, 1214, 1241, 1271, and the absence of `any` / `as unknown as` / `@ts-ignore` in the new code.
- Verified `src/api/dashboard.test.ts` covers `space_ids` repeated-key serialization and empty-value dropping — good but minimal. No fetcher-level tests for `getDashboardOverview` / `getDashboardTrend` (e.g. axios mock asserting URL + query params) — fine for this PR scope but consider follow-up.
- Verified i18n key sets — `en-US/dashboard.json` and `zh-CN/dashboard.json` are 100% aligned (no missing keys either direction).
- Not verified: actual runtime behavior against `im-test.deepminer.com.cn` (taking the PR description's validation note at face value).

yujiawei (changes requested):
# Code Review — PR #85 (octo-admin)

**Scope:** `feat: add opanalytics dashboard data views` — rewires the admin Dashboard from a placeholder KPI page into a full operations-analytics view (overview KPIs, donut/bar/trend charts, paginated Space / Channel / Direct-chat tables, ETL trigger). 7 files, +2304/-160.

## Verification summary

- ✅ `tsc --noEmit` — clean (exit 0)
- ✅ `vitest run` — 23/23 pass, including the 3 new `buildDashboardSearchParams` cases
- ✅ i18n key parity — `en-US` and `zh-CN` `dashboard.json` have identical key sets; all `dashboard.*` and `common:*` keys used in the component (`common:action.refresh`, `common:action.search`, `common:table.total`) exist in both locales
- ✅ Error-envelope change in `src/api/index.ts` is backward-compatible (falls back to `data.msg` then `error.message`; `http_status` falls back to `response.status`); 404→trend-unavailable and 409→ETL-in-progress are handled correctly
- ✅ Race-safety: every fetcher uses a monotonic `seq`-ref guard so stale responses are dropped — solid pattern, applied consistently (one gap noted below)
- ✅ Server contract: direct-chats are correctly treated as global-only (cleared + an explanatory empty card when a Space filter is active), matching the backend's `private_active_count=0`-under-filter behavior

Overall this is well-structured, type-safe, and consistent with the existing codebase. One functional defect blocks merge; the rest are non-blocking.

## P1 — blocking

### 1. Channel table "Last active" sort is a dead control (dataIndex/key mismatch)

`src/pages/Dashboard/index.tsx:1204-1212`

```tsx
{
  title: t('channels.column.lastActive'),
  dataIndex: 'last_active_at',   // <- field name
  key: 'last_active',            // <- sort key
  sorter: true,
  sortOrder: channelSortBy === 'last_active' ? orderToAntd(channelOrder) : null,
  ...
}
```

The underlying field is genuinely `last_active_at` (`DashboardChannelItem`, `src/api/dashboard.ts`), so `dataIndex` must stay `last_active_at` for the cell to render. But `sorterKey()` (`:165-169`) prefers `sorter.field` over `sorter.columnKey`, and in antd v5 `SorterResult.field` is derived from `dataIndex`:

```tsx
function sorterKey<T extends string>(sorter, fallback) {
  if (typeof sorter.field === 'string') return sorter.field      // -> 'last_active_at'
  if (typeof sorter.columnKey === 'string') return sorter.columnKey
  return fallback
}
```

So on click, `handleChannelTableChange` (`:937-955`) gets `key = 'last_active_at'`, `isChannelSortBy('last_active_at')` is `false` (`CHANNEL_SORT_VALUES` only contains `'last_active'`, `:86-92`), the branch is skipped, and `channelSortBy/channelOrder` never update — **the channel "Last active" header silently does nothing**. The user cannot toggle asc/desc or re-apply last-active sort after sorting another column. The table still *loads* correctly because the default state is `last_active`/`desc`, so this is a broken interaction rather than a data bug — hence P1, not P0.

The Space and Direct-chat last-active columns are unaffected (their `dataIndex` and `key` both equal `last_active`); only the channel column has the field-name divergence.

**Fix (any one):** have `sorterKey()` prefer `columnKey` over `field`; or normalize `'last_active_at' → 'last_active'` before the `isChannelSortBy` check; or add a per-column sort-key mapping. This is the right fix to apply uniformly since the same `sorterKey` helper drives all three tables.

## P2 — recommended, non-blocking

### 2. Direct-chats reset doesn't bump the seq guard (latent race)

`src/pages/Dashboard/index.tsx:728-735` clears direct-chat state and returns early when a Space filter is selected, but does **not** increment `directSeq`:

```tsx
if (selectedSpaceIds.length > 0) {
  setDirectChats([]); setDirectTotal(0); return   // directSeq.current unchanged
}
```

If a `fetchDirectChats(seq=N)` is in flight at that moment, its success handler (`:666-668`) still sees `seq === directSeq.current` and re-populates the just-cleared state. No user-visible impact today (the filtered branch renders the empty card, not the Table, and deselecting refetches), but it deviates from the seq-discipline every other fetcher follows. Fix: `++directSeq.current` in the reset branch, or re-check `selectedSpaceIds` inside `fetchDirectChats`.

### 3. convType "Human" segment is derived as `total − agent`, ignoring `human_msg_count`

`src/pages/Dashboard/index.tsx:966-973` sets `value = total_msg_count`, `secondary = agent_msg_count`; `HorizontalBars` then renders the primary segment (labeled **Human messages**, `:1383`) as `(value − secondary)/value`. The backend already exposes `human_msg_count` (`DashboardMessageCompositionItem`, `src/api/dashboard.ts`) but it is unused. If `total_msg_count ≠ human_msg_count + agent_msg_count` for a conv_type (system/other messages folded into the total), the surplus is silently attributed to "Human messages." Prefer driving the human segment from `human_msg_count` directly so the label matches the source value.

### 4. Space-name cell is an `<a>` used as a button — not keyboard accessible

`src/pages/Dashboard/index.tsx:996`

```tsx
<a className="cell-primary" onClick={() => openChannelDrawer(record)}>{name || '-'}</a>
```

No `href`, `role`, `tabIndex`, or key handler — it is out of the tab order and not exposed to assistive tech, so keyboard/screen-reader users can't open the channel drawer. It's the only non-antd interactive control on the page. Use `Typography.Link` / `Button type="link"`, or add `role="button" tabIndex={0}` + an Enter/Space `onKeyDown`.

### 5. Trend chart `aria-label` always announces the empty-state title

`src/pages/Dashboard/index.tsx:408` renders `role="img" aria-label={emptyTitle}`, but this branch only runs when `maxValue > 0` (the empty case returns `<EmptyChart/>` earlier at `:371`), and `emptyTitle` is `t('charts.empty.title')` ("No chartable data"). A populated trend chart is announced as "No chartable data." Pass a descriptive name such as `t('charts.trend.title')`.

### 6. Space multi-select `onSearch` fires a request per keystroke (no debounce)

`src/pages/Dashboard/index.tsx:1271` — `onSearch={(value) => void fetchSpaceOptions(value)}` calls the spaces API on every character. The seq-guard prevents stale state but not the network traffic. Debounce (~300ms) before calling `fetchSpaceOptions`.

### 7. `.dashboard-page { min-width: 760px }` overflows below the 720px breakpoint

`src/styles/admin.css:613-615` hard-pins the root to `min-width:760px`, while the only dashboard responsive rule (`@media (max-width:720px)`, `:978-982`) merely stacks the heading row. Below 760px the whole content area scrolls horizontally before that breakpoint can help, defeating the otherwise-responsive metric grid and antd `xs=24` cols (relevant for narrow viewports / the drawer). Lower/drop the floor, or align the breakpoint to it.

## Nits

- **No-op ternary** (`:372`): `hint={loading ? emptyHint : emptyHint}` — both arms identical; `loading` is otherwise unused in `TrendLineChart`. Compare `:1341` which does the loading/empty distinction correctly. Either collapse to `hint={emptyHint}` or pass `t('charts.loading')` while loading.
- **Hardcoded `title: 'ID'`** (`:1115`, `:1214`): the only user-facing strings in the file not routed through `t()`. Low impact (language-neutral), add a key for consistency.
- **Channel `scroll={{ x: 1380 }}`** (`:1526`) is below the sum of its 11 explicit column widths (~1580). antd uses `max(sum, scroll.x)` so it still works, but the value is misleading/inconsistent (Space and Direct tables match their sums).
- **Copyable member UIDs / channel IDs** (`:1083-1085`, `:1119-1124`): admin-only and intended, no injection risk (antd escapes, no `dangerouslySetInnerHTML`). Advisory only — just confirm role-gating is sufficient for exposing direct-chat participant UIDs.

## Verdict

**CHANGES_REQUESTED** — one verified P1 functional defect (dead channel last-active sort) shipped as part of this PR's own sortable-table feature. It's a small, localized fix. The P2/nit items are quality improvements that can follow.

lml2468 (dismissed):
### APPROVE — All three P0s fixed, all P1s addressed, plus the cross-reviewer sort bug

Re-reviewed against head `bce2df87cf6f` (diff `8a6d7ad..bce2df`: `Dashboard/index.tsx` +172/-69, i18n key additions in `en/zh` `dashboard.json` and `common.json`, CSS deltas).

#### My previous P0/P1 — all fixed and verified at head

- **P0-1 dead-code in `TrendLineChart` (was L372)** → now L411 `hint={loading ? loadingHint : emptyHint}`. New `loadingHint` prop wired to `t('charts.trend.loading')` at the call site (L1473). Loading and empty are now visually distinguishable.
- **P0-2 error toasts bypassing i18n (8 sites)** → introduced `dashboardErrorKey()` (L101-117) mapping `ApiError.code` and `error.status` to 7 translation keys (`error.notFound`, `error.forbidden`, `error.requestInvalid`, `error.queryFailed`, `error.etlAlreadyRunning`, `error.etlTriggerFailed`, `error.fallback`) with the raw `error.message` as `defaultValue`. All 9 catch sites now go through `dashboardErrorMessage(error)` (L596/611/636/667/698/716/740/889/891). New error keys added in both `en-US/dashboard.json` and `zh-CN/dashboard.json` — verified 100% key alignment between locales.
- **P0-3 `onSearch` with no debounce (was L1271)** → `handleSpaceOptionSearch` (L755-763) wraps `fetchSpaceOptions` in a 300ms `setTimeout`, timer stored in `spaceOptionsSearchTimer` ref, cleared on every keystroke and on unmount (`useEffect(() => () => clearSpaceOptionsSearchTimer(), [...])` at L808). Also cleared on RangePicker change (L1353).
- **P1-4 hardcoded `'ID'`** → L1199 and L1301 now `t('common:column.id')`; added to both locale files.
- **P1-5 bare `<a onClick>`** → L1083 replaced with `<Button type="link" className="dashboard-table-link cell-primary">` — keyboard-focusable, announced as a button.
- **P1-6 ETL no confirm** → wrapped in `<Popconfirm>` (L1324-1339) with `title`/`description`/`okText`/`cancelText` all via `t()`. `disabled={etlLoading}` blocks re-open while in-flight.
- **P1-7 `fetchSpaceOptions` not refreshing on range change** → `fetchSpaceOptions` already had `rangeParams` in its deps; added `useEffect(() => { void fetchSpaceOptions() }, [fetchSpaceOptions])` (L781-783) so options auto-refresh, and RangePicker `onChange` now prunes stale options keeping only currently-selected ones (L1354).
- **P2 chart `aria-hidden`** → DonutChart now `role="img" aria-label={ariaLabel}` (L246-248); HorizontalBars + TrendLineChart take a real ariaLabel rather than reusing `emptyTitle`.

#### Cross-reviewer issues — also fixed

- **Channel "Last active" sort no-op (mochashanyao P0 + yujiawei P0)** — root cause: AntD column has `dataIndex='last_active_at'` but `key='last_active'`. Original `sorterKey` returned `sorter.field` (= dataIndex = `last_active_at`) first, which failed `isChannelSortBy` since the API enum is `last_active`. Fix at L185-189 swaps to check `columnKey` before `field` — correct AntD API. Verified all three table change handlers now route through this, and `isSpaceSortBy` / `isDirectSortBy` paths still work because their column `key` already equals their `dataIndex`.
- **Drawer state residue on Space switch (Jerry-Xin)** — `openChannelDrawer` (L926-937) now bumps `channelsSeq` and immediately resets `channels=[]`, `channelsTotal=0`, `channelsLoading=false`. Any in-flight fetch from the previous space is dropped by the seq guard, no zombie rows.
- **Direct-chats reset missing seq guard (yujiawei)** — L791-795 now bumps `directSeq.current` before clearing state when `selectedSpaceIds.length > 0`. Stale direct-chats responses cannot leak back in.
- **Top spaces chart ignored space filter (Jerry-Xin)** — `topSpaceMessageRows` (L911-924) filters by `selectedSpaceIds`; `fetchChartSpaces` bumps `page_size` to 200 when a filter is active (L671) so the post-filter pool stays meaningful.

#### Bonus quality wins

- `applySpaceSearch`, `refreshAll`, `handleRunEtl`, `openChannelDrawer` are now properly `useCallback`-wrapped with explicit deps — stale-closure risk reduced.
- `CONV_TYPE_VALUES` extracted as a `const` array with a comment explaining the API schema constraint — replaces the inline `[1,2,3,4]`.
- `HorizontalBars` now uses `primary + secondary` segment ratios instead of computing primary as `(value - secondary)/value`, which broke when API returned a `human_msg_count + agent_msg_count != total_msg_count` row.

#### Verified vs sampled

- Verified by direct grep at head `bce2df87cf6f`: every line cited above (sorterKey L185-189, dashboardErrorKey L101-117, 9 catch sites, handleSpaceOptionSearch L755-763, `<Button type="link">` L1083, `<Popconfirm>` L1324-1339, channel column `last_active` L1292-1297, openChannelDrawer L926-937, directSeq guard L791-795, topSpaceMessageRows L911-924).
- Verified `dashboard.json` and `common.json` en/zh key sets are 100% aligned and contain all referenced keys.
- Not verified at runtime: actual AntD `columnKey` behavior for tables with mixed `dataIndex`/`key` (trusting the well-documented AntD contract and the symmetric fix in `sorterKey`).

Solid round. Ship it.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-admin PR#85 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-admin/pull/85
**Head SHA:** `bce2df87cf6fe696344a70e56fa4fe3bd6f4d71b` (commit 2/2 — "fix: address dashboard review feedback")
**Scope:** +2478 / -163 across 9 files — feat: add opanalytics dashboard data views

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| API client types & serialization | ✅ | `src/api/dashboard.ts` — well-typed interfaces, `buildDashboardSearchParams` handles arrays/empty values, 3 unit tests pass |
| Error envelope backward compat | ✅ | `src/api/index.ts:34-40` — new `error.code`/`error.http_status` extraction falls back to legacy `msg` path |
| Race condition protection | ✅ | All 7 async fetchers use `useRef` sequence counters to discard stale responses |
| Lazy section loading | ✅ | `useLazySection` with IntersectionObserver; direct section correctly gated on `spacesLoadedOnce` |
| i18n completeness | ✅ | Both en-US and zh-CN dashboard.json have matching key sets (100 keys each, including new error/toast keys) |
| Space filter → data propagation | ✅ | `selectedSpaceIds` flows to overview/trend; direct chats hidden when filter active |
| Channel drawer sort key | ✅ | `key: 'last_active'` → `sorterKey()` returns `'last_active'` → matches `DashboardChannelSortBy` type |
| Commit 2 fixes from prior review | ✅ | Top Spaces chart now filters by selected spaces; drawer clears channels on space switch; Popconfirm on ETL; debounced search; structured error messages; aria-labels; bar chart segment math hardened |
| Chart data refetch on space swap | ⚠️ | `fetchChartSpaces` depends on `selectedSpaceIds.length` not values — see P2 below |
| Date range → spaces table pagination | ⚠️ | `RangePicker` onChange resets direct/channel pages but not `spacePage` — see P2 below |

## 2. Findings

### P2 — Top Spaces chart stale on space swap (same count)

**File:** `src/pages/Dashboard/index.tsx:665-680` (`fetchChartSpaces`) + `:909-923` (`topSpaceMessageRows`)

**Diff-scope:** **New** — introduced by commit 2's fix for the prior review's "Top Spaces chart ignores Space filter" finding.

**Data-flow trace:**
1. `fetchChartSpaces` dependency array: `[dashboardErrorMessage, mergeSpaceOptions, rangeParams, selectedSpaceIds.length]` — only `.length`, not the array values.
2. `topSpaceMessageRows` memo: `[chartSpaces, selectedSpaceIds]` — filters `chartSpaces` by current `selectedSpaceIds`.
3. When user swaps space A → space B (same count=1): `selectedSpaceIds.length` unchanged → `fetchChartSpaces` does NOT refetch → `chartSpaces` still holds data fetched for space A.
4. `topSpaceMessageRows` filters `chartSpaces` for space B's ID → space B not in cached data → chart shows empty or fewer bars.

**Impact:** Chart temporarily shows incorrect/empty data until user changes selection count, refreshes, or changes date range. Not a full breakage (user can recover), but produces user-visible incorrect chart rendering.

**Fix direction:** Change dependency from `selectedSpaceIds.length` to `selectedSpaceIds` (or add a serialized key like `selectedSpaceIds.join(',')`). This will refetch chart data when the actual selection changes.

---

### P2 — Spaces table pagination not reset on date range change

**File:** `src/pages/Dashboard/index.tsx:1340-1355` (`RangePicker` onChange)

**Diff-scope:** **New** — the RangePicker onChange handler resets `directPage` and `channelPage` but omits `spacePage`.

**Data-flow trace:**
1. User navigates to page 5 of spaces table.
2. User changes date range in RangePicker.
3. `onChange` resets `directPage(1)` and `channelPage(1)` — but `spacePage` stays at 5.
4. `fetchSpaces` refetches with `page_index: 5` + new date range → if new range has fewer results, page 5 is empty.
5. User sees empty spaces table until they manually navigate back.

**Impact:** Minor UX inconsistency — direct chats and channels tables reset correctly, but spaces table doesn't. User can recover by clicking page 1.

**Fix direction:** Add `setSpacePage(1)` to the RangePicker onChange handler, alongside the existing `setDirectPage(1)` and `setChannelPage(1)`.

---

### Nits (non-blocking)

1. **Debounce timer not cleaned on dep change** (`src/pages/Dashboard/index.tsx:764-771`): `handleSpaceOptionSearch` captures `fetchSpaceOptions` in a 300ms setTimeout. If `fetchSpaceOptions` changes (e.g., date range change) before the timer fires, the stale closure runs. Sequence counter prevents data corruption, but one unnecessary API call may fire. Consider clearing the timer in `fetchSpaceOptions`'s effect or using a ref-based debounce.

2. **`fetchChartSpaces` page_size=200 cap** (`src/pages/Dashboard/index.tsx:668`): When >6 spaces are selected, fetches top 200 by message volume. If a selected space ranks below top 200, it won't appear in the chart. Acceptable for a "Top Spaces" chart, but consider documenting this limit or using a space-ID-specific endpoint if available.

3. **`RangePicker` onChange clears non-selected space options** (`src/pages/Dashboard/index.tsx:1352`): `setSpaceOptions(prev.filter(...))` drops all non-selected options on date change. The `fetchSpaceOptions` effect refetches immediately, but there's a brief window where the dropdown shows only selected items. Not harmful since the refetch is fast.

## 3. Suggestions

- Add `selectedSpaceIds` (not just `.length`) to `fetchChartSpaces` deps — this is the one-line fix for P2 #1.
- Add `setSpacePage(1)` to RangePicker onChange — one-line fix for P2 #2.
- Consider adding component-level tests for filter/sort state transitions (the prior review also suggested this). Current tests cover query serialization well but miss the higher-risk UI state flows.

## 4. Extra Findings

- **Positive:** Commit 2 thoroughly addressed all findings from the prior review (Top Spaces filter, drawer stale state, error handling, ETL confirmation, debounced search, aria-labels, bar chart math, mobile responsive CSS). The diff between commits shows disciplined, targeted fixes.
- **Positive:** The `dashboardErrorKey` function provides structured error mapping with i18n fallbacks, and the `ApiError` envelope extraction in `api/index.ts` is backward-compatible with the legacy `msg` format.
- **Positive:** Sequence counter pattern consistently applied across all 7 fetchers — no stale response races.

## 5. Data-Flow Trace Summary

| Consumer | Upstream Source | Data Flows Correctly? |
|----------|----------------|----------------------|
| KPI metric cards | `getDashboardOverview` → `overview` → `currentOverview` | ✅ (EMPTY_OVERVIEW fallback for null) |
| Message composition donut | `currentOverview.human_msg_count` / `agent_msg_count` | ✅ |
| Member composition donut | `currentOverview.human_member_total` / `agent_total` | ✅ |
| Top Spaces bars | `listDashboardSpaces` → `chartSpaces` → filtered by `selectedSpaceIds` | ⚠️ (stale on swap, see P2) |
| Trend line chart | `getDashboardTrend` → `trendRows` | ✅ (404 → unavailable state) |
| Conv type bars | `currentOverview.message_composition` → mapped by `CONV_TYPE_VALUES` | ✅ (empty array fallback) |
| Spaces table | `listDashboardSpaces` → `spaces` | ✅ (lazy loaded, seq guarded) |
| Direct chats table | `listDashboardDirectChats` → `directChats` | ✅ (hidden when spaces selected, seq incremented) |
| Channels drawer | `listDashboardChannels(space_id)` → `channels` | ✅ (cleared on drawer open, seq guarded) |
| Error messages | `ApiError.code`/`status` → `dashboardErrorKey` → i18n | ✅ (fallback chain) |

---

[Octo-Q] verdict: **APPROVE** — no P0/P1 findings. Two P2 issues (chart stale on space swap, spaces pagination not reset on date change) are non-blocking correctness nits with one-line fixes. Recommend addressing before merge but not blocking.

yujiawei (changes requested):
# Code Review — PR #85 (octo-admin)

Solid, well-structured rebuild of the dashboard into real opanalytics data views. The API layer is cleanly typed, `buildDashboardSearchParams` is correctly unit-tested (including the repeated-`space_ids` contract and empty-value dropping), race conditions are guarded with per-fetch sequence refs, and the new error-envelope handling is a good addition. `npm test` passes (23/23, incl. 3 new dashboard tests) and `npm run build` (tsc + vite) is clean.

I'm requesting changes for **one** functional issue (P1). Everything else below is non-blocking polish (P2) that can land in this PR or a follow-up.

## Blocking (P1)

### 1. Trend view is gated on an unrelated optional field, so a present-but-empty overview permanently disables it
`src/pages/Dashboard/index.tsx` (`trendFeatureReady`, ~L555; trend effect ~L778–785; `fetchTrend` ~L678–701)

```ts
const trendFeatureReady = Array.isArray(overview?.message_composition)
const trendFeatureChecked = overview !== null
...
useEffect(() => {
  if (trendFeatureReady) { void fetchTrend(); return }
  setTrendRows([])
  setTrendUnavailable(trendFeatureChecked)
}, [fetchTrend, trendFeatureChecked, trendFeatureReady])
```

Whether the `/trend` endpoint is callable is inferred from the presence of `overview.message_composition` — a field of a **different** response. `message_composition` is typed optional (`DashboardOverview.message_composition?` in `src/api/dashboard.ts:118`, and it is the only optional field on that interface), so an overview object can legitimately arrive without it: an empty/quiet tenant with `omitempty`-style serialization, a partial response, or any backend that simply doesn't populate that array. In those cases `trendFeatureReady` is `false`, `fetchTrend` is never called, and the trend card shows "unavailable" even though `/trend` may be perfectly healthy.

Worse, this short-circuits the correct availability probe that already exists: `fetchTrend` handles `404 → setTrendUnavailable(true)` (and `setTrendUnavailable(false)` on success). The `message_composition` gate makes that real probe unreachable in the failure case.

**Fix:** drop the `message_composition` gate — always call `fetchTrend` in the effect and let its existing 404 handling mark the feature unavailable. If a pre-flight gate is genuinely wanted, gate on a field that actually reflects trend availability.

## Non-blocking suggestions (P2)

**Correctness-adjacent**
- **`http_status === 0` swallowed by `||`** — `src/api/index.ts:41`: `errorEnvelope?.http_status || error.response?.status`. A literal `0` http_status falls through to the axios status. Use `??` if `0` is ever a meaningful value (low risk, but `??` is the correct intent here).
- **Top Spaces chart drops selected-but-low-volume spaces** — `fetchChartSpaces` doesn't pass `space_ids` (the spaces endpoint has no such param); it fetches the global top-200 by `total_msg` and filters client-side. A selected space ranked below position 200 in the range never appears in the chart, while the KPIs/trend (which do pass `space_ids`) still reflect it. Edge case, but a silent chart-vs-KPI disagreement.
- **conv_type chart hard-codes `[1,2,3,4]`** — `convTypeMessageRows` iterates `CONV_TYPE_VALUES` and ignores any `conv_type` outside 1–4, so a future type would be silently omitted from the chart (tables already fall back to `convType.unknown`). The math itself is safe. Documented via the inline comment, so this is a known limitation; consider unioning with the actual types present.

**Lazy-loading coupling**
- The direct-chats lazy section gates on `spacesLoadedOnce` (`useLazySection(spacesLoadedOnce)`), so its IntersectionObserver only attaches after the Spaces section has fetched once. It self-heals (no leak), but the chained readiness is non-obvious — a one-line comment would prevent a future "simplify" from changing load behavior.

**Architecture / maintainability**
- The component is ~1640 lines mixing 5 pure chart primitives, 6 fetchers, 3 column defs, and 3 table-change handlers. The chart primitives (`DonutChart`/`HorizontalBars`/`TrendLineChart`/`MetricCard`/`EmptyChart`) and pure helpers (`formatNumber`/`formatPercent`/`buildTrendPath`/sort guards) are state-independent and would benefit from extraction (and become unit-testable, as `buildDashboardSearchParams` already demonstrates).
- The three `handle*TableChange` handlers are structurally identical (only setters + `isXSortBy` guard differ) — a small `makeTableChangeHandler` factory removes the copy-paste and the three `as SorterResult<unknown>` casts.
- `Dashboard` is eagerly imported at the route (`src/App.tsx`), so it + dayjs + the antd Table/Drawer/DatePicker surface land in the main chunk, feeding the existing vite >500kB warning. Route-level `React.lazy` + `Suspense` (at least for this now-heaviest page) is the highest-leverage fix. The in-page `useLazySection` only defers rendering, not the initial JS payload.
- `refreshAll` / mount effects can fan out 4–6 concurrent GETs with no `AbortController`; superseded requests still complete server-side (seq guards already prevent stale-state writes, so this is about cancellation, not correctness).

**Accessibility**
- `HorizontalBars` puts `aria-label` on a plain `<div>` with no role (~L422); a generic div doesn't expose `aria-label` to most screen readers. The two SVG charts correctly use `role="img"` — add `role="img"` (or `role="group"`/`figure`) to the bars container for consistency. Per-row values are visible text, so only the chart-level label is dropped.
- The trend line chart announces only its title + two totals; the per-bucket series lives inside `role="img"` SVG with no tabular fallback. Donut/bars are covered by their text legends; the trend chart is the gap. Optional: embed summary numbers in the aria-label or add a visually-hidden table.
- Trend lines and stacked-bar segments are distinguished by color only (legends pair color with text, so it's recoverable). Optional polish: dashed stroke / pattern on the agent series. Status and conv-type pills already render text labels — those are fine.

**Note:** i18n coverage is complete — every `t()` key used in the dashboard exists in both `en-US` and `zh-CN` (verified), and the dynamic `convType.${n}` keys all resolve with a `convType.unknown` fallback.

## Summary
Good work overall — the data plumbing, race-guarding, and tests are well done. Please address the trend-availability gate (P1); the P2 items are suggestions and not merge blockers.

lml2468 (approved):
### APPROVE — Trend coupling unblocked, http_status fallback fixed, chartSpaces deps corrected

Re-reviewed at head `52db434489be` (diff `bce2df..52db43`: `src/api/index.ts` ±1, `src/pages/Dashboard/index.tsx` +6/-13).

#### All three follow-ups addressed

**1. Trend feature coupling — fixed**
`trendFeatureReady` (`Array.isArray(overview?.message_composition)`) and `trendFeatureChecked` are deleted entirely (verified zero grep matches). The trend `useEffect` (L778-780) now unconditionally calls `void fetchTrend()`. Backend "trend endpoint not deployed" is detected by `fetchTrend`'s own 404 handler (L692-693 → `setTrendUnavailable(true)`), which is the correct ownership boundary — trend availability shouldn't depend on whether the overview endpoint happens to populate a different optional field. `refreshAll` also drops the guard (L942).

**2. `http_status === 0` no longer silently ignored — fixed**
`src/api/index.ts` L41: `errorEnvelope?.http_status || error.response?.status` → `errorEnvelope?.http_status ?? error.response?.status`. Backend can now signal "no upstream HTTP status" (custom 0) without it being shadowed by axios's response status. Correct nullish-coalescing usage.

**3. `fetchChartSpaces` deps — fixed**
L676: `selectedSpaceIds.length` → `selectedSpaceIds`. Resolves the issue both mochashanyao and yujiawei flagged: swapping one selected space for another (same `.length`) now correctly invalidates the memoized fetcher and refires the top-spaces chart load.

#### Other observations at this head

- HorizontalBars adds `role="img"` next to its existing `aria-label` (L330) — completes the chart-as-image SR semantics.
- mochashanyao's other nit ("日期范围变更未重置 spaces 分页") was already covered before this round — `RangePicker` `onChange` at L1340-1349 calls `setSpacePage(1)` / `setDirectPage(1)` / `setChannelPage(1)`. No code change needed.
- `refreshAll` retains `selectedSpaceIds.length` in its dep array (not the full array). Acceptable here because the only branching that depends on that value is the boolean `selectedSpaceIds.length === 0` — none of the closures inside change with which specific spaces are selected.

#### Verified vs sampled

- Grep at head `52db434489be`: zero residual `trendFeatureReady`/`trendFeatureChecked`; `fetchTrend` 404 path intact at L692-693; `fetchChartSpaces` deps array confirmed `[dashboardErrorMessage, mergeSpaceOptions, rangeParams, selectedSpaceIds]` at L676; `??` operator at `src/api/index.ts` L41.
- Did not re-verify the unchanged surface from the previous round.

Clean, minimal fix round. Ready to merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-admin PR#85 Review Report
**PR:** Mininglamp-OSS/octo-admin #85 — feat: add opanalytics dashboard data views
**Head SHA:** 52db434489be16ddab48bc0df5bbacb3821e76af
**Scope:** 9 files, +2471 / −163

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| API client layer (`dashboard.ts`) | ✅ | Clean types, proper `URLSearchParams` serialization, `unwrap` pattern consistent with axios |
| Error interceptor update (`api/index.ts`) | ✅ | Backward-compatible: `errorEnvelope?.message \|\| msg \|\| error.message` chain preserves existing behavior |
| `buildDashboardSearchParams` tests | ✅ | 3 cases cover array serialization, empty-value drop, and combined params |
| Dashboard component data flow | ✅ | All consumed data traced to API getters; race conditions handled via seq-ref pattern |
| Lazy section loading (`useLazySection`) | ✅ | IntersectionObserver with proper cleanup; fallback for no-IO browsers |
| i18n coverage | ✅ | Both en-US and zh-CN fully updated with matching key sets |
| CSS | ✅ | Scoped under `.admin-shell`, responsive breakpoints included |

---

## 2. Findings

### F1 — P2: `Select` uses `onOpenChange` instead of `onDropdownVisibleChange`

**File:** `src/pages/Dashboard/index.tsx`, Space filter `<Select>` (~line 1470 in diff)

```tsx
onOpenChange={(open) => {
  if (open && spaceOptions.length === 0) void fetchSpaceOptions()
}}
```

**Diff-scope:** **new** — this code is introduced by this PR.

**Issue:** Ant Design v5 `Select` component's documented prop for dropdown visibility is `onDropdownVisibleChange`, not `onOpenChange`. The `onOpenChange` prop is used by `Dropdown`, `Popover`, and `Tooltip` — but not `Select`. TypeScript won't catch this because JSX doesn't enforce excess-property checking on component props.

**Runtime impact:** The handler is silently ignored. When a user clears the space filter (which removes non-selected options via the `RangePicker.onChange` filter), then re-opens the dropdown, the refetch won't fire. The user sees only the previously-selected spaces' options instead of the full list.

**Mitigating factors:**
- Initial `fetchSpaceOptions()` fires on mount via `useEffect` → options are populated on first render
- `onSearch` (debounced) still works independently for finding spaces
- The bug only manifests after clearing + reopening the dropdown

**Suggested fix:** Rename `onOpenChange` → `onDropdownVisibleChange`.

---

### F2 — P2: `fetchChartSpaces` hard-codes `page_size: 200` when spaces are filtered

**File:** `src/pages/Dashboard/index.tsx`, `fetchChartSpaces` callback

```tsx
page_size: selectedSpaceIds.length > 0 ? 200 : 6,
```

**Diff-scope:** **new** — introduced by this PR.

**Issue:** When `selectedSpaceIds` is non-empty, the code fetches up to 200 spaces for the "Top Spaces" chart. If a deployment has >200 spaces and the user selects many, the chart data will be incomplete (missing spaces beyond the first 200 sorted by `last_active`).

**Practical risk:** Low — selecting >200 spaces is unlikely in typical usage. But the cap is undocumented and could silently produce incomplete charts in large deployments.

**Suggested fix:** Either document the cap or use a dedicated chart-scoped API endpoint that returns aggregated top-N without pagination.

---

### F3 — P2 (nit): `Popconfirm` `disabled` doesn't prevent keyboard activation

**File:** `src/pages/Dashboard/index.tsx`, ETL button

```tsx
<Popconfirm ... disabled={etlLoading}>
  <Button type="primary" icon={<SyncOutlined />} loading={etlLoading}>
```

**Diff-scope:** **new** — introduced by this PR.

**Issue:** While `disabled` on `Popconfirm` prevents the popup from appearing, the `Button` itself is not disabled during ETL loading (it shows a spinner via `loading={etlLoading}` but remains clickable). A fast double-click could trigger the Popconfirm before `etlLoading` flips to `true` (React state update is async). The `handleRunEtl` function doesn't guard against double-submission.

**Practical risk:** Low — the server-side ETL likely has its own concurrency guard (the 409 `etl_already_running` error code is handled). But adding `disabled={etlLoading}` to the `Button` itself would be cleaner.

---

## 3. Data-Flow Trace (all consumed data → upstream source)

| Consumed data | Source | Verified flow |
|---|---|---|
| `metricCards[*].value` (space_total, group_total, etc.) | `getDashboardOverview()` → `overview` state → `currentOverview` (fallback `EMPTY_OVERVIEW`) | ✅ API response fields match consumption; null-safe via `??` fallback |
| `messageComposition` (donut chart) | `currentOverview.human_msg_count` / `agent_msg_count` | ✅ Falls back to 0/0 → donut shows empty state |
| `memberComposition` (donut chart) | `currentOverview.human_member_total` / `agent_total` | ✅ Same pattern |
| `topSpaceMessageRows` (bar chart) | `listDashboardSpaces()` → `chartSpaces` state, filtered by `selectedSpaceIds` | ✅ Client-side filter + sort + slice(0,6) |
| `trendRows` (line chart) | `getDashboardTrend()` → `trendRows` state | ✅ 404 → `trendUnavailable` flag; empty list → empty state |
| `convTypeMessageRows` (bar chart) | `currentOverview.message_composition` → mapped by `conv_type` 1-4 | ✅ Missing types default to 0 values |
| `spaces` (table) | `listDashboardSpaces()` → `spaces` state | ✅ Gated by `spacesSectionReady` (IntersectionObserver) |
| `directChats` (table) | `listDashboardDirectChats()` → `directChats` state | ✅ Skipped when `selectedSpaceIds.length > 0`; seq-ref invalidates in-flight fetches on filter change |
| `channels` (drawer table) | `listDashboardChannels(spaceId)` → `channels` state | ✅ Only fetched when drawer opens; `drawerSpace` null-check gates the call |
| `spaceOptions` (Select dropdown) | `listDashboardSpaces()` → merged into `spaceOptions` via Map dedup | ✅ Multiple fetch sources (initial, search, spaces-list) merge correctly |

**Race condition handling:** All async fetchers use a `seq` ref pattern (`++seq.current`, compare on return). This correctly invalidates stale responses. ✅

---

## 4. Additional Observations

- **Error envelope handling** (`api/index.ts`): The new `error.code` field and nested `error.error` envelope parsing is backward-compatible. Existing endpoints returning `{ msg: "..." }` still work via the fallback chain.
- **`buildDashboardSearchParams`**: Correctly handles `false` and `0` values (they pass through; only `undefined`/`null`/`''` are dropped). This is correct for boolean/numeric params.
- **`formatTime`**: Uses `dayjs.unix(value)` — correctly treats `last_active` / `last_active_at` as Unix timestamps (seconds). `0` or `null` → `'-'`.
- **Test coverage**: Only `buildDashboardSearchParams` is tested (3 cases). No component tests or API wrapper tests. Acceptable for a feature PR of this nature, but consider adding at least one integration test for the error interceptor change.

---

## 5. Verdict

All three findings are P2 (no production-blocking path, no user-visible incorrect data, no regression). The most actionable is F1 (`onOpenChange` → `onDropdownVisibleChange`) which silently drops a refetch handler.

**[Octo-Q] verdict: APPROVE** — no P0/P1 findings. Three P2 items noted for follow-up.

Jerry-Xin (approved):
## Re-review Summary

All five blocking issues from the previous round have been resolved:

1. **Loading hint dead-code** — Fixed: `TrendLineChart` now correctly renders `loadingHint` vs `emptyHint` based on `loading` prop state (L405).
2. **Error toasts bypass i18n** — Fixed: `dashboardErrorMessage` now wraps errors through `t(dashboardErrorKey(error), { defaultValue: fallback })` with proper localization keys.
3. **Space search no debounce** — Fixed: `handleSpaceOptionSearch` uses `setTimeout(300ms)` with `clearSpaceOptionsSearchTimer` to debounce API calls.
4. **Channel "Last active" sort mismatch** — Fixed: `key='last_active'` now correctly matches the `channelSortBy === 'last_active'` comparison for `sortOrder`.
5. **TrendLineChart loading prop unused** — Fixed: the `loading` boolean is now consumed in the conditional `loading ? loadingHint : emptyHint`.

### 💬 Non-blocking notes

- 🟡 `fetchChartSpaces` fetches a global top-N and filters client-side by selected Space IDs. If a user selects a Space outside the top 200, the chart may show incomplete data. Consider fetching chart Spaces scoped by `space_ids` param for accuracy.
- 🟡 `encodeURIComponent(spaceId)` recommended in the URL path interpolation at `src/api/dashboard.ts:202` as a defensive measure.

Verified: `npm test` passes (23 tests), `npm run build` succeeds.

Jerry-Xin (approved):
## Re-review Summary

All five blocking issues from the previous round have been resolved:

1. **Loading hint dead-code** — Fixed: `TrendLineChart` now correctly renders `loadingHint` vs `emptyHint` based on `loading` prop state (L405).
2. **Error toasts bypass i18n** — Fixed: `dashboardErrorMessage` now wraps errors through `t(dashboardErrorKey(error), { defaultValue: fallback })` with proper localization keys.
3. **Space search no debounce** — Fixed: `handleSpaceOptionSearch` uses `setTimeout(300ms)` with `clearSpaceOptionsSearchTimer` to debounce API calls.
4. **Channel "Last active" sort mismatch** — Fixed: `key='last_active'` now correctly matches the `channelSortBy === 'last_active'` comparison for `sortOrder`.
5. **TrendLineChart loading prop unused** — Fixed: the `loading` boolean is now consumed in the conditional `loading ? loadingHint : emptyHint`.

### 💬 Non-blocking notes

- 🟡 `fetchChartSpaces` fetches a global top-N and filters client-side by selected Space IDs. If a user selects a Space outside the top 200, the chart may show incomplete data. Consider fetching chart Spaces scoped by `space_ids` param for accuracy.
- 🟡 `encodeURIComponent(spaceId)` recommended in the URL path interpolation at `src/api/dashboard.ts:202` as a defensive measure.

Verified: `npm test` passes (23 tests), `npm run build` succeeds.

yujiawei (approved):
# Code Review — PR #85 (octo-admin)

**Scope:** `feat: add opanalytics dashboard data views` — rebuilds the admin Dashboard from a placeholder KPI page into a full operations-analytics view (overview KPIs, donut/bar/trend SVG charts, paginated Space/Channel/Direct-chat tables with sorting + pagination, a Space→channel Drawer, and an ETL trigger), adds a new typed API layer `src/api/dashboard.ts` (with `buildDashboardSearchParams` unit tests), extends `ApiError` with an error-envelope `code`, and adds i18n keys + CSS. Reviewed at head `52db434`.

## Verdict: Approve

The follow-ups from the previous round are all confirmed fixed at this head, and the remaining findings are non-blocking polish.

- **`npm test`** passes (23/23, incl. 3 new dashboard tests).
- **`npm run build`** (tsc + vite) is clean.
- **i18n is complete and symmetric**: every `t()` key used in the dashboard exists in both `en-US` and `zh-CN` (`dashboard.json` 98/98, `common.json` 10/10), and the dynamic `convType.${n}` keys all resolve with a `convType.unknown` fallback.

### Previous-round follow-ups — all addressed
1. **Trend availability gate removed.** `trendFeatureReady`/`trendFeatureChecked` are gone; `fetchTrend()` is now called unconditionally and the existing `404 → setTrendUnavailable(true)` probe is the single source of truth for trend availability. Correct ownership boundary.
2. **`http_status` nullish-coalescing** — `src/api/index.ts:41` now uses `??` instead of `||`, so a literal `0` is no longer shadowed.
3. **`fetchChartSpaces` deps** — now keys on `selectedSpaceIds` (the array), so swapping one selected space for another correctly re-fires the chart fetch.

The API interceptor change is additive and safe for existing callers: the success path is untouched, the `data.msg` fallback is preserved, and 401 logout still keys off the raw transport status before reading the envelope.

## Non-blocking suggestions (P2)

These do not block merge and can land here or as a follow-up.

**Correctness / data display**
1. **Selected space outside the global top-200 is silently missing from the "Top Spaces by Messages" chart.** `fetchChartSpaces` (`src/pages/Dashboard/index.tsx:655-675`) fetches the global top-200 by `total_msg` and never passes `space_ids`; `topSpaceMessageRows` (`~L905-919`) then filters that page client-side. A selected space ranked below #200 globally never enters `chartSpaces`, so it vanishes from the chart even though it is explicitly selected (chart can show fewer bars than selected, or be empty, with no truncation hint). The KPIs/trend do pass `space_ids`, so this is a silent chart-vs-KPI disagreement. `DashboardSpaceListParams.space_ids` already exists and serializes correctly — passing it when a filter is active would guarantee the selected spaces are present.

**React / performance**
2. **Non-unique React key in the Top-Spaces bars.** `topSpaceMessageRows` rows carry no stable id; `HorizontalBars` renders `key={row.label}` (`L342`) where `label = space.name || space.space_id` (`L911`). Two spaces with the same name collide → duplicate-key warning and possible mis-reconciliation. The conv-type and donut charts are not affected (fixed distinct labels). Thread `space_id` through the row and key on it.
3. **`fetchChartSpaces` over-refetches.** Because the callback depends on the whole `selectedSpaceIds` array but only consumes it as a `page_size` toggle (200 vs 6), swapping between two non-empty selections fires an identical-payload `/spaces` request. Depending on `selectedSpaceIds.length > 0` (a boolean) would limit refetches to the empty↔non-empty boundary. (This is compatible with fix #3 above, which is about the trend/spaces fetchers that *do* send `space_ids`.)

**Accessibility**
4. **`HorizontalBars` `role="img"` hides its own data from screen readers.** `role="img"` + `aria-label` sits on the outer `<div>` (`L330`) that also contains the real text (bar labels, values, legend). `role="img"` makes the node atomic, so AT announces only the chart title and drops the actual data. `DonutChart` and `TrendLineChart` do this correctly by placing `role="img"` on the inner `<svg>` only and keeping the legend as sibling HTML. Move the role onto a visual-only element (or drop it on the wrapper).
5. **Truncated bar labels have no `title`/tooltip.** `truncateLabel` cuts space names to 18 chars (`L172`) but the rendered `<span>` (`L344`) has no `title` attr, so the full name isn't recoverable on hover. Combined with #4 it makes long names effectively unreadable. Add `title={row.label}` or a Tooltip.

**UX polish**
6. **Conv-type chart shows "not available yet" during loading.** The conv-type `HorizontalBars` only receives `charts.convType.empty` as its hint (`L1478`); while the overview request is in flight, `currentOverview` is `EMPTY_OVERVIEW` (zeros) so it renders the empty state with "Conversation type distribution is not available yet" instead of a loading state. The Top-Spaces and trend charts already switch on a loading flag — reuse `charts.loading` here with `overviewLoading`.
7. **Trend chart legend omits the "total" series.** `TrendLineChart` draws a filled total area + human line + agent line, but the legend (`~L462-465`) labels only Human and Agent. The most prominent element (the shaded total area) is unlabeled. Add a "Total" legend entry or treat the area as purely decorative.

### Noted as future-proofing only (not defects today)
- `buildDashboardSearchParams` serializes via `String(value)`, which would emit `"true"/"false"` for a boolean param — but no Params interface declares a boolean field today, and the empty-value guard correctly preserves `0`/`false`. Worth a glance only if a boolean query param is ever added.

## Summary
Well-structured rebuild: cleanly typed API layer, race-guarded fetchers (per-fetch sequence refs), unit-tested query serialization, complete bilingual i18n, and a thoughtful lazy-loading scheme. The one prior blocker is fixed and nothing remaining blocks merge. Addressing the P2 items (especially #1 chart-vs-KPI disagreement and #4 bars a11y) would tighten correctness and accessibility.

PR #88, opened 2026-06-10
feat(dashboard): hover tooltips and KPI visual polish

an9xyz:
## Summary
- Hover tooltips on every chart in the dashboard (donut segments, horizontal bars, trend line points with a vertical guide and end markers)
- KPIs re-grouped into **Scale** / **Activity** sections with a sticky anchor nav; the active-rate detail is rendered as a green status pill
- Metric cards gain a lift on hover with a brand-tinted shadow, border and icon
- Trend chart adds an absolute/share toggle; Top Space bars switch to log scale to surface long-tail spaces
- `tabular-nums` scoped to table number columns and chart tooltips so Inter keeps its proportional glyphs on display digits (KPI 209 / 568 no longer look distorted)
- Spaces table: messages column rendered as a human/bot split bar; inactive rows dimmed
- i18n: rename Agent label to Bot across dashboard strings

## Notes for reviewers
- The dashboard Agent -> Bot rename is intentionally scoped to dashboard strings. `appBots.json` connect-guide copy remains Agent terminology because it still binds via the `--agent` CLI flag; `appBots.test.ts` now guards this distinction in both locales.
- `DonutChart` gained a `compact` prop and matching `.dashboard-chart-card-compact` styles but no caller passes `compact={true}` yet; this is staged for a follow-up layout that uses smaller donuts.
- Main dashboard lists now default to 10 rows per page; drawer drilldowns keep the previous 20-row default. Users can still choose larger page sizes from the table controls.
- The Spaces table now sorts the merged Message mix column by `total_msg`; separate human/agent message sort columns were removed with the visual consolidation. Direct-chat type is shown as a pill in the Members column rather than a separate sortable column.

## Test plan
- [x] `npm test`
- [x] `npm run build`
- [ ] Hover the message-composition / member-composition donuts — tooltip shows value and share, segment widens slightly
- [ ] Hover Top Space and Conv-type horizontal bars — tooltip shows total + human/bot split
- [ ] Hover along the trend line — vertical dashed guide + markers appear; tooltip shows total, human, bot, bot share and bot/human ratio
- [ ] Toggle trend **Absolute / Share** and **Day / Week** — both render correctly
- [ ] Hover any KPI card — card lifts ~3px, border picks up brand tint, icon turns brand color; active-rate pill stays green
- [ ] Spaces table — number columns line up vertically; inactive rows are visibly dimmed; clicking a Space name opens the drilldown drawer
- [ ] Drilldown drawer — tooltips still work; number columns line up
- [ ] Switch between zh-CN and en-US — all new strings (`nav.*`, `kpi.scope.*`, `kpi.group.*`, `charts.tooltip.*`, `charts.trend.absolute/share`, `charts.topSpaces.scaleLog`, `spaces.column.messages`, `spaces.toolbar.total`) render in both locales


Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-admin`, but the trend chart has a user-triggerable runtime crash in Share mode when trend data is empty/unavailable.

🔴 Blocking

- 🔴 Critical — `TrendLineChart` crashes for empty `rows` when `displayMode === 'share'`.
  In [src/pages/Dashboard/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-lht23xhk/repo/src/pages/Dashboard/index.tsx:564), Share mode forces `maxValue` to `100`, so the empty-state guard at lines 571-573 is skipped even when `rows.length === 0`. The component then builds `totalPoints` from an empty array and dereferences `totalPoints[totalPoints.length - 1].x` at line 592. This is reachable before the trend request returns, after a 404/unavailable trend response, or any valid empty trend response by toggling Absolute/Share. Add an explicit `rows.length === 0` guard before computing paths, and likely treat all-zero share data as empty as well.

💬 Non-blocking

- 🟡 Warning — The App Bot connect guide now tells users to bind a “Bot to a Bot” while the generated command still uses `--agent`.
  The changed strings in [src/i18n/locales/en-US/appBots.json](/Users/mlamp/.openclaw/tmp/pr-review-lht23xhk/repo/src/i18n/locales/en-US/appBots.json:60) and [src/i18n/locales/en-US/appBots.json](/Users/mlamp/.openclaw/tmp/pr-review-lht23xhk/repo/src/i18n/locales/en-US/appBots.json:63) conflict with the actual command emitted by [src/pages/AppBots/connectGuide.ts](/Users/mlamp/.openclaw/tmp/pr-review-lht23xhk/repo/src/pages/AppBots/connectGuide.ts:20). If “Bot” and “Agent” are distinct product concepts, these copy changes should be reverted or clarified before release.

✅ Highlights

- The dashboard UI changes are relevant to this repository and align with the existing React/Ant Design structure.
- `npm run build` passes after installing dependencies with `npm ci`.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# octo-admin PR#88 Review Report
**PR:** feat(dashboard): hover tooltips and KPI visual polish
**Head SHA:** `551b7ecd3bd2bd13de75963161468bc3c3836325`
**Files:** 6 changed (+946 / −183)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Donut chart tooltips | ✅ | `DonutChart` wraps each SVG `<circle>` segment and legend item in `<Tooltip>` with correct `ChartTooltipContent`; `formatPercent` guards `total===0` → `'0%'` (L170-173) |
| Horizontal bar tooltips | ✅ | `HorizontalBars` wraps each bar row in `<Tooltip>`; `segmentedTooltipLines` computes shares against `segmentTotal = Math.max(row.value, primary+secondary, 1)` — no div-by-zero |
| Trend line hover/markers | ✅ | Hotspot rects correctly tile the plot width; markers use `valueForHuman`/`valueForAgent` for y-position; CSS `:hover`/`:focus-visible` toggles opacity |
| KPI Scale/Activity split | ✅ | `metricCards.slice(0,4)` → Scale grid(4-col); `.slice(4)` → Activity grid(3-col); responsive breakpoints at 1280/960/720px |
| IntersectionObserver nav | ✅ | All four section IDs (`dashboard-kpis`, `dashboard-charts`, `dashboard-spaces`, `dashboard-direct`) are on always-rendered outer divs; observer finds them at mount |
| `tabular-nums` scoping | ✅ | Applied only to `.dashboard-number-column`, `.dashboard-message-split-head strong`, `.dashboard-bar-row strong`; tooltip portal gets its own declaration |
| MessageSplitCell | ✅ | Guards `total > 0` for percent; `minWidth: 4px` when value > 0 prevents invisible slivers |
| i18n zh-CN / en-US parity | ✅ | Both locales add identical key sets (`nav.*`, `kpi.scope.*`, `kpi.group.*`, `charts.tooltip.*`, etc.) |
| **Trend share mode with empty data** | **❌ P1** | See Finding #1 below |

---

## 2. Findings

### Finding #1 — P1: TrendLineChart crashes in Share mode when trend data is empty

**Severity:** P1 (blocking)
**Diff-scope:** **New** — the `displayMode` prop and `'share'` branch are introduced by this PR. The old code always ran in absolute mode where `maxValue <= 0` catches the empty-rows case.

**Root cause (data-flow trace):**

1. `TrendLineChart` receives `rows={trendRows}` from parent state (`useState<DashboardTrendItem[]>([])` — initially empty).
2. In share mode, `maxValue` is hardcoded to `100` (L567), bypassing the `if (maxValue <= 0) return <EmptyChart>` guard (L573).
3. With `rows = []`, `totalPoints = pointsFor(...)` returns `[]` (L586).
4. `areaPath` computation accesses `totalPoints[totalPoints.length - 1].x` → `totalPoints[-1]` is `undefined` → **TypeError: Cannot read properties of undefined (reading 'x')** (L592).
5. This crash happens during render, taking down the entire Dashboard component.

**Reachable scenarios:**
- Trend API hasn't returned yet (initial render) + user toggles to Share.
- `trendUnavailable = true` (API not available in this environment) + user toggles to Share.
- Date range yields zero trend buckets + user toggles to Share.

**File:** `src/pages/Dashboard/index.tsx`, L567-592

```tsx
// L567: share mode always sets maxValue=100, skipping the empty-rows guard
const maxValue = displayMode === 'share'
    ? 100
    : Math.max(...rows.map(...), 0)

if (maxValue <= 0) { return <EmptyChart ... /> }  // never reached in share mode

// L592: crashes when totalPoints is empty
const areaPath = `${totalPath} L${totalPoints[totalPoints.length - 1].x.toFixed(1)},${bottom} ...`
```

**Fix direction:** Add `if (rows.length === 0)` early return before the path computations (after the maxValue check), or guard the `areaPath` computation with `totalPoints.length > 0`. The simplest fix:

```tsx
if (rows.length === 0) {
  return <EmptyChart title={emptyTitle} hint={loading ? loadingHint : emptyHint} />
}
```

---

### Finding #2 — P2: appBots.json "bind this Bot to a Bot" is self-contradictory

**Diff-scope:** **New** — PR changed "Agent" → "Bot" in `appBots.json` guide strings.
**Severity:** P2 (UX copy issue, not a functional bug).

The connect guide now reads *"Send the following to OpenClaw to bind this Bot to a Bot"* while still referencing the `--agent` CLI flag. If Bot (IM-side) and Agent (backend) are distinct product concepts, this creates confusion. PR author already flagged this in the PR body notes.

**Files:** `src/i18n/locales/en-US/appBots.json` L60-68, `src/i18n/locales/zh-CN/appBots.json` L60-68

---

### Finding #3 — P2: Direct chats table silently removes UID display

**Diff-scope:** **New** — PR replaces the old `<Space direction="vertical">` (name + UID) with a single-line `<div>` showing only names + conv_type pill.
**Severity:** P2 (information removal without explanation).

The old members column rendered:
```
Alice / Bob
uid_abc / uid_def
```
The new code renders only `Alice / Bob` with the conv_type pill inline. User IDs are no longer visible in the table. If UIDs are used for debugging or support lookups, this is a regression.

**File:** `src/pages/Dashboard/index.tsx`, L1391-1404

---

### Finding #4 — P2: `dashboard-filter-toolbar` CSS class has no definition

**Diff-scope:** **New** — class added to the top toolbar div.
**Severity:** P2 (dead code, no visual impact).

The JSX applies `className="toolbar dashboard-filter-toolbar"` but `admin.css` has no `.dashboard-filter-toolbar` rule. The toolbar works via the base `.toolbar` class, so this is harmless but suggests either a missing style rule or a leftover from development.

**File:** `src/pages/Dashboard/index.tsx` L1581; `src/styles/admin.css` (absent)

---

### Finding #5 — P2/nit: Indentation error in `handleDirectTableChange`

**Diff-scope:** **New** — the line was re-indented incorrectly during the rename.
**Severity:** Nit.

```tsx
    setDirectPage(pagination.current || 1)
-    setDirectPageSize(pagination.pageSize || PAGE_SIZE)
+      setDirectPageSize(pagination.pageSize || MAIN_LIST_PAGE_SIZE)
```

Extra 2-space indent on `setDirectPageSize`.

**File:** `src/pages/Dashboard/index.tsx`, L1239

---

### Finding #6 — P2/nit: `row.primary` passed to HorizontalBars but never consumed

**Diff-scope:** **New** — PR adds `primary: space.human_msg_count` to `topSpaceMessageRows` and `primary: item?.human_msg_count || 0` to `convTypeMessageRows`.
**Severity:** Nit (dead data, no functional impact).

`HorizontalBars` always recomputes `primary = Math.max(row.value - secondary, 0)` internally (L462), ignoring `row.primary`. The passed value happens to match the recomputed value (since `value = human + agent`), so there's no inconsistency — just unused data.

**File:** `src/pages/Dashboard/index.tsx`, L1149, L1291

---

### Finding #7 — P2/nit: `last_active` columns get `dashboard-number-column` class

**Diff-scope:** **New** — PR adds `className: 'dashboard-number-column'` to timestamp columns.
**Severity:** Nit.

`dashboard-number-column` applies `font-variant-numeric: tabular-nums` and `text-align: right !important`. Timestamps like "2026-06-10 14:30" are not pure numbers; tabular-nums is harmless on digit-heavy strings but semantically imprecise.

**File:** `src/pages/Dashboard/index.tsx`, L1382, L1421

---

## 3. Suggestions

1. **Fix #1 (P1):** Add `if (rows.length === 0)` early return in `TrendLineChart` after the `maxValue` check. This is a one-line fix.
2. **Fix #2:** Either revert the `appBots.json` Agent→Bot changes (as the PR body suggests) or clarify the product terminology. If "Bot" and "Agent" are distinct, the guide should say "bind this Bot to an Agent" (original) or update the `--agent` flag reference too.
3. **Fix #3:** Consider keeping the UID line in the direct chats table, or document the intentional removal in the PR body.
4. **Fix #4:** Either add the missing `.dashboard-filter-toolbar` CSS rule or remove the class.

---

## 4. Additional Observations

- **`compact` prop on DonutChart:** No caller passes `compact={true}`. PR body acknowledges this is staged for a follow-up. The CSS rules are in place and look correct. No action needed.
- **`color-mix()` browser support:** The CSS uses `color-mix(in srgb, ...)` extensively for brand-tinted shadows and backgrounds. This is supported in all modern browsers (Chrome 111+, Firefox 113+, Safari 16.2+) but not in IE/legacy Edge. Assuming admin panel targets modern browsers, this is fine.
- **Trend legend in share mode:** The legend still shows absolute totals (`humanTotal`, `agentTotal`) even when the chart displays percentages. This is arguably correct (legend = summary, chart = per-bucket view) but could confuse users. Consider adding a "%" indicator or switching legend to averages in share mode.

---

## 5. Data-Flow Trace Summary

| Consumed data | Upstream source | Flows correctly? |
|---|---|---|
| KPI card values | `currentOverview.*` from API → `metricCards` useMemo | ✅ |
| Donut chart items | `messageComposition` / `memberComposition` useMemo → `items` prop | ✅ |
| Top Space bar rows | `chartSpaces` → `topSpaceMessageRows` useMemo (L1143-1155) → `HorizontalBars.rows` | ✅ |
| Conv-type bar rows | `currentOverview.message_composition` → `convTypeMessageRows` useMemo (L1281-1296) → `HorizontalBars.rows` | ✅ |
| Trend line rows | `trendRows` state → `TrendLineChart.rows` | ⚠️ Empty array + share mode = crash (Finding #1) |
| Spaces table rows | `spaces` state → `Table.dataSource` | ✅ |
| MessageSplitCell human/agent | `record.human_msg_count` / `record.agent_msg_count` from table row | ✅ |
| IntersectionObserver active section | `document.getElementById(DASHBOARD_SECTION_IDS)` → always-rendered divs | ✅ |
| `formatPercent` divisor | Various `total` computations | ✅ All guarded (`if (!total) return '0%'`) |

---

[Octo-Q] verdict: REQUEST_CHANGES — Finding #1 is a P1 crash in the new share mode when trend data is empty (reachable via initial render, unavailable API, or empty date range). One-line fix: add `if (rows.length === 0)` early return in `TrendLineChart`. Remaining findings are P2/nit.

lml2468 (approved):
**APPROVE** — solid, well-scoped dashboard polish. `tsc --noEmit` clean, i18n keys 100% aligned (en/zh, both `dashboard.json` and `appBots.json`), no regressions. Verified every claim against head `551b7ecd3bd2`.

### Should-fix (before/with merge) — you flagged this yourself
- `src/i18n/locales/{en-US,zh-CN}/appBots.json`: the Agent→Bot rename made the connect-guide copy self-contradictory and now-incorrect:
  - `detail.guide.intro` / `create.guide.intro`: *"bind this Bot to a Bot"* — reads as a no-op.
  - `detail.guide.footer` / `create.guide.footer`: *"To bind to a different Bot, change the `--agent` parameter"* — still names the `--agent` CLI flag while the surrounding text dropped "Agent".
  - `detail.guide.body.line2`: *"the current Session's Bot"*.
  
  This is shipped, user-visible CLI guidance — Bot (IM-side) and Agent (backend binding target) are distinct concepts, so this rename is wrong here. Recommend reverting the 8 `appBots.json` lines per locale and keeping the rename scoped to `dashboard.json` (where it's correct). It's isolated and doesn't touch functionality, so I'm not blocking on it — but please address rather than merge as-is.

### Nits (non-blocking)
- `src/pages/Dashboard/index.tsx:1242`: `setDirectPageSize(...)` is over-indented (stray 2 spaces) inside `handleDirectTableChange`. Cosmetic; lint/format will catch it.
- `DonutChart` `compact` prop + `.dashboard-chart-card-compact` / `.dashboard-chart-body.compact` styles have no caller yet (you noted this is staged). Dead until the follow-up lands — fine to keep, just flagging so it doesn't get forgotten.

### Verified correct (the risky bits)
- **Trend share mode**: `valueForHuman`/`valueForAgent` guard `total > 0` before dividing; `maxValue` pins to 100 in share mode; axis renders `%`; the area path is gated to `absolute` only (`L1628`), and the total line only ever feeds the area — so share mode correctly shows just the two normalized lines. Tooltips keep raw absolute counts in both modes — good call.
- **HorizontalBars log scale**: `Math.log1p(value)/Math.log1p(max)` is monotonic and safe at value=0 (`log1p(0)=0`); `Math.max(4, …)` keeps tiny long-tail bars visible. Segment split math unchanged from the reviewed-in-#85 version.
- **MessageSplitCell / merged "Message mix" column**: `key='total_msg'` matches `spaceSortBy === 'total_msg'` and `isSpaceSortBy` includes `'total_msg'` (in `SPACE_SORT_VALUES`), so sort still works after collapsing the human/agent/total columns into one. `total>0` guards the percentage widths.
- **IntersectionObserver anchor nav**: the four section `id`s are on always-rendered wrappers (the lazy gate is inside), so `getElementById` resolves at mount; observer is cleaned up on unmount. `rootMargin`/`threshold` look reasonable.
- **KPI regroup**: `slice(0,4)` Scale / `slice(4)` Activity = all 7 cards, no drop; `metricMeta`/`privateMetricMeta` memo deps updated correctly (no stale `selectedSpaceIds.length`/`rangeLabel` leftover).
- **Page-size constants**: `PAGE_SIZE` fully replaced (no stray refs); main lists default to 10, drawer to 20, with matching `pageSizeOptions`.
- `is_active` confirmed on `DashboardSpaceItem` (`src/api/dashboard.ts:112`) — `rowClassName` dimming is type-safe. All new icons/components (`Tooltip`, `ReloadOutlined`, `Segmented`, `ReactNode`) are imported.

Note: `npm test` shows 9 failures in `src/store/feature.test.ts` (`localStorage` undefined) — these reproduce identically on `main` and are unrelated to this PR (a local jsdom env issue; PR touches no test/store files).

lml2468 (changes requested):
**REQUEST CHANGES** — correcting my earlier APPROVE. @Jerry-Xin's 🔴 is correct and I was wrong: there is a real, reproducible runtime crash in `TrendLineChart` Share mode on empty data. Apologies for the miss.

### 🔴 Blocking — TrendLineChart crashes in Share mode with empty rows
`src/pages/Dashboard/index.tsx:565` — in `share` mode `maxValue` is hardcoded to `100`:
```ts
const maxValue = displayMode === 'share' ? 100 : Math.max(...)
```
This **bypasses the empty-data guard** at L571 (`if (maxValue <= 0) return <EmptyChart/>`), which only ever fires in `absolute` mode. With `rows = []`, execution continues to L592:
```ts
const areaPath = `${totalPath} L${totalPoints[totalPoints.length - 1].x.toFixed(1)}, ...`
```
`totalPoints` is `[]`, so `totalPoints[-1]` is `undefined` → `TypeError: Cannot read properties of undefined (reading 'x')`. Reproduced via a standalone runtime trace: `absolute` + `[]` → guard hits (EmptyChart); `share` + `[]` → **crash**.

**Reachability:** the Absolute/Share `Segmented` lives in the card `extra` and is always rendered, and `trendRows` is set to `[]` whenever the trend endpoint returns no data (L912) or an empty list (L908). So: toggle to Share while the trend chart is empty (empty date range, or a tenant/range with no messages) → the dashboard throws. In Absolute mode the identical empty state renders `EmptyChart` cleanly — which is why this is easy to miss.

**Fix** — make the guard independent of `displayMode`, e.g.:
```ts
if (rows.length === 0 || maxValue <= 0) {
  return <EmptyChart title={emptyTitle} hint={loading ? loadingHint : emptyHint} />
}
```
(or compute `maxValue` from the data in both modes and let `<= 0` catch it). A quick render test for `displayMode='share'` + `rows={[]}` would lock this in.

### 🟡 Should-fix — appBots.json connect-guide copy (unchanged from my prior note, also raised by Jerry-Xin)
The Agent→Bot rename leaked into `src/i18n/locales/{en-US,zh-CN}/appBots.json` and made the shipped CLI guidance self-contradictory: *"bind this Bot to a Bot"* and *"To bind to a different Bot, change the `--agent` parameter"*. Bot (IM-side) and Agent (backend) are distinct concepts — recommend reverting the 8 lines/locale and keeping the rename scoped to `dashboard.json`.

### Nit
- `index.tsx:1242`: stray over-indentation on `setDirectPageSize(...)`.

### Still verified-correct (unaffected by the above)
- Share-mode **value** math (`valueForHuman`/`valueForAgent`) guards `total > 0` and axis renders `%` — the bug is purely the missing **empty-rows** guard, not the per-row division.
- Log-scale bars (`log1p`, safe at 0), MessageSplitCell sort key (`total_msg` in `SPACE_SORT_VALUES`), IntersectionObserver anchors (ids on always-rendered wrappers), KPI regroup (all 7 cards), `PAGE_SIZE` fully replaced, `is_active` typed — all confirmed.
- `tsc --noEmit` clean; i18n keys 100% aligned en/zh. (TS won't catch this crash — it's a runtime empty-array access.)
- The 9 `feature.test.ts` failures are pre-existing on `main` (local jsdom `localStorage`), unrelated to this PR.

Once the empty-Share guard lands I'm happy to re-approve.

yujiawei (changes requested):
# Code Review — PR #88 (octo-admin)

Reviewed at head `551b7ecd3bd2bd13de75963161468bc3c3836325`. Scope: dashboard hover tooltips, KPI section regrouping + sticky anchor nav, trend absolute/share toggle, Top Spaces log scale, tabular-nums scoping, Spaces message-mix cell, and an Agent→Bot i18n rename.

Overall this is a high-quality, carefully-scoped UI polish PR. Accessibility (`tabIndex`/`aria-label` on SVG segments, bars, trend points), portal-aware `tabular-nums` scoping, and the share-mode axis/area handling are all done thoughtfully. One change must be addressed before merge.

## Required change (blocking)

### 1. `appBots.json` Agent→Bot rename ships self-contradictory copy — revert it

The blanket Agent→Bot replacement was correct for the dashboard, but it also rewrote the App-Bot **connect guide** strings, where "Bot" (the IM-side entity) and "Agent" (the backend binding target) are distinct concepts. The result is copy that contradicts itself:

- `src/i18n/locales/en-US/appBots.json:60` / `:79` — *"Send the following to OpenClaw to bind this Bot to a Bot:"*
- `src/i18n/locales/en-US/appBots.json:65` — *"Bind Octo bot {{name}} to a Bot."*
- `src/i18n/locales/en-US/appBots.json:66` — *"Binds to the current Session's Bot by default; check the Bot ID via /status."*
- `src/i18n/locales/zh-CN/appBots.json:60` / `:65` / `:66` — same in Chinese ("绑定到 Bot" / "将 Octo bot {{name}} 绑定到 Bot")

It also leaves a now-mismatched reference to the CLI flag. The footer (`...:63`/`:82`) still reads *"change the --agent parameter"*, and the actual command is still built with `--agent`:

```ts
// src/pages/AppBots/connectGuide.ts:19-20
const agent = t('appBots:detail.guide.agentPlaceholder')   // now "bot-id"
return `npx ... bind ... --agent <${agent}>`
```

So the surrounding prose says "bind a Bot to a Bot" while the emitted command and footer still speak of `--agent`. The PR description already calls this out as likely-wrong. Please revert the `appBots.json` changes (both locales, `detail.guide.*` and `create.guide.*`, including `agentPlaceholder`) and keep the rename confined to the dashboard strings, which is the intended product change.

## Non-blocking (nits / follow-ups)

- **Indentation glitch** — `src/pages/Dashboard/index.tsx:1242`: `setDirectPageSize(...)` is over-indented by two spaces relative to the surrounding block. Cosmetic only.
- **Staged dead code** — `DonutChart`'s `compact` prop and the `.dashboard-chart-card-compact` / `.dashboard-chart-body.compact` CSS have no caller yet (author-acknowledged). Fine to keep if a follow-up lands soon; otherwise consider deferring to the PR that actually uses it.
- **Orphaned sort enum entries** — the Spaces table replaced the separate `human_msg_count` / `agent_msg_count` columns with a single `MessageSplitCell` (`total_msg`), but `SPACE_SORT_VALUES` (`index.tsx:81-88`) still lists `human_msg_count` / `agent_msg_count`. No column surfaces those sorts anymore, so they're now unreachable via UI. Harmless, but worth trimming for clarity.

- Share mode: `maxValue` pinned to 100, axis renders `%`, area path suppressed, markers use `valueForHuman/Agent` consistently — correct.
- Top Spaces log scale: bar width uses `log1p`, while the human/agent split inside each bar stays proportional to `segmentTotal` (linear) — the intended behavior.
- IntersectionObserver: all four `#dashboard-*` anchors are rendered unconditionally (the lazy gating is only on inner table content), so the `[]`-dep observer wires up correctly on mount.
- `common:action.refresh` key exists in `common.json` — the new refresh button resolves.
- No dangling references to the removed `highlight` prop or the old `PAGE_SIZE` constant.

Verdict: CHANGES_REQUESTED — addressing item 1 (revert the `appBots.json` rename) is sufficient to clear this; the rest are optional.

yujiawei (changes requested):
## Follow-up — adding a confirmed P1 (supersedes the previous blocker priority)

After a second pass I'm escalating an additional blocker that my first review missed. This is now the top reason for the requested changes; the `appBots.json` copy issue from my earlier review still stands as a secondary blocker.

### P1 — `TrendLineChart` throws a render-time TypeError in Share mode when trend data is empty

`src/pages/Dashboard/index.tsx:564-592`

In Share mode `maxValue` is hardcoded to `100`, so the empty-data guard never fires, but `areaPath` is still computed eagerly:

```ts
const maxValue = displayMode === 'share'
  ? 100                                  // L564-565: skips the empty guard below
  : Math.max(...rows.map(...), 0)

if (maxValue <= 0) { return <EmptyChart ... /> }   // L571: never reached in share mode

// L592: totalPoints is [] when rows is [] → totalPoints[-1] is undefined → throws
const areaPath = `${totalPath} L${totalPoints[totalPoints.length - 1].x.toFixed(1)},${bottom} L${totalPoints[0].x.toFixed(1)},${bottom} Z`
```

`areaPath` is built unconditionally even though it's only rendered when `displayMode === 'absolute'` (L628), so the crash happens before that branch is reached.

**Reachable** — the Absolute/Share `Segmented` toggle (L1718-1726) renders unconditionally in the chart card, and the charts section is not behind a lazy gate. With `trendRows` empty, toggling to Share crashes the whole Dashboard. Three concrete paths to empty rows:
- Initial render before the trend API resolves (`trendRows` starts `[]`).
- `trendUnavailable` environments — the catch sets `setTrendRows([])` (L912) and `setTrendUnavailable(true)` (L914).
- A date range that yields zero trend buckets.

**Fix:** add an early return before the path math (after the `maxValue` check):

```ts
if (rows.length === 0) {
  return <EmptyChart title={emptyTitle} hint={loading ? loadingHint : emptyHint} />
}
```

(or guard the `areaPath` construction with `totalPoints.length > 0`).

### Recap of blockers
1. **P1 (new, above):** Share-mode crash on empty trend data.
2. **P1/copy:** `appBots.json` Agent→Bot rename ships self-contradictory connect-guide copy ("bind this Bot to a Bot") while `connectGuide.ts` still emits `--agent` — revert the `appBots.json` changes (detailed in my earlier review).

Verdict unchanged: changes requested.

lml2468 (dismissed):
**APPROVE** — re-review at `387b6729bf94`. The blocking 🔴 is fixed and covered by a regression test; the appBots copy is fully reverted; indentation fixed. Verified each item against the new SHA (incremental diff vs `551b7ecd3bd2`).

### 🔴 → fixed & tested: TrendLineChart Share-mode empty-data crash
- `index.tsx:570` adds a data-keyed guard `hasTrendData = rows.some(... > 0)` and the early return is now `if (!hasTrendData || maxValue <= 0)` — independent of `displayMode`, so Share mode no longer skips it.
- `index.tsx:595` also belt-and-suspenders the `areaPath`: only built in `absolute` mode, `''` otherwise — so even if execution reached it, no `totalPoints[-1].x` access.
- Reproduced the fix with a runtime trace across `{[]+share, []+absolute, all-zero-rows+share, real-data+share/absolute}` — empty/zero cases hit the guard, real data renders. The new `hasTrendData` check also correctly catches the all-zero-rows case the old `maxValue=100` would have let through.
- **Regression test added** (`src/pages/Dashboard/index.test.tsx:172`): renders `<Dashboard/>`, clicks the Share toggle with empty trend data, asserts `charts.empty.title` renders. This would have thrown on the old code. Passes.

### 🟡 → fixed: appBots.json connect-guide copy
Both locales reverted to exactly the `main` blob (verified `git diff main 387b6729bf94 -- appBots.json` is empty). The Agent→Bot rename is now correctly scoped to `dashboard.json` only. An `appBots.test.ts` guard was added too.

### 🟡 → fixed: indentation
`index.tsx:1247` `setDirectPageSize(...)` now correctly indented.

### Bonus improvements in this commit (not requested, all sound)
- Direct-chat members cell restored the uid sub-line (`.dashboard-direct-members-uid mono`) that the prior commit had dropped — good, that was a small info regression I hadn't flagged.
- Removed `dashboard-number-column` from the two `last_active` time columns (they're not numeric — correct; tabular-nums on timestamps wasn't meaningful).

### Remaining nit (non-blocking, optional follow-up)
- `SPACE_SORT_VALUES` (`index.tsx:81`) still lists `'human_msg_count'` / `'agent_msg_count'`, but after the message columns were merged into the single sortable "Message mix" (`total_msg`) column, no Spaces column can trigger those two. They're harmless (valid API fields, just unreachable from the UI). yujiawei raised this — worth dropping in a later cleanup, not worth blocking on.

### Verification
- `tsc --noEmit` clean.
- New tests pass (`index.test.tsx`, `appBots.test.ts`). The 9 `feature.test.ts` failures remain pre-existing on `main` (local jsdom `localStorage`), unrelated.
- i18n keys still 100% aligned en/zh.

Good turnaround — the fix is minimal, correct, and now guarded by a test. LGTM.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-admin` and the dashboard changes are generally well-scoped; I found no merge-blocking issues.

💬 Non-blocking

🟡 Warning: `src/pages/Dashboard/index.tsx:1034` creates `IntersectionObserver` without the fallback used by `useLazySection` at `src/pages/Dashboard/index.tsx:140`. In environments without `IntersectionObserver`, the new sticky-nav effect can throw after mount. Consider guarding with `if (!('IntersectionObserver' in window)) return`.

🟡 Warning: `src/pages/Dashboard/index.tsx:885-891` still fetches top spaces globally and only filters selected spaces client-side. A selected long-tail space outside the first 200 global results can disappear from the Top Spaces chart. This appears pre-existing, but it is more noticeable with this PR’s “surface long-tail spaces” behavior.

✅ Highlights

🔵 Suggestion: The new `appBots.test.ts` is a useful guard against the Agent/Bot terminology regression called out in the PR notes.

Verification: `npm test` passed, and `npm run build` passed.

yujiawei (dismissed):
# Code Review — PR #88 (octo-admin)

**Scope:** 6 files (`appBots.test.ts`, `en-US/dashboard.json`, `zh-CN/dashboard.json`, `Dashboard/index.test.tsx`, `Dashboard/index.tsx`, `styles/admin.css`). Reviewed at head `387b672`.

**Verdict: APPROVED.** No correctness, security, or data-loss issues. Type check (`tsc --noEmit`), the full test suite (`vitest run`, 25/25), all pass on the head commit. The findings below are non-blocking cleanup / nits.

## What this PR does
Hover tooltips across all dashboard charts (donut, horizontal bars, trend line), a trend Absolute/Share toggle, KPI cards split into Scale/Activity groups with a sticky anchor nav, a combined human/agent "Message mix" cell for the Spaces table, per-column tabular-nums alignment, inactive-row dimming, and a label rename of "Agent → Bot" throughout the dashboard i18n.

## 1. Verification

- ✅ **Build / types:** `tsc --noEmit` clean.
- ✅ **Tests:** 25/25 pass, including the two new suites.
- ✅ **Cross-file i18n guard (the important one).** `src/i18n/locales/appBots.test.ts` is a new test that asserts specific keys/values in `appBots.json` — a file **not** included in this diff. I verified every assertion against the actual `en-US/appBots.json` and `zh-CN/appBots.json` on disk: all 16 keys (`detail.guide.intro/footer/body.line1-3/agentPlaceholder`, `create.guide.intro/footer`, both locales) exist and match. The test's intent is sound: the dashboard renames its labels "Agent → Bot", and this guard deliberately pins the OpenClaw connect-guide copy to keep saying "Agent" (the bind target there really is an Agent, not a Bot). Good defensive coverage of a rename that could otherwise leak across surfaces. (`src/i18n/locales/appBots.test.ts:1`)
- ✅ **Empty-state regression fix.** In Share mode `maxValue` is a constant `100`, so the old `if (maxValue <= 0)` empty guard could never fire and an all-zero dataset would render an empty/garbage chart. The new `hasTrendData` check (`index.tsx:570`) closes this; `Dashboard/index.test.tsx` exercises exactly the "switch to Share mode with no data" path. (`src/pages/Dashboard/index.tsx:563-575`)
- ✅ **Share-mode math.** `valueForHuman`/`valueForAgent` guard `total > 0` before dividing; `maxValue=100` and `yFor` scale consistently; the total area path is correctly suppressed in Share mode (a constant 100% band would be meaningless). (`src/pages/Dashboard/index.tsx:555-596`)
- ✅ **i18n parity.** en-US and zh-CN gained the same new dashboard keys (nav.*, kpi.scope.*, kpi.group.*, charts.tooltip.*, charts.trend.absolute/share, etc.). No missing translations.
- ✅ **Page-size constants.** `PAGE_SIZE` cleanly replaced by `MAIN_LIST_PAGE_SIZE` (10) / `DRAWER_PAGE_SIZE` (20) and matching option arrays; all call sites updated, no stale `PAGE_SIZE` / `highlight` references remain.
- ✅ **tabular-nums scoping** is well-reasoned and documented — applied only to alignment-sensitive columns (not large KPI display digits), with explicit re-declarations for the body-level tooltip and drawer portals that fall outside `.dashboard-page`. (`src/styles/admin.css:1585`, `:1883`, `:1896`)

## 2. Issues

**P2 — Dead `compact` code path on `DonutChart`.** The `compact` prop and its branch (size 152 / radius 54 / strokeWidth 18, plus the `dashboard-chart-body.compact` className) are defined, but neither `DonutChart` call site passes `compact` (`index.tsx:1677`, `:1690`), so it is always `false`. Correspondingly the `.dashboard-chart-card-compact` CSS block (5 rules, `admin.css:839-892`) and the `.dashboard-chart-body.compact` rules are never matched by any rendered element. No runtime impact, but it's unused surface area. Either wire it up to a real compact layout or drop the prop + CSS to keep the component honest.

## 3. Suggestions (nits, non-blocking)

- The `IntersectionObserver` anchor-nav effect runs with `[]` deps (`index.tsx:~999`). It works because all four section ids live on always-rendered wrappers (the lazy content is nested inside). Worth a one-line comment noting that invariant, since moving an `id` onto a lazily-rendered child later would silently break active-section highlighting.
- `DonutChart` tooltip uses `key={item.label}`; fine for the fixed human/agent items, but `HorizontalBars` already uses the safer `${row.label}-${index}` for user-derived labels — consider the same pattern in the donut for consistency if items ever become dynamic.

## 4. Additional observations
- No new network calls, no auth/permission surface touched, no `dangerouslySetInnerHTML`, no secrets. Purely presentational + i18n.
- Accessibility is a net positive: chart segments/rows/points get `tabIndex={0}` and descriptive `aria-label`s, and the anchor nav has `aria-label`.

Nice, well-tested visual-polish PR. The cross-file test↔JSON pairing and the Share-mode empty-state fix are the two things that could have bitten and both are handled correctly.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-admin PR#88 Review Report
**PR:** Mininglamp-OSS/octo-admin #88 — feat(dashboard): hover tooltips and KPI visual polish
**Head SHA:** 387b6729bf94b6e26ba40a6af22aa1397d407489
**Scope:** 6 files, +1167 / −169

---

## 1. Verification Summary

| Area | Status | Evidence |
|---|---|---|
| Tooltip data-flow (donut, bars, trend) | ✅ | Data traced from API types → useMemo → chart props → tooltip render; all fields are required `number` in API schema, with `\|\| 0` defensive guards at consumption points |
| `formatPercent` safety | ✅ | `src/pages/Dashboard/index.tsx:155` — `!total` guard returns `'0%'`; no division-by-zero reachable |
| `yFor` (trend SVG coordinate mapping) | ✅ | `index.tsx:571` — guarded by `maxValue <= 0` early return rendering `<EmptyChart>`; never divides by zero |
| Trend share mode with empty data | ✅ | `index.tsx:565-569` — `hasTrendData` check + `maxValue <= 0` guard; test in `index.test.tsx` covers this path |
| Log scale bar widths | ✅ | `index.tsx:453` — `Math.log1p(value) / Math.log1p(max)` preserves ordering; `Math.max(4, ...)%` ensures minimum visibility |
| IntersectionObserver lifecycle | ✅ | `index.tsx:1025-1047` — cleanup via `observer.disconnect()` in effect return; section IDs on always-rendered `<div>` elements |
| i18n key parity (en-US / zh-CN) | ✅ | All new keys (`nav.*`, `kpi.scope.*`, `kpi.group.*`, `charts.tooltip.*`, `charts.trend.absolute/share`, `charts.topSpaces.scaleLog`, `spaces.column.messages`, `spaces.toolbar.total`) present in both locale files |
| appBots.test.ts assertions | ✅ | All 16 assertions verified against actual JSON values in both locale files |
| CSS class coverage | ✅ | `.toolbar-spacer` defined at `admin.css:1486`; `.dashboard-toolbar-meta` at `admin.css:1280`; all new classes properly scoped under `.admin-shell` |
| Race-condition protection (refresh button) | ✅ | `fetchSpaces` uses sequence counter (`spacesSeq`); stale responses discarded |

---

## 2. Findings

### P2-1 — Default page size silently halved (20 → 10)

**Diff-scope:** **new** — introduced by this PR.

`PAGE_SIZE = 20` replaced with `MAIN_LIST_PAGE_SIZE = 10` (`index.tsx:62`). The spaces table and direct-chats table now default to 10 rows per page instead of 20. The drawer table retains 20 via `DRAWER_PAGE_SIZE`.

This is a behavioral change that affects every user's default view. Not a bug, but the PR description doesn't mention it. Users who relied on seeing 20 rows will need to manually change the page size.

**Suggestion:** Call this out in the PR description or release notes. Consider whether 10 is the right default given the previous 20-row convention.

### P2-2 — Lost sort dimensions on spaces and direct-chats tables

**Diff-scope:** **new** — columns removed by this PR.

- **Spaces table:** `human_msg_count` and `agent_msg_count` sortable columns replaced by a single `MessageSplitCell` column (`index.tsx:1357-1370`). Users can no longer sort by human-messages or agent-messages independently — only by the combined total.
- **Direct chats table:** `conv_type` sortable column removed; type pill merged into the Members column (`index.tsx:1395-1413`). Users can no longer sort by conversation type.

Both are intentional design trade-offs (visual density over sort flexibility), but worth documenting.

**Suggestion:** If sorting by these dimensions is still valuable, consider adding sort controls to the split-cell column header or a secondary sort dropdown.

### P2-3 — `compact` prop on `DonutChart` is dead code

**Diff-scope:** **new** — added by this PR with no caller.

`DonutChart` accepts `compact?: boolean` (`index.tsx:340`) with matching CSS (`.dashboard-chart-card-compact`, `.dashboard-chart-body.compact`), but no caller passes `compact={true}`. The PR description acknowledges this as "staged for a follow-up layout."

**Suggestion:** Acceptable if the follow-up is imminent. Otherwise, defer the prop and CSS until the consuming layout is ready to avoid shipping dead code.

### P2-4 — `totalPoints` computed but unused in share mode

**Diff-scope:** **new** — `totalPoints` was previously used for the area fill; this PR suppresses the area in share mode (`index.tsx:597`) but still computes `totalPoints` unconditionally (`index.tsx:574`).

Minor unnecessary computation. Not a performance concern (trend data is small), but the code could be cleaner.

**Suggestion:** Gate `totalPoints` and `totalPath` computation behind `displayMode === 'absolute'`.

---

## 3. Suggestions

1. **Page size change communication:** The 20→10 default is the most user-visible change in this PR. Consider mentioning it in the PR body or a changelog entry.
2. **Sort dimension preservation:** If the removed sort columns are used by admins, consider alternative sort affordances (e.g., a sort-by dropdown in the toolbar).
3. **Trend chart total line in share mode:** Currently only human and agent lines are drawn in share mode. Consider whether a "total" line (always at 100%) adds value or would be confusing.

---

## 4. Additional Observations

- **Anchor nav accessibility:** The sticky anchor nav uses `<a href="#id">` links with proper `aria-label` on the `<nav>`. The `scroll-margin-top: 72px` on sections accounts for the sticky nav height. Good implementation.
- **Keyboard accessibility:** All interactive chart elements (donut segments, bar rows, trend points, legend items) have `tabIndex={0}` and `aria-label` attributes. Focus-visible styles mirror hover styles. Good a11y coverage.
- **Responsive grid:** The metric grid degrades gracefully: 4→3→2→1 columns across breakpoints (1280/960/720px). The anchor nav becomes horizontally scrollable at 720px.
- **`!important` usage in `.dashboard-row-inactive`:** Necessary to override Ant Design's inline cell styles. Scoped to the specific class, so no bleed risk.

---

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Intermediate Transforms | Reaches Consumer? |
|---|---|---|---|
| Donut tooltip `item.value` | `DashboardOverview.message_composition` (API) → `messageComposition` useMemo | `items[].value` from API response | ✅ Yes — API type is required `number`; `total > 0` guard before rendering segments |
| Donut tooltip `total` | `items.reduce((sum, item) => sum + item.value, 0)` | Computed in `DonutChart` | ✅ Yes — early return `<EmptyChart>` if `total <= 0` |
| Bar tooltip `row.primary` | `DashboardSpaceItem.human_msg_count` (API) → `topSpaceMessageRows` useMemo | `primary: space.human_msg_count` (no `\|\| 0`) | ✅ Yes — `HorizontalBars` applies `row.primary ?? ...` fallback |
| Bar tooltip `row.secondary` | `DashboardSpaceItem.agent_msg_count` (API) → `topSpaceMessageRows` useMemo | `secondary: space.agent_msg_count` (no `\|\| 0`) | ✅ Yes — `HorizontalBars` applies `row.secondary ?? 0` fallback |
| Trend tooltip `row.total_msg_count` | `DashboardTrendItem.total_msg_count` (API) → `trendRows` state | Direct pass-through | ✅ Yes — required `number` in API type; `\|\| 0` at consumption |
| Trend `valueForHuman(row)` | `row.human_msg_count / row.total_msg_count * 100` (share mode) | Computed in `TrendLineChart` | ✅ Yes — `total > 0` guard prevents division by zero |
| Trend `valueForAgent(row)` | `row.agent_msg_count / row.total_msg_count * 100` (share mode) | Computed in `TrendLineChart` | ✅ Yes — same guard |
| `spacesTotal` (toolbar meta) | `listDashboardSpaces` API → `res.count \|\| 0` → `spacesTotal` state | Direct assignment in `fetchSpaces` | ✅ Yes — server-side total count, not client-side array length |
| `metricMeta` string | `selectedSpaceIds.length` + `rangeLabel` + i18n `t()` | Template via `t('kpi.meta.scopeRange', ...)` | ✅ Yes — included in `metricCards` useMemo deps |
| `MessageSplitCell` percentages | `record.human_msg_count` + `record.agent_msg_count` (API) | `total = human + agent`; `humanPercent = human/total*100` | ✅ Yes — `total > 0` guard via `minWidth: human > 0 ? 4 : 0` |

---

## Verdict

No P0 or P1 issues found. All data flows are correctly traced from API sources through intermediate transforms to consumption points, with appropriate guards for edge cases (division by zero, empty data, race conditions). The four P2 findings are design trade-offs and minor code quality observations, not blockers.

**[Octo-Q] verdict: APPROVE** — No blocking issues. The P2 items are suggestions for the author's consideration.

lml2468 (approved):
**APPROVE** — re-review at `bb8d9c1824cf`. Pure follow-up cleanup (1 file, +8/-8) that resolves the outstanding 🟡 nits from the last round. No new risk; the 🔴 fix from `387b6729bf94` is untouched and still test-guarded.

### Nits resolved (all verified)
- **Orphaned sort enum** (yujiawei / mochashanyao): `human_msg_count` / `agent_msg_count` removed from `SPACE_SORT_VALUES` (`index.tsx:81`), matching the merged "Message mix" column which sorts by `total_msg`. Correctly scoped — `CHANNEL_SORT_VALUES` keeps both because the channels drawer still exposes those columns as individually sortable (verified `channelSortBy === 'human_msg_count'/'agent_msg_count'` at L1508/L1518). `isSpaceSortBy` still typechecks; `tsc` clean.
- **IntersectionObserver env guard** (Jerry-Xin / Steve / me): `if (!('IntersectionObserver' in window)) return` added to the anchor-nav effect (`index.tsx:1031`), matching the existing `useLazySection` guard. Clarifying comment added too.
- **DonutChart React keys** (yujiawei): segment + legend keys changed to `${item.label}-${index}`, defending against duplicate labels.
- **Share-mode total path** tidy: `totalPoints`/`totalPath` now computed only in `absolute` mode (`[]`/`''` otherwise). Safe — both are consumed exclusively inside the `displayMode === 'absolute'` `areaPath` branch (L594); trend markers use `valueForHuman`/`valueForAgent`, untouched. Minor perf nicety on top of the prior crash fix.

### Verification
- `tsc --noEmit` clean.
- Regression tests pass: `index.test.tsx` (empty-trend → share toggle → empty state, the 🔴 guard) and `appBots.test.ts`.
- No dangling references to the removed enum values.
- The 9 `feature.test.ts` failures remain pre-existing on `main` (local jsdom `localStorage`), unrelated.

### Still open (non-blocking, optional follow-up)
- `DonutChart` `compact` prop/styles still have no caller (staged for a future layout — mentioned in the PR body).
- Top Spaces global fetch + client-side long-tail filtering — pre-existing from #85, tracked separately.
- Default main-list page size 20→10 is an intentional UX change; a one-line note in the PR description would help (mochashanyao).

Clean, surgical cleanup. LGTM.

Jerry-Xin (approved):
Scope check passed: this PR modifies dashboard UI, dashboard i18n, styles, and related tests within `Mininglamp-OSS/octo-admin`. No blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: In share mode, the trend chart axis and lines switch to percentages, but the legend still shows absolute totals at `src/pages/Dashboard/index.tsx:678`. That can be misleading next to the percentage scale built at `src/pages/Dashboard/index.tsx:555` and rendered at `src/pages/Dashboard/index.tsx:625`.
- 🟡 Warning: The new dashboard test heavily mocks Ant Design and API calls, and only checks that empty share mode remains renderable (`src/pages/Dashboard/index.test.tsx:116`, `src/pages/Dashboard/index.test.tsx:172`). It does not cover the new tooltip, table split-bar, inactive-row, or i18n rendering paths from the PR test plan.

**✅ Highlights**
- The PR is well-scoped to dashboard polish and keeps API fetch sequencing/race guards intact.
- Good regression coverage was added for preserving Agent terminology in App Bot binding copy.
- Tooltip content is centralized, and table number styling is scoped more carefully than a global `tabular-nums` approach.

**Verification**
- Could not run local checks because dependencies are not installed in this checkout: `vitest` and `tsc` were not found when running `npm test` and `npm run build`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-admin PR#88 Review Report
**PR:** Mininglamp-OSS/octo-admin#88 — feat(dashboard): hover tooltips and KPI visual polish
**Head SHA:** bb8d9c1824cf14c7c2ff8b77b9a3841b4a241efc
**Files:** 6 changed (+1173 / -175)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Tooltip data flow (donut) | ✅ | `DonutChart` receives `items` from `messageComposition`/`memberComposition` memos → `formatPercent(item.value, total)` with `total = items.reduce(...)` guard; `total <= 0` early-returns `EmptyChart` (index.tsx ~L350) |
| Tooltip data flow (bars) | ✅ | `HorizontalBars.segmentedTooltipLines()` uses `segmentTotal = Math.max(row.value, primary+secondary, 1)` — division-by-zero impossible (index.tsx ~L470) |
| Tooltip data flow (trend) | ✅ | `TrendLineChart` tooltip reads `row.total_msg_count`, `human_msg_count`, `agent_msg_count` directly from `DashboardTrendItem` API type; `agentRatio` guards `human > 0` (index.tsx ~L635) |
| Trend share mode | ✅ | `valueForHuman/Agent` compute `(count/total)*100` with `total > 0` guard → 0 fallback; `maxValue = 100` in share mode; `hasTrendData` check prevents empty-chart crash (index.tsx ~L555-575) |
| Log scale bars | ✅ | `scaledWidth` uses `Math.log1p(value)/Math.log1p(max)` — `log1p(0)=0`, `log1p(max)/log1p(max)=1`; `max <= 0` early-returns `EmptyChart` (index.tsx ~L450) |
| KPI section split | ✅ | `metricCards.slice(0,4)` → Scale, `.slice(4)` → Activity; 7 cards total in `metricCards` useMemo (index.tsx ~L1054-1118) |
| IntersectionObserver | ✅ | `useEffect([], [])` observes `DASHBOARD_SECTION_IDS` via `document.getElementById`; all 4 IDs are on always-rendered wrappers (`dashboard-kpis`, `-charts`, `-spaces`, `-direct`); cleanup via `observer.disconnect()` (index.tsx ~L1023-1048) |
| MessageSplitCell | ✅ | `total = human + agent`; `humanPercent = total > 0 ? ... : 0` — safe guard; `minWidth: human > 0 ? 4 : 0` prevents invisible slivers (index.tsx ~L310-335) |
| i18n parity | ✅ | en-US and zh-CN dashboard.json both add identical key sets (nav.*, kpi.scope.*, kpi.group.*, charts.tooltip.*, etc.); appBots.json intentionally untouched |
| appBots test guard | ✅ | `appBots.test.ts` asserts Agent terminology preserved in appBots.json — prevents accidental over-rename |
| Inactive row dimming | ✅ | `rowClassName` reads `record.is_active` from `DashboardSpaceItem` (API type confirms `is_active: boolean`); CSS uses `!important` to override Ant Design link colors — standard pattern |
| Tabular-nums scoping | ✅ | Applied to `.dashboard-number-column`, `.dashboard-message-split-head strong`, `.dashboard-bar-row strong`, tooltip portal, drawer — correctly avoids KPI display digits |

## 2. Findings

### No P0/P1 findings.

### P2 — Notes (non-blocking)

**P2-1: `compact` prop dead code (staged)**
- **Diff-scope:** new — PR adds `compact` prop to `DonutChart` + `.dashboard-chart-card-compact` CSS, but no caller passes `compact={true}`.
- **Impact:** Zero runtime impact. PR body explicitly notes this is staged for a follow-up.
- **Recommendation:** Acceptable as-is. If the follow-up slips, remove to keep the surface clean.

**P2-2: `SPACE_SORT_VALUES` drops `human_msg_count` / `agent_msg_count`**
- **Diff-scope:** new — PR removes these from the UI sort dropdown. The API type `DashboardSpaceSortBy` still includes them (backend still accepts them).
- **Impact:** Users lose the ability to sort the Spaces table by human-only or agent-only message counts. The combined "Message mix" column sorts by `total_msg` only.
- **Recommendation:** Intentional UX simplification per PR scope. If users request per-type sorting later, it can be re-added without backend changes.

**P2-3: Default page size 20 → 10**
- **Diff-scope:** new — `MAIN_LIST_PAGE_SIZE = 10` replaces `PAGE_SIZE = 20`.
- **Impact:** Users accustomed to 20 rows per page will see fewer rows by default. The page-size dropdown still allows 10/20/50/100.
- **Recommendation:** Acceptable UX decision. No code issue.

**P2-4: Dashboard test uses extensive mocks**
- **Diff-scope:** new — `index.test.tsx` mocks `antd`, `@ant-design/icons`, `react-i18next`, `../../api`, and `../../api/dashboard`.
- **Impact:** Test validates component structure and share-mode toggle rendering, but mock fidelity may drift from real Ant Design behavior over time.
- **Recommendation:** Acceptable for the coverage it provides. Consider adding a smoke test with real Ant Design components when test infra supports it.

## 3. Suggestions

- None blocking. The PR is well-structured with clean separation of concerns.

## 4. Additional Observations

- **Defensive coding quality is high:** Every data consumption point has proper guards (`total <= 0`, `total > 0`, `Math.max(..., 1)`, `human > 0`, `hasTrendData`). No path leads to NaN, Infinity, or division-by-zero.
- **CSS architecture is clean:** Tooltip styles are correctly scoped outside `.admin-shell` since Ant Design portals render at body level. The `!important` on `.dashboard-number-column` is a pragmatic override of Ant Design's inline cell styles.
- **Accessibility:** SVG segments and bars gain `tabIndex={0}` + `aria-label` for keyboard navigation. Anchor nav uses semantic `<nav>` with `aria-label`. Good practice.

## 5. Data-Flow Trace Summary

| Consumer | Upstream Source | Path Verified |
|----------|----------------|---------------|
| DonutChart `items` | `messageComposition` / `memberComposition` useMemo → `getDashboardOverview` API → `DashboardOverview.message_composition` | ✅ `total` guard, early return on empty |
| HorizontalBars `rows` (Top Spaces) | `topSpaceMessageRows` useMemo → `listDashboardSpaces` API → `DashboardSpaceItem.human_msg_count / agent_msg_count` | ✅ `primary` field newly added, `log1p` safe |
| HorizontalBars `rows` (Conv Type) | `convTypeMessageRows` useMemo → `DashboardOverview.message_composition` | ✅ `secondary` always set, tooltip path correct |
| TrendLineChart `rows` | `getDashboardTrend` API → `DashboardTrendResp.list` → `DashboardTrendItem` | ✅ share mode math guarded, absolute mode unchanged |
| KPI MetricCards | `getDashboardOverview` API → `DashboardOverview` fields | ✅ `metricMeta`/`privateMetricMeta` pre-computed, deps correct |
| MessageSplitCell | `DashboardSpaceItem.human_msg_count / agent_msg_count` from table `record` | ✅ `total > 0` guard, `minWidth` prevents invisible bars |
| IntersectionObserver `activeSection` | `document.getElementById(DASHBOARD_SECTION_IDS)` → always-rendered divs | ✅ IDs match JSX, cleanup on unmount |
| `is_active` for row dimming | `DashboardSpaceItem.is_active` from `listDashboardSpaces` API | ✅ Type confirmed in API schema |

---

[Octo-Q] verdict: **APPROVE** — No P0/P1 findings. All data flows traced and verified. Clean implementation with proper defensive guards. P2 notes are informational only.

yujiawei (approved):
# Code Review — PR #88 (octo-admin)

**Verdict: APPROVED** — a well-scoped, well-tested dashboard UX polish PR. No correctness, security, or build-breaking issues found. `tsc --noEmit` passes clean and all 25 unit tests (including the 2 new test files) pass locally against head `bb8d9c1`. The findings below are non-blocking nits/suggestions.

## What this PR does

- Adds hover tooltips (Ant Design `Tooltip`) to all four charts: donut, horizontal bars, trend line, with keyboard-focusable hotspots (`tabIndex`, `aria-label`).
- KPI visual polish: splits metric cards into "Scale" / "Activity" sections, replaces the `highlight` color flag with a `detail` line, unifies meta copy via `kpi.meta.scopeRange`.
- New trend display mode toggle (Absolute / Share), log-scale option for top-spaces bars, an in-page anchor nav driven by `IntersectionObserver`, a merged "Message mix" split cell for the spaces table, right-aligned numeric columns.
- Page-size constants split into main-list (10) vs drawer (20) with distinct option sets.
- Terminology change: "Agent(s)" → "Bot(s)" across dashboard i18n (both locales).

## Verification

| Item | Result |
|---|---|
| `tsc --noEmit` | ✅ clean |
| `vitest run` | ✅ 6 files / 25 tests pass |
| Section ids vs `IntersectionObserver` targets | ✅ all four (`dashboard-kpis/charts/spaces/direct`) present on always-rendered wrappers |
| `SPACE_SORT_VALUES` vs table columns | ✅ removed `human_msg_count` / `agent_msg_count` sort values match the removed sortable columns; merged column sorts by `total_msg` which is still listed |
| New i18n keys present in both en-US and zh-CN | ✅ all `nav.*`, `kpi.scope.*`, `kpi.group.*`, `charts.tooltip.*`, `charts.trend.absolute/share`, `charts.topSpaces.scaleLog`, `spaces.toolbar.total` present in both locales |
| Share-mode empty-state render | ✅ covered by new `index.test.tsx` |

## Findings

### P2 — Dead `compact` prop on `DonutChart`

`DonutChart` gained a full `compact` prop branch (size 152/54/18 vs 168/60/20 and a `.compact` class), but neither call site (message composition / member composition, around `index.tsx:1677` and `:1690`) passes `compact`. The branch and the `.dashboard-chart-body.compact` CSS are currently unreachable. Either wire it up where intended or drop it to avoid speculative dead code.

### P2 — Orphaned i18n keys left after the refactor

Several keys are no longer referenced anywhere in `src` (`*.tsx`/`*.ts`) after this PR but remain in both locale files:

- `kpi.meta.range`, `kpi.meta.selectedRange`, `kpi.meta.globalOnly`, `kpi.meta.filtered` — replaced by `kpi.meta.scopeRange` / `kpi.scope.*`.
- `spaces.column.totalMessages`, `spaces.column.humanMessages`, `spaces.column.agentMessages` — the three columns were merged into the new "Message mix" (`spaces.column.messages`) cell.
- `direct.column.type` — the standalone type column was folded into the members cell.

Not harmful, but they'll rot. Suggest removing them (and keeping en-US/zh-CN in sync) in this PR or a quick follow-up.

### Nit — "Agent" → "Bot" terminology

The rename is applied consistently across both locales and matches the existing `appBots` direction, so no objection. Just flagging it's a user-visible wording change spanning KPIs, charts, and table columns — worth a heads-up to product/docs if "Agent" appears in screenshots or help content elsewhere.

### Nit — `agentRatio` guard

In `TrendLineChart` the tooltip's `agentRatio` uses `human > 0 ? (agent/human) : '-'`, so periods with bot activity but zero human messages render `-`. That's a reasonable choice (avoids divide-by-zero); just confirming it's intentional rather than showing `∞`.

## Positives

- Tooltips are accessible: focusable `<g>` wrappers with descriptive `aria-label`s, not mouse-only.
- The `IntersectionObserver` effect correctly guards for `'IntersectionObserver' in window` and observes always-rendered section wrappers (comment explains the lazy-content nesting), with proper `disconnect()` cleanup.
- Share/absolute mode handles the empty-data path cleanly (`hasTrendData` short-circuit) and the new test locks that behavior in.
- `topSpaceMessageRows` now passes `primary` explicitly instead of deriving it — clearer and avoids the implicit `value - secondary` path.

PR #94, opened 2026-06-17
fix(spaces): wire space-view member list to new search endpoint (#93)

### LINKED ISSUE #93: Space-management member list: pagination missing and search not working
## Summary

In the **space-management view** (a space owner/admin managing their own space, `scope.kind === 'space'`), the member list **cannot paginate** and **search does not work**. Both functions only work correctly in the super-admin view.

Confirmed by code analysis (not yet manually reproduced).

## Affected area

- `src/pages/Spaces/SpaceMembersPanel.tsx`
- `src/hooks/useSpaceScope.ts` (`buildUserScope`)
- `src/api/space-user.ts` (`listSpaceUserMembers`)

### 1. Pagination never appears

The space-user endpoint `GET /v1/space/{spaceId}/members` returns a **bare array** with **no total count**:

```ts
// src/api/space-user.ts:89-95
export const listSpaceUserMembers = (
  spaceId: string,
  params: { page?: number; limit?: number } = {},
) =>
  unwrap(api.get<SpaceUserMember[]>(`/v1/space/${spaceId}/members`, { params }))
```

`buildUserScope.listMembers` therefore sets `count` to the length of the **current page** (≤ `page_size`):

```ts
// src/hooks/useSpaceScope.ts:182-206
return { count: filtered.length, list: filtered.map(...) }
```

The table only renders pagination when `total > PAGE_SIZE`:

```ts
// src/pages/Spaces/SpaceMembersPanel.tsx:325-337  (PAGE_SIZE = 20)
pagination={ total > PAGE_SIZE ? { ... } : false }
```

Since `total` always equals the current page length (≤ 20), `total > 20` is **always false** → the pagination control is **never shown**. Any space with more than 20 members makes members #21+ completely inaccessible.

### 2. Search does not work

The search box is rendered **unconditionally** — it never checks `scope.supportsMemberSearch`:

```ts
// src/pages/Spaces/SpaceMembersPanel.tsx:289-303
<Input placeholder={...} value={keyword} onChange={...} onPressEnter={handleSearch} />
<Button icon={<SearchOutlined />} onClick={handleSearch}>...</Button>
```

But in the space-management view `supportsMemberSearch: false` (`useSpaceScope.ts:178`), and the backend does not accept a `keyword` param. The keyword is applied as a **client-side filter over the already-loaded current page only**:

```ts
// src/hooks/useSpaceScope.ts:187-194
const keyword = params.keyword?.toLowerCase()
const filtered = keyword
  ? list.filter((m) => m.name?.toLowerCase().includes(keyword) || m.uid?.toLowerCase().includes(keyword))
  : list
```

So search only matches within the first ≤20 members. Combined with issue #1 (no pagination), any member beyond the first page can never be found — the search box looks functional but effectively cannot find them.

## Why super-admin view is fine

The super view uses `manager.listSpaceMembers` → `GET /v1/manager/spaces/{spaceId}/members`, which supports `keyword` / `page_index` / `page_size` and returns a `ListResp` with a real `count`. Hence both pagination and search work there.

## Expected behavior

In the space-management view:
- Pagination should work for spaces with more than `PAGE_SIZE` members.
- Search should either work across all members (server-side), or the search box should be hidden / disabled when `supportsMemberSearch` is false to avoid a misleading UI.

## Suggested fix directions

- Have the space-user members endpoint return a total count (e.g. `{ count, list }`) so the frontend can paginate correctly, or expose `member_count` consistently.
- Gate the search box on `scope.supportsMemberSearch`, and add `keyword` support to the space-user members endpoint if server-side search is desired.

## Reproduction (to be confirmed manually)

1. Log in as a space owner/admin (non-super).
2. Open the Space Admin console for a space that has more than 20 members.
3. Observe: no pagination control appears; searching only matches members on the first page.

an9xyz:
## Summary

Fixes #93. In the **space-management view** (`scope.kind === 'space'`, a space owner/admin managing their own space) the member list could not paginate and search did not work — both only worked in the super-admin view.

**Root cause:** the space-user endpoint `GET /v1/space/{id}/members` returned a **bare array with no total count**, so `buildUserScope.listMembers` reported `count = current-page length (≤ 20)`. The table only renders pagination when `total > PAGE_SIZE`, so it was never shown, and `keyword` was only filtered client-side over the already-loaded page.

This adapts the frontend to **[octo-server#389](https://github.com/Mininglamp-OSS/octo-server/pull/389)**, which adds `GET /v1/space/{id}/members/search` returning a `{ count, list }` envelope with server-side keyword matching (nickname / username / UID / email, phone by last 4 digits).

## Why it fits the server's permission gate

The new search endpoint requires **role ≥ 1 (admin/owner)**. The space admin console is only ever reached for spaces where the viewer is role ≥ 1 — `SpaceEntry.tsx` / `SpaceAdminLayout.tsx` filter to `managed = list.filter(s => s.role >= 1)` and redirect otherwise — so the gate is always satisfied here.

## Changes

- **`src/api/space-user.ts`** — replace `listSpaceUserMembers` (bare array) with `searchSpaceUserMembers` hitting the new endpoint and returning a real `count`.
- **`src/hooks/useSpaceScope.ts`** — drive `buildUserScope.listMembers` via the new endpoint (real `count` → pagination works; server-side `keyword`) and set `supportsMemberSearch: true`.
- **`src/pages/Spaces/SpaceMembersPanel.tsx`** — gate the search box on `scope.supportsMemberSearch` and collapse the now-equivalent placeholders into one i18n key.
- **i18n (zh-CN / en-US)** — unify the search placeholder to reflect the searchable fields; drop the redundant `placeholderSuper`.
- **`src/api/space-user.test.ts`** — regression test locking the endpoint path, param names, and `{ count, list }` envelope.

## Testing

- `tsc --noEmit` clean
- `vitest run` — 31/31 pass (2 new)
- `vite build` succeeds

## Notes

- Depends on the server contract from octo-server#389 being deployed.
- The old `GET /v1/space/{id}/members` endpoint is left untouched server-side but is no longer used by this client.

https://claude.ai/code/session_01M2ybNmM3DepUvRFKiB16YF

---
_Generated by [Claude Code](https://claude.ai/code/session_01M2ybNmM3DepUvRFKiB16YF)_


lml2468 (approved):
**APPROVE** — clean, well-scoped fix for the space-view member pagination/search bug (#93), with a real regression test. Verified against head `4f3a05be2f0e`: `tsc` clean, new tests pass, `vite build` succeeds.

### Correctness — verified
- **Root cause & fix are right**: the old `GET /members` returned a bare array, so `buildUserScope.listMembers` set `count = page length (≤20)`, and the table only shows pagination when `total > PAGE_SIZE` → never appeared; keyword was filtered client-side over the loaded page only. Switching to `GET /members/search` returning `{ count, list }` gives a real total (pagination works) and server-side keyword matching. Confirmed the new `count` flows through to `setTotal` in the panel's `fetchData`.
- **Search now wired correctly**: `supportsMemberSearch: true` for user scope; the panel gates the search box on `scope.supportsMemberSearch` and `fetchData` passes `keyword: kw || undefined` → `listMembers` → the new endpoint. Super scope is unchanged (passes `params` straight to `manager.listSpaceMembers`, keyword included), so super search still works — the placeholder unification is purely cosmetic.
- **Permission fit holds**: the new endpoint requires role ≥ 1, and the space admin console is only reachable for `role >= 1` spaces (`SpaceEntry`/`SpaceAdminLayout` filter `managed = list.filter(s => s.role >= 1)` and redirect otherwise), so the gate is always satisfied. Frontend depends on octo-server#389 being deployed (correctly called out as a dependency).
- No leftover references to the removed `listSpaceUserMembers`; `unwrap` just returns `.data`.

### Tests
- New `space-user.test.ts` locks the endpoint path, the new param names (`keyword`/`page_index`/`page_size`, not the old `page`/`limit`), and that `count` comes from the envelope (137) not the page length — exactly the regression that caused #93. Also covers keyword omission. Good, targeted coverage.

### Nit (non-blocking)
- The "omits keyword" test calls `searchSpaceUserMembers` directly without a `keyword` key, so it asserts a clean `{ page_index, page_size }`. The real caller in `useSpaceScope` always builds `{ keyword: params.keyword, ... }`, i.e. `keyword: undefined` when empty. Harmless — axios drops `undefined` query params at serialization, and the search box passes `kw || undefined` — but the test exercises the API layer directly rather than the through-`useSpaceScope` path, so it wouldn't catch a regression where the hook started sending an empty-string keyword. Optional: add a hook-level case or pass `''`/`undefined` explicitly. Not worth blocking.

### Verification
- `tsc --noEmit` clean; `vite build` succeeds.
- `space-user.test.ts` (2 new) passes; suite 22 pass. The 9 `feature.test.ts` failures are pre-existing on `main` (jsdom `localStorage`) and unrelated — PR touches no feature files.
- i18n: placeholder unified, `placeholderSuper` removed from both locales (no dangling refs).

Solid fix with the right test. LGTM.

Jerry-Xin (approved):
The PR is relevant to `octo-admin` and correctly wires the space-management member table to the new scoped search endpoint.

💬 Non-blocking
- 🟡 Warning: This frontend now depends on `GET /v1/space/{id}/members/search` being available. If `octo-admin` can be deployed ahead of the matching `octo-server` release, the space-admin member list will fail with 404 until the backend is updated. The PR notes this dependency, so this is a rollout concern rather than a code blocker.

✅ Highlights
- The `SpaceScope` abstraction remains clean: super-admin still uses manager APIs, while space-admin uses the scoped user API.
- Pagination now uses the server-provided `count`, fixing the root cause for member lists larger than one page.
- Search support is explicitly advertised through `supportsMemberSearch`, and stale `placeholderSuper` references were removed.
- Regression coverage in `src/api/space-user.test.ts` locks the endpoint path, params, and response envelope.

Verification performed:
- `npm test` passed: 31 tests.
- `npm run build` passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-admin PR#94 Review Report (automated review)
**PR:** Mininglamp-OSS/octo-admin#94
**Head SHA:** `4f3a05be2f0ec83e128570e570b8d191c82962d2`
**Title:** fix(spaces): wire space-view member list to new search endpoint (#93)
**Files:** 6 changed (+115 / −36)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| New API function `searchSpaceUserMembers` hits correct endpoint | ✅ | `space-user.ts:116` → `GET /v1/space/{id}/members/search` |
| `{ count, list }` envelope correctly unwrapped | ✅ | `space-user.ts:112-118`, `useSpaceScope.ts:183-193` |
| Client-side keyword filter removed (now server-side) | ✅ | `useSpaceScope.ts` diff: 7 lines of `.filter()` removed |
| `supportsMemberSearch: true` for space scope | ✅ | `useSpaceScope.ts:178` |
| Search UI gated on `scope.supportsMemberSearch` | ✅ | `SpaceMembersPanel.tsx:289` |
| Old `listSpaceUserMembers` fully removed, zero dangling refs | ✅ | grep confirms no remaining references |
| Old `placeholderSuper` i18n key fully removed | ✅ | grep confirms no remaining references |
| Permission gate (role ≥ 1) verified via 3-layer frontend defense | ✅ | `SpaceEntry.tsx:98`, `SpaceAdminLayout.tsx:139-143`, Skeleton gate prevents API during validation |
| Server-side role gate confirmed | ✅ | octo-server#389: `role >= admin` + `checkSpaceActive` |
| Test covers endpoint path, params, envelope shape | ✅ | `space-user.test.ts:24-63` |
| `tsc --noEmit` / `vitest run` / `vite build` reported clean | ✅ | PR body (31/31 tests pass, 2 new) |

## 2. Findings

### F1 — Deployment dependency: octo-server#389 not yet merged (P2 note)

**Diff-scope:** new — introduced by this PR's endpoint switch.

This PR replaces the old `GET /v1/space/{id}/members` (bare array) with `GET /v1/space/{id}/members/search` ({ count, list } envelope). The server endpoint lives in octo-server#389, which is currently **OPEN** (not merged, not deployed).

If this frontend PR is deployed before the server PR:
- The space-admin member list will 404 (was already broken — no pagination/search — so not a regression in user experience, but the list will show empty/error instead of the degraded 20-item bare array).

**Severity:** P2 (deployment coordination, not a code defect). The code is correct against the documented server contract.
**Recommendation:** Merge octo-server#389 first, or coordinate deploy order.

### F2 — `status: 1` hardcoded in member mapping (P2 note, pre-existing)

**Diff-scope:** pre-existing — carried forward from old code, NOT introduced by this PR.

In `useSpaceScope.ts:191`, `buildUserScope.listMembers` maps search results with `status: 1` hardcoded. The `SpaceUserMemberSearchItem` type (`space-user.ts:89-98`) does not include a `status` field.

**Cross-reference with server contract (octo-server#389):** The server endpoint returns **only `status=1` (active) members** by design. Removed/banned members are excluded server-side. So the hardcode is correct against the server contract.

**Comparison with super-admin path:** `buildSuperScope.listMembers` (`useSpaceScope.ts:130-131`) passes through real `SpaceMember.status` from the manager endpoint, which can return non-active members. The asymmetry is intentional — different endpoints, different contracts.

**Severity:** P2 (pre-existing, correct against documented contract). A code comment documenting the assumption would help future maintainers.

## 3. Data-Flow Trace

| Consumed data | Upstream source | Verified flow |
|---|---|---|
| `res.count` (pagination total) | Server `{ count, list }` envelope → `searchSpaceUserMembers` → `unwrap` → `.count` | ✅ Flows to `SpaceMembersPanel.setTotal` → `Table.pagination.total`. Pagination appears when `total > PAGE_SIZE` (20). |
| `res.list` (member rows) | Server envelope → `.list` → `.map()` → `MemberItem[]` | ✅ Flows to `SpaceMembersPanel.setData` → `Table.dataSource`. |
| `params.keyword` | `SpaceMembersPanel.keyword` state → `fetchData(kw)` → `kw \|\| undefined` → server param | ✅ Empty string becomes `undefined` (omitted by axios). Non-empty sent to server for substring match. |
| `params.page_index / page_size` | `fetchData(nextPage)` → `PAGE_SIZE=20` constant | ✅ Correctly forwarded as `page_index` / `page_size` (matching server param names). |
| `scope.supportsMemberSearch` | `buildUserScope` → `true` (line 178) | ✅ Gates search Input + Button in `SpaceMembersPanel.tsx:289`. |
| Permission gate (role ≥ 1) | `SpaceEntry.tsx:98` filters `managed = list.filter(s => s.role >= 1)` → `SpaceAdminLayout.tsx:139-143` re-validates on mount | ✅ Three-layer defense prevents role=0 from reaching the search API. Server independently enforces. |

## 4. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. This PR only modifies the space-user (buildUserScope) path. The super-admin (buildSuperScope) path is untouched. No symmetric add/remove pair affected.
- **C2 — Control-flow ordering / nesting:** Clear. `searchSpaceUserMembers` has a single call-site (`buildUserScope.listMembers`). No nested/double-invocation risk. No security controls (regex/escape/sanitize) in this diff.
- **C3 — Authorization boundary ≠ capability boundary:** Clear. The new endpoint is reachable only through `scope.api.listMembers`, which is only called from `SpaceMembersPanel.fetchData`. The panel is only rendered inside `SpaceAdminLayout` (space view) or `SpaceDetailDrawer` (super view), both gated. Server independently enforces role ≥ 1.
- **C4 — Authorization lifecycle / container-member state cascade:** Clear. Server-side `checkSpaceActive` rejects banned (status=2) and disbanded (status=0) spaces before the member query runs. Frontend `SpaceAdminLayout.canEdit` also checks `detail.status === 1`. The cascade is properly handled at both layers.

## 5. Cross-Round Blocker Recheck (R6)

## 6. Additional Observations

- **Test quality:** The two new tests in `space-user.test.ts` mock the axios instance and assert the exact request shape (path, param names, envelope unwrap). This is good regression coverage for the API contract. The tests don't cover error paths, but `unwrap` is a shared utility and error handling is tested elsewhere.
- **`SpaceUserMember` type retained:** The old type (`space-user.ts:19-25`) is still used for `updateSpaceUserMemberRole`'s role parameter type. Not dead code — correct.
- **i18n unification clean:** Both `zh-CN` and `en-US` drop `placeholderSuper` and update `placeholder` to the broader field list. No orphaned keys.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The code correctly wires the space-view member list to the new search endpoint, with proper data-flow, permission gates verified at 3 frontend layers + server-side, and clean removal of the old client-side filter and dead i18n keys. Two P2 notes: (1) deployment must follow octo-server#389 landing, (2) the pre-existing `status: 1` hardcode is correct per server contract but could use a comment.

yujiawei (approved):
# Code Review — PR #94 (octo-admin)

Reviewed at head `4f3a05be2f0ec83e128570e570b8d191c82962d2` against base `main`. Verified locally: `tsc --noEmit` clean, `vitest run` 31/31 pass, `vite build` succeeds.

## Verdict: APPROVED

The change correctly rewires the space-management member list to the new `GET /v1/space/{id}/members/search` endpoint, fixing the missing-pagination and broken-search bugs described in #93. The root-cause analysis in the PR description is accurate: the old `GET /members` returned a bare array with no total, so `count` collapsed to the current page length and pagination never rendered. The new `{ count, list }` envelope fixes that, and `keyword` now matches server-side instead of over the already-loaded page.

## What I verified

- **Permission gate is satisfied (not a regression).** The new endpoint requires role ≥ 1. `buildUserScope.listMembers` now *always* calls `searchSpaceUserMembers`, even with no keyword (`src/hooks/useSpaceScope.ts:182`). This is safe under current routing: `SpaceAdminLayout` filters `/space/my` to `managed = role >= 1` and redirects away otherwise (`src/pages/SpaceAdmin/SpaceAdminLayout.tsx:139`, `:168`), and `MembersTab` passes the loaded `detail.role` into the scope (`src/pages/SpaceAdmin/tabs.tsx:17`). The `<Outlet>` only renders after `detail` resolves (`SpaceAdminLayout.tsx:332`), so a role-0 viewer never reaches this panel through the routed admin console.
- **Super-admin path is unaffected.** `SpaceDetailDrawer` calls `useSpaceScope()` with no role, but it is mounted only under the super-admin Spaces page; with `scope === 'super'` the hook returns `buildSuperScope`, which uses the `/v1/manager/...` member endpoint, not the new search endpoint (`src/hooks/useSpaceScope.ts:241`, `:129`). The `role ?? 0` default is never exercised here.
- **Pagination mapping is correct.** Panel sends `page_index` / `page_size` (`SpaceMembersPanel.tsx:58`), user scope forwards the same names unchanged (`useSpaceScope.ts:183`), and the API method passes them through (`space-user.ts:110`). The earlier `page`/`limit` → `page_index`/`page_size` mismatch is gone.
- **Keyword omission is correct.** The panel passes `keyword: kw || undefined` (`SpaceMembersPanel.tsx:61`); the axios serializer drops `undefined`-valued params, so an empty search issues an unfiltered request. The new regression test locks this (`space-user.test.ts`).
- **No dangling references.** `listSpaceUserMembers` is fully removed; a repo-wide search returns no remaining callers, consistent with the clean `tsc` build.
- **The `supportsMemberSearch` gate is sound.** Both scope builders set `supportsMemberSearch: true` (`useSpaceScope.ts:125`, `:178`), so collapsing the search-box render behind that flag does not hide the box for super-admins.

## Non-blocking suggestions (P2)

- **Coupling of the plain member list to an admin-only endpoint.** Because `buildUserScope.listMembers` always hits the role ≥ 1 search endpoint, the *basic* member list now depends on the same permission gate as search. Today the routing guards guarantee role ≥ 1, so this is fine. But it removes the previous ability for a role-0 member to list members, and ties an everyday read to an admin gate. If a future entry point ever renders this panel for a role-0 member, the list (not just search) will fail. Consider a short comment at `useSpaceScope.ts:182` noting that this path is only valid for role ≥ 1, so the invariant is explicit for the next reader.
- **Server contract dependency.** This client now hard-depends on octo-server#389 (`{ count, list }` envelope, `keyword`/`page_index`/`page_size` params) being deployed. The PR notes this; just make sure the server side ships first/together to avoid a window where the member list 404s.

## Coverage note

This is a static + build/test review. I did not exercise the live backend or run browser-level checks, so the actual server response shape, the role ≥ 1 enforcement, and the phone-last-4 matching behavior of `GET /v1/space/{id}/members/search` are taken from the PR description and octo-server#389, not verified against a running server. Recommend a quick manual smoke test (admin space with > 20 members: pagination shows; search by nickname/UID/email returns server-filtered results) before/after the server deploy.

PR #100, opened 2026-06-17
ci: add always-run CI Gate aggregate job (OCT-7)

lml2468:
Part of OCT-7: make CI-green an enforced merge gate.

This repo's `build` job is a **matrix** job. A skipped matrix job collapses to a single un-suffixed `build` context, while a running one expands to per-combo contexts — so neither can be safely required directly without risking a perma-blocked PR. 

This adds an always-run **CI Gate** job (`if: always()`, `needs: [changes, build]`) that reports failure iff a needed job failed/was cancelled (skipped/success pass). After merge, branch protection will require only the `CI Gate` context.

Verified pattern on canary octo-server (OCT-7).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,3 +89,31 @@ jobs:
       - run: npx tsc --noEmit
       - run: npm test
       - run: npm run build
+
+  ci-gate:
+    name: CI Gate
+    # Aggregate gate: always runs so the required status check ALWAYS reports,
+    # even when path-filtered/matrix jobs are skipped on docs-only PRs. This is
+    # the robust pattern for required checks over path-filtered or matrix CI
+    # (a skipped matrix job collapses to a single un-suffixed context, so the
+    # per-combo contexts can never be safely required directly). Fails iff a
+    # needed job failed or was cancelled; 'skipped'/'success' count as passing.
+    # Branch protection should require ONLY the "CI Gate" context for this repo.
+    if: ${{ always() }}
+    needs: [changes, build]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify upstream CI results
+        env:
+          RESULTS: ${{ join(needs.*.result, ',') }}
+        run: |
+          set -euo pipefail
+          echo "needed job results: ${RESULTS}"
+          IFS=',' read -ra items <<< "${RESULTS}"
+          for r in "${items[@]}"; do
+            if [ "$r" = "failure" ] || [ "$r" = "cancelled" ]; then
+              echo "::error::CI Gate: a required job concluded '${r}'"
+              exit 1
+            fi
+          done
+          echo "CI Gate: all required jobs passed or were skipped"


mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-admin PR#100 Review Report — ci: add always-run CI Gate aggregate job (OCT-7)
**Head SHA:** `172f155612b3be80c8fb87c3413dd8f153bb951c`
**Files changed:** `.github/workflows/ci.yml` (+28 / −0)
**Routing:** complexity=simple (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `ci-gate` always runs (`if: always()`) | ✅ | `.github/workflows/ci.yml` L99 |
| `needs: [changes, build]` covers all upstream jobs | ✅ | `changes` (L29) and `build` (L72) are the only other jobs in this workflow |
| Failure detection logic (failure/cancelled → exit 1) | ✅ | L108–L115: iterates `needs.*.result`, exits non-zero on `failure` or `cancelled` |
| Skipped/success treated as passing | ✅ | L108–L115: only `failure` and `cancelled` trigger exit 1; `skipped` and `success` fall through |
| Bash syntax (`set -euo pipefail`, `read -ra`, `<<<`) | ✅ | L106–L115: valid bash; `read -ra` returns 0 when input has data |
| No injection risk (`needs.*.result` is GH-controlled) | ✅ | Values are constrained to `success|failure|cancelled|skipped` by the runtime |

## 2. Findings

**No P0/P1/P2 findings.**

### Data-flow trace (per iron rule)

| Consumed data | Upstream source | Runtime value reaches consumer? |
|---|---|---|
| `needs.*.result` | GitHub Actions runtime; array of result strings from `changes` and `build` jobs | ✅ Yes — `join(needs.*.result, ',')` produces a comma-separated string assigned to `$RESULTS` |
| `$RESULTS` env var | `join()` expression at L105 | ✅ Yes — passed via `env:` block to the step |
| `items[]` array | `read -ra items <<< "${RESULTS}"` at L108 | ✅ Yes — bash here-string feeds `RESULTS` into `read`, which splits on `,` |
| Each `r` in loop | `for r in "${items[@]}"` at L109 | ✅ Yes — iterates each result token |

### Scenario matrix

| Scenario | `changes` result | `build` result | `ci-gate` behavior | Correct? |
|---|---|---|---|---|
| Code PR, all pass | success | success | pass | ✅ |
| Code PR, build fails | success | failure | **fail** | ✅ |
| Docs-only PR | success | skipped | pass | ✅ |
| Draft PR | success | skipped | pass | ✅ |
| `changes` job fails | failure | skipped (dep failed) | **fail** (catches `changes` failure) | ✅ |
| `build` cancelled (e.g. concurrency) | success | cancelled | **fail** | ✅ |
| Both cancelled (workflow cancel) | cancelled | cancelled | **fail** | ✅ |

## 3. Suggestions

None. The implementation is clean and follows the well-established aggregate-gate pattern (also verified on octo-server per PR body).

## 4. Additional Observations

- **Concurrency interaction:** The workflow has `cancel-in-progress: true` (L24). When a new push cancels the old run, the old `ci-gate` is also cancelled. This is expected — the new run's `ci-gate` will report for the new head. No issue.
- **Branch protection:** After merge, branch protection should require only the `CI Gate` context (as noted in the PR body and the job's comments). The existing `build (18)` / `build (20)` per-matrix contexts should be removed from required checks.

## 5. R5 Blind-spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or create/delete paths in this CI-only change.
- **C2 (control-flow ordering / nesting):** N/A — single linear bash script, no nested calls or reused security controls.
- **C3 (authorization boundary ≠ capability boundary):** N/A — no endpoints, tools, or credentials involved.
- **C4 (authorization lifecycle / container-member cascade):** N/A — no auth changes.

## 6. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE** — No P0/P1/P2 findings. The CI Gate job correctly implements the always-run aggregate pattern: `if: always()` ensures it reports on every run, `needs.*.result` captures all upstream job outcomes, and the failure/cancelled check correctly gates the merge. Skipped and success are treated as passing, which is correct for path-filtered and draft-PR scenarios.

yujiawei (approved):
# Code Review — PR #100 (octo-admin)

## Verdict: APPROVED

This adds a `CI Gate` aggregate job to `.github/workflows/ci.yml` so a single required status check always reports, even when path-filtered or matrix jobs are skipped. The implementation is the correct, robust pattern for this problem. No blocking issues found.

## What it does

`ci-gate` runs with `if: ${{ always() }}`, depends on `[changes, build]`, collects `join(needs.*.result, ',')` into `RESULTS`, and fails iff any needed job concluded `failure` or `cancelled`. `success` and `skipped` pass. Branch protection is intended to require only the `CI Gate` context.

## Correctness analysis (no false-green found)

I specifically hunted for any path where the gate goes green while a real job failed or failed to run:

- **Matrix failures are not hidden.** `build` runs a `node-version: [18, 20]` matrix (`ci.yml:79-81`). Downstream `needs.build.result` collapses the matrix to one logical result, but without `continue-on-error`, any failed leg makes the `build` dependency `failure`, which the loop catches at `ci.yml:113-116`. A single broken Node version cannot slip through.
- **`changes` failing is covered.** `changes` is listed directly in `needs` (`ci.yml:103`), so a paths-filter error surfaces as `failure` in `RESULTS` and fails the gate.
- **`changes` fail → `build` skip is still caught.** If `changes` fails, `needs.changes.outputs.code` is empty, so `build`'s `if` (`ci.yml:75-77`) evaluates false and `build` is skipped. The gate still sees `changes.result == failure` and fails. No false-green via the skip path.
- **Docs-only skip is intentional and safe.** On docs-only PRs `code` resolves to `false` (`ci.yml:42`, `ci.yml:77`), `build` is skipped, the gate passes on `skipped`. Drafts skip `build` too (`ci.yml:75-77`), and drafts are not mergeable; `ready_for_review` is in the trigger list (`ci.yml:11`) so the real run fires before merge.
- **`set -euo pipefail` is fine.** `RESULTS` is always populated from `env` (`ci.yml:108`) and runners use Bash 5.x; the empty-array expansion concern does not apply here.
- **Requiring only `CI Gate` is coherent** because workflow-level path filtering is deliberately avoided (`ci.yml:16-20`) and the gate always produces a status against the head SHA.

## Suggestion (non-blocking, P2 / maintainability)

`ci-gate.needs` is the single source of truth for what the gate enforces. If a future required job is added to this workflow, it must also be added to `needs: [changes, build]` or the gate silently stops covering it. Worth a short inline comment on the `needs:` line, or a follow-up note, so the coupling isn't forgotten. Not a defect in this diff.

## Notes

- The inline comments explaining the pattern (`ci.yml:95-101`) are accurate and genuinely helpful for the next maintainer.
- Verified against head SHA `172f155612b3be80c8fb87c3413dd8f153bb951c`; single file, 28 insertions, no deletions.

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/octo-admin` and the CI aggregate gate is ready to merge.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- 🔵 The new `ci-gate` job in `.github/workflows/ci.yml:93` correctly uses `if: ${{ always() }}` with `needs: [changes, build]`, so it still reports when upstream jobs are skipped.
- 🔵 The failure logic in `.github/workflows/ci.yml:108-119` correctly fails only on `failure` or `cancelled`, while allowing `success` and `skipped`, matching the intended branch-protection strategy for path-filtered and matrix CI.
- 🔵 YAML parsing succeeds locally for the workflow set. `actionlint` was not installed in the environment, so I could not run that additional static check.

PR #110, opened 2026-06-27
ci(octo-issue-feed): opt into webhook-only triage

lml2468:
Adds enable_im_notify: false. Backed by Mininglamp-OSS/.github#80.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -16,6 +16,7 @@ jobs:
       issue_url: ${{ github.event.issue.html_url }}
       issue_author: ${{ github.event.issue.user.login }}
       event_action: ${{ github.event.action }}
+      enable_im_notify: false
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
       TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository's issue triage automation, but it currently can break the workflow because the called reusable workflow ref does not accept the new input.

🔴 Blocking

- 🔴 Critical — `.github/workflows/octo-issue-feed.yml:19` passes `enable_im_notify: false` to `Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1`, but the `@v1` reusable workflow contract does not declare `enable_im_notify`. The input is currently declared only on the `.github` repo's default branch (`main`); neither the `v1` tag nor `v1.1.2` includes it. GitHub reusable workflows reject undeclared inputs at workflow-planning time, so the `notify` job will fail before any step runs on every `issues: [opened, reopened]` event once this merges. Fix: bump the caller ref to a tag/SHA that actually declares `enable_im_notify` (release a new `v1.x` from `main` and/or move the `v1` floating tag forward), then update this `uses:` ref — or hold this PR until `v1` is updated.

💬 Non-blocking

- 🔵 Suggestion — `.github/workflows/octo-issue-feed.yml:22` now lacks a trailing newline. Harmless for GitHub Actions but worth restoring for text-file hygiene.

✅ Highlights

- The change is narrowly scoped and aligns with the stated goal of webhook-only triage once the reusable-workflow contract mismatch is resolved.

PR #115, opened 2026-07-16
feat(system-mcp): admin surface for platform-provided MCPs

l-s-c:
## Summary

Adds a new **System MCP** page to the admin console — list / detail /
create / edit / soft-delete every `visibility=system` MCP across all
Spaces. Backed by the octo-marketplace admin surface
(`docs/api/mcp-v1.md §9`); no new server ships here.

Layout follows the console's Users / Spaces / SystemSetting patterns:

- **List**: `page-title` header + `.toolbar` + antd Table. Row click
  opens detail; no per-row action column.
- **Detail**: read-only Drawer with `className=\"admin-shell admin-drawer\"`
  so the console's `--a-*` design tokens cascade into the portalled
  body. Sections mirror octo-web's dmworkmcp order (connection /
  tools / examples / faqs / notes). Inline-swap delete confirm on
  the footer (never a nested modal stack).
- **Create + Edit**: shared 3-step wizard Modal (basics → connection
  → docs), structurally aligned with `octo-web`'s
  `dmworkmcp/McpCreateModal` so the two consoles feel like one
  product. antd primitives replace Semi UI; visible flow is
  identical.

## ⚠️ Update — response to review (2026-07-16)

The prior blocking review flagged that `VITE_MARKETPLACE_ADMIN_TOKEN`
was baked into the JS bundle at build time, so the `mcp.read` /
`mcp.write` capability distinction was UI-only — anyone with `mcp.read`
could grab the token from devtools and call the write endpoints.

**Commit 27535e9** replaces the shared secret with per-user auth:

- `mcp.ts` now sends the operator's own Octo login token in the `Token`
  header (same header every other admin surface uses). No X-Admin-Token.
- 401 handler mirrors the shared axios interceptor: `logout()` + hard
  redirect to `/admin/login`.
- `VITE_MARKETPLACE_ADMIN_TOKEN` retired from `.env.example` / `.env`.

**Depends on companion PR**
[`kense-lab/octo-marketplace#10`](https://github.com/kense-lab/octo-marketplace/pull/10)
which adds `WrapMarketAdmin` middleware — resolves the token via
octo-server `/v1/auth/verify` and admits only `role=superAdmin`. That
PR must merge and deploy **before** this one, otherwise every admin
MCP request will 401 (marketplace still expecting X-Admin-Token).

## ⚠️ Update 2 — prod nginx routing (2026-07-16)

Second blocking review flagged that `nginx.conf.template` only proxies
[...]

PR #116, opened 2026-07-20
feat: add MCP, skill market, and system skill admin

LarsonLian:
## Summary
- add admin console pages and APIs for platform-provided MCP management
- add Skill Market category/public skill administration
- add System Skill category/skill management
- wire navigation, permissions, i18n, upload/download flows, probe helpers, and dev proxy config

## Testing
- Not run in this PR creation step


Jerry-Xin (changes requested):
The PR is relevant to octo-admin, but it introduces two authorization vulnerabilities that block merging.

## 🔴 Blocking

1. 🔴 Critical — Marketplace administrator credentials are exposed to every browser.

   `VITE_MARKETPLACE_ADMIN_TOKEN` is compiled into the frontend bundle and attached to requests in `src/api/mcp.ts:22-32` and `src/api/skill.ts:19-29`. Any visitor who downloads the JavaScript can extract this token and directly invoke marketplace write/delete endpoints, bypassing `mcp.write` and `skill.write`.

   Move these operations behind an authenticated server-side/BFF endpoint that enforces the current administrator’s capabilities. No marketplace administrator credential should be shipped to the browser.

2. 🔴 Critical — Read-only users can mutate Skills through the hidden route.

   `src/App.tsx:249-253` allows `/system-skill` with only `skill.read`, but that page exposes unrestricted writes:

   - Upload: `src/pages/SystemSkill/SkillTable.tsx:158-160`
   - Category create/edit/delete: `src/pages/SystemSkill/CategoryTab.tsx:49-76,115-127`
   - Skill create/update: `src/pages/SystemSkill/SkillFormModal.tsx:170-208`

   Unlike the `/skill-market` implementation, this surface never checks `skill.write`. Remove the duplicate route or consistently gate every mutation with `skill.write`; server-side enforcement remains mandatory.

## 💬 Non-blocking

- 🟡 Warning — Both navigation locale files define `systemMcp` twice (`src/i18n/locales/en-US/nav.json:11,15` and `zh-CN/nav.json:11,15`). JSON parsing silently keeps the latter value.
- 🟡 Warning — `/skill-market` and `/system-skill` are duplicate administration implementations with diverging permissions and upload behavior. Consolidating them would prevent further security and maintenance drift.
- 🔵 Suggestion — Add API and permission tests covering read-only administrators, mutation rejection, upload failures, and Skill pagination. Current new tests cover only MCP probe helpers.

## ✅ Highlights

- TypeScript checking and the production Vite build succeed.
- MCP probe helpers have useful edge-case tests.
- The presigned uploads avoid leaking the admin header to object-storage hosts.

The full test run had 40 passing tests and 9 unrelated `localStorage` failures under the current Node 26 environment.

lml2468 (changes requested):
## Review @ `33709f62074c` — 独立本地复核（非转述）

从 merge-base `3e427999`(= 当前 `main`)的本地 checkout 出发,实际安装依赖、编译、构建、跑测试后得出结论。**结论:REQUEST_CHANGES** —— 存在 2 个 P1 阻塞。

### 验证记录（实际执行）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build`(tsc && vite build)— ✅ 3216 modules，产物正常
- 本 PR 自带 `probeHelpers.test.ts` — ✅ 14/14
- 全量 `vitest run` — 9 处失败**全在 `src/store/feature.test.ts`**;blame 核实本 PR 未触碰,既有问题,不阻塞。

---

### 🔴 Gate 2 阻塞发现

#### ① [P1 / 安全] `VITE_MARKETPLACE_ADMIN_TOKEN` 被编译进浏览器 bundle —— UI 能力门形同虚设

`src/api/mcp.ts:22` 和 `src/api/skill.ts:19` 均为:
```ts
const ADMIN_TOKEN = import.meta.env.VITE_MARKETPLACE_ADMIN_TOKEN || ''
// 拦截器:config.headers['X-Admin-Token'] = ADMIN_TOKEN
```
`.env.example` 里带默认值 `VITE_MARKETPLACE_ADMIN_TOKEN=dev-admin-token`,且注释自认「inlined into the JavaScript bundle at build time … visible in the browser DevTools」。

**实测证据**(我本地跑的):
```
VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED npm run build
→ grep 'PROVE_ME_LEAKED' dist/assets/*.js  →  CONFIRMED: 明文出现在打包产物里
```
**触发 → 后果**:marketplace 生产端需要 `X-Admin-Token`(dev 才 `AUTH_ENABLED=false`)。一旦生产配置该 token,它就被静态编译进 JS,任何拿到 `mcp.read` / `skill.read`(甚至任意能加载后台页面的人)都能从 DevTools 抠出 admin token,直接调用 `/admin/*` 的**写/删**接口 —— `mcp.read` vs `mcp.write` / `skill.read` vs `skill.write` 的能力区分在服务端被这枚共享密钥完全绕过。

这正是 **PR #115 已经修复过的头号阻塞**(#115 改用操作者自身 Octo 登录 token + 服务端 `role=superAdmin` 校验,bundle 里零密钥)。本 PR 独立分支,携带的是修复**之前**的写法,把该漏洞重新带回。
**修复**:对齐 #115 —— 删除 `VITE_MARKETPLACE_ADMIN_TOKEN` 与 `X-Admin-Token`,改用 `useAuthStore.getState().token` 走 `Token` 头,由 marketplace 回调 octo-server 校验 superAdmin。`mcp.ts` 与 `skill.ts` 两处都要改。

#### ② [P1 / 可靠性] 生产 nginx 缺 `/market/api/` 反代 —— 上线后所有 MCP/Skill 后台操作失效

`src/api/*.ts` 全部走 `/market/api/v1/*`;`vite.config.ts` 只加了 **dev** proxy。但 `nginx.conf.template` **本 PR 未触碰**,无 `/market` location(grep 命中 0)。生产环境 `/market/api/*` 会落到 SPA fallback(`/admin/index.html`)→ 返回 HTML,axios 报错,list/detail/create/edit/delete/probe/upload 全线 502/解析失败。

这也是 **#115 修复过的第二个阻塞**(#115 在 `nginx.conf.template` 加了对称的 `location /market/api/` + `MARKETPLACE_BACKEND` env + Dockerfile 默认值)。本 PR 未包含该修复。
**修复**:移植 #115 的 nginx `/market/api/` 块(`rewrite ^/market/(.*) /$1 break` + 变量式 `proxy_pass`)、Dockerfile 的 `MARKETPLACE_BACKEND` 默认与部署 runbook。

---

### 🟠 中等(强烈建议一并处理)

3. **[MEDIUM] SystemSkill 页面无写操作能力门控** — `SkillMarket/*.tsx` 正确用 `hasManagerCapability(..., 'skill.write')` 门控写按钮;但 `SystemSkill/index.tsx` + `SkillTable.tsx` 的「上传/新建、编辑、删除」入口**未做 `skill.write` 门控**,仅有 `skill.read` 的管理员也能看到并点击写操作。与控制台其余页面(含本 PR 的 SkillMarket)的一致性约定不符。服务端应兜底,但前端门控缺失属可修的信号缺陷。

4. **[MEDIUM] `mcp.ts` / `skill.ts` 无 401 处理** — token 过期/失效时不 `logout()` 也不跳转,用户停在坏页反复打 401。#115 的 `mcp.ts` 镜像了共享拦截器(401 → `logout()` + `/admin/login`),本 PR 的版本回退了这一点。

### 🟡 低优先级

5. **SystemMcp 部分是 #115 修复前的旧版**,携带以下已知问题:
   - `FormModal.tsx:handleSubmit` 用**本地化字符串比较**决定跳步(`err === t('form.nameRequired') || err.startsWith('服务标识')`)—— en-US 语言下 `startsWith('服务标识')` 永假,slug 校验失败会跳错步骤。#115 已改为 `{message, step}` 结构化返回。
   - 缺 bearer-auth sentinel(`__OCTO_SECRET_PLACEHOLDER__`)注入,bearer MCP 可存成无 Authorization 头。
   - `systemMcp.json` 的 `form` 内 `sectionTools`/`sectionToolsDesc` 重复键,后者覆盖前者。
6. **多处硬编码中文**(`FormModal.tsx` 空态/`+ 新增一条`、`slugInvalid` 的中文 defaultValue 等),en-US 界面显示中文。

---

### Gate 1 — 规格符合度：部分满足
新增 MCP / Skill Market / System Skill 三个后台面(路由 / 能力键 / i18n / 导航 / 上传下载 / probe)与描述一致,能力键 `mcp.*` / `skill.*` 已入 enum。但描述称已「wire … permissions」,而 SystemSkill 写门控缺失(发现 3),属**静默偏离**。

### 覆盖与盲区(需人工/后端联调)
- 未跑 marketplace 后端:skill 上传→parse 轮询→create、reupload、category 409、download presigned 全链路需真实 marketplace 环境手测。
- `ApiError` 新增可选 `details` 参数向后兼容,无回归。

---

**Verdict: REQUEST_CHANGES。** 两个 P1(密钥进 bundle 破坏权限模型、生产 nginx 缺 `/market` 反代导致上线即失效)必须修复。二者 #115 均已给出可直接移植的修复方案。建议要么把本 PR 里的 SystemMcp/marketplace-client 对齐 #115 的最终版,要么在 #115 合入后 rebase 复用其修复,避免把已修的漏洞重新引入 main。

mochashanyao (changes requested):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — P1 capability gate missing on SystemSkill pages (all write operations exposed without `skill.write` check, while sibling pages SkillMarket and SystemMcp correctly enforce it). Fix is mechanical: apply the same `canWrite`/`hasManagerCapability` pattern.

yujiawei (changes requested):
# Code Review — PR #116 (octo-admin)

Reviewed at head `33709f62074cee3ec93e7abb5871103c72d17adb` (merge-base `9e51cdc`). Scope: new MCP admin, Skill Market, and System Skill admin surfaces + marketplace API clients. `tsc && vite build` and `vitest run` (49 tests) both pass locally; `pnpm install --frozen-lockfile` succeeds. This is a security-sensitive change, so authorization and secret handling are called out for manual verification below.

## 1. Spec / scope

The diff matches what the PR description promises (MCP catalog admin, Skill Market category/skill admin, System Skill admin, nav + capability wiring, i18n, upload/download/probe flows, dev proxy). No out-of-scope feature creep spotted. **Scope: OK.**

One structural note (not a scope violation, since the body lists both): the PR ships **two** skill-admin surfaces — `SkillMarket` (nav-exposed, capability-gated) and `SystemSkill` (route-only, not in the nav, ungated). See P1 below — this duplication is where the main defect lives.

## 2. Findings

### 🔴 P1 — `SystemSkill` mutation controls are not gated by `skill.write` (privilege inconsistency)

`src/App.tsx:249` mounts the route behind **only** `skill.read`:

```tsx
<Route path="system-skill" element={
  <CapabilityRoute capability="skill.read"><SystemSkill /></CapabilityRoute>
} />
```

Its sub-components expose create/edit/delete with **no `skill.write` check**:
- `src/pages/SystemSkill/SkillTable.tsx:158` — upload button
- `src/pages/SystemSkill/DetailDrawer.tsx:108-119` — edit + delete (Popconfirm)
- `src/pages/SystemSkill/CategoryTab.tsx:115,123-139` — category create/edit/delete

This is directly inconsistent with the sibling `SkillMarket` page, which *does* gate every mutation on `skill.write` (`SkillMarket/SkillTab.tsx:25-26,210,264`, `SkillMarket/CategoryTab.tsx:19-20,145`). A manager holding only `skill.read` can navigate to `/system-skill` by direct URL and reach all mutation controls.

**Why this matters here specifically:** the marketplace client authenticates with a single shared `X-Admin-Token` and does **not** forward the per-user `token` (`src/api/skill.ts:28` vs the primary client `src/api/index.ts:26-28`). So the marketplace backend has no per-user identity to enforce `skill.write` on — the client-side capability check is effectively the only thing hiding these operations from a read-only manager. A missing client gate therefore has real authorization consequences, not just cosmetic ones.

**Ask:** either gate `SystemSkill`'s mutations on `skill.write` (parity with `SkillMarket`), or, if `SystemSkill` is superseded by `SkillMarket`, remove the redundant route so it can't act as an ungated back door.

### 🔴 P1 / needs human security sign-off — shared marketplace admin token is inlined into the browser bundle

`src/api/mcp.ts:22` and `src/api/skill.ts:19`:

```ts
const ADMIN_TOKEN = import.meta.env.VITE_MARKETPLACE_ADMIN_TOKEN || ''
```

Any `VITE_`-prefixed value is baked into the shipped JS at build time and is sent on every marketplace admin request (`X-Admin-Token`, `mcp.ts:31` / `skill.ts:28`). Anyone able to load the console assets can extract it from DevTools and then create/edit/delete/probe/upload marketplace records directly, bypassing all UI capability gating.

To the author's credit, `.env.example:28-38` documents this trade-off explicitly, recommends network-level defense-in-depth, and names a BFF follow-up. Because this is a deliberate, documented architectural decision (not an accidental leak), I'm flagging it for **explicit human security sign-off** rather than treating it as a code defect to rewrite in this PR. Please confirm before merge:
1. Marketplace admin endpoints are **not** reachable from the public internet (origin/network restricted), AND
2. The admin console static assets are not served to unauthenticated users (otherwise the token is extractable by anyone).

If either assumption doesn't hold in the target deployment, the BFF proxy should land before this ships.

### 🟡 P2 — `SkillMarket` upload has no client-side size/type guard

`src/pages/SkillMarket/SkillUploadModal.tsx:108` uses `accept=".zip"` + `beforeUpload={() => false}`; the submit path (`handleSubmit`, line 27+) obtains a presigned URL and uploads with no extension/MIME/size validation. The `SystemSkill` uploader already enforces `.zip` + a 20 MiB cap (`SystemSkill/SkillFormModal.tsx:27,42`). Recommend matching that here so oversized/unsupported files fail fast client-side.

### 🟡 P2 — `window.open(url, '_blank')` on a backend-supplied URL without `noopener`/scheme check

`src/pages/SkillMarket/SkillTab.tsx:103` and `src/pages/SkillMarket/SkillDetailDrawer.tsx:128` open an API-provided download URL directly. Add `'noopener,noreferrer'` and validate the scheme is `http(s):` before opening, to avoid reverse-tabnabbing / `javascript:` execution if a record ever returns a hostile URL.

### 🟡 P2 — Stale-response race in `SkillMarket/SkillTab` list loading

`src/pages/SkillMarket/SkillTab.tsx:45-64` — `load()` for search/category/sort/pagination has no request-generation guard or cancellation. A slow earlier request can resolve after a newer one and overwrite `rows`/`total`/`page`, so the displayed filter can disagree with the displayed data. Track a request id (or abort in-flight) and drop stale responses.

### 🟡 P2 — Detail drawer renders `headers`/`env` values in plaintext (verify no real secrets round-trip)

`src/pages/SystemMcp/DetailDrawer.tsx:210-214,233-237` print header/env values verbatim. The create/probe path is careful (the ephemeral probe bearer is never persisted and never seeded from `editing`, `FormModal.tsx:206-234`), and the code comments state the backend returns `SECRET_PLACEHOLDER` sentinels rather than real secret values. Please confirm server-side that no real credential ever round-trips into `quickStart.headers`/`env` on GET, since this view would display it.

## 3. Verify before merge (human)
- [ ] Marketplace admin surface network isolation + asset auth (P1 token-in-bundle assumptions).
- [ ] Backend does **not** return real secret values in MCP detail `headers`/`env`.
- [ ] Marketplace enforces its own authorization / SSRF egress controls on `/admin/mcps/probe` (the probe forwards an operator-supplied URL + arbitrary headers incl. `Authorization`; `probeHelpers.ts:87-104`) — none of that is verifiable from this frontend PR.

## Verdict

**Changes requested.** No build/correctness breakage (build + 49 tests green), but two P1-level items should be resolved first: (1) the `SystemSkill` mutation controls need `skill.write` gating for parity with `SkillMarket` — or the redundant route removed; and (2) the shared-token-in-bundle model needs explicit human security sign-off on the deployment assumptions. The P2 items (upload validation, `noopener`, list race, secret rendering) are non-blocking cleanups.

lml2468 (dismissed):
## Re-review @ `1f3a0418225d` — 独立本地复核（非转述）

上一轮我在 `33709f62` 给出 REQUEST_CHANGES（2 个 P1 + 2 个 MEDIUM）。本轮针对新增的 `1f3a041 fix(marketplace): remove bundled admin token` 逐条实证复核。**结论:所有阻塞项已修复 → APPROVE。**

### 验证记录（实际执行 @ 新 head）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3216 modules，产物正常
- 全量 `vitest run` — 9 处失败**仍全在 `src/store/feature.test.ts`**（本 PR 未触碰，既有问题，不阻塞）
- **关键回归测试**:`VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED npm run build` → grep 打包产物 → **不再出现**（上一轮此处 CONFIRMED 泄露）。

---

### 上轮阻塞项 —— 逐条核实已修复

**✅ P1① 密钥进 bundle** — 已修复。`mcp.ts` / `skill.ts` 均删除 `VITE_MARKETPLACE_ADMIN_TOKEN` 与 `X-Admin-Token`,改为 `useAuthStore.getState().token` 走 `token` 头(与共享 `api/index.ts` 拦截器一致,marketplace 回调 octo-server 校验 `role=superAdmin`)。`.env.example` 中的 secret 行已删除。实测新 build 不再内联任何 token。对齐了 #115 的最终方案。

**✅ P1② 生产 nginx 缺 `/market/api/` 反代** — 已修复。`nginx.conf.template` 新增对称 `location /market/api/`(`rewrite ^/market/(.*) /$1 break` + 变量式 `proxy_pass`,复用已有 `resolver`);`Dockerfile` 补 `ENV MARKETPLACE_BACKEND=octo-marketplace:8092`;README(中英)补部署说明。

**✅ MEDIUM③ SystemSkill 写门控缺失** — 已修复。`SystemSkill/index.tsx` 计算 `canWrite = hasManagerCapability(..., 'skill.write')` 并下传至 `SkillTable`/`CategoryTab`/`DetailDrawer`/`SkillFormModal`;各写入口(上传/建/改/删)均加 `canWrite` 门控,且 handler 内二次 `if (!canWrite) return` 兜底。与 `/skill-market` 一致。

**✅ MEDIUM④ `mcp.ts`/`skill.ts` 无 401 处理** — 已修复。两个客户端拦截器均加 `401 → logout() + 跳 /admin/login`,镜像共享拦截器。`skill.ts` 还兼容了 `err`/`error` 双 envelope。

---

### Gate 1 规格 ✅ / Gate 2 代码质量 ✅（无阻塞）

对抗性复核未发现新的 🔴/🟠。权限模型现在正确:UI 门控 + 服务端 superAdmin 强制,无共享密钥旁路。

### 🟡 剩余非阻塞(LOW，建议 follow-up，不 gate merge)

1. **nav.json 重复键** — en/zh 的 `nav.json` 均有两个 `systemMcp` 键(第 11、15 行),JSON 后者覆盖:System MCP 侧栏 label 实际显示「MCP Market / MCP 市场」而非「System MCPs / 系统 MCP」。删掉重复的第 11 行或改键名即可。
2. **`/system-skill` 路由无侧栏入口** — `App.tsx` 注册了 `system-skill`(gated `skill.read`),但 `MainLayout` 只 push 了 `/skill-market`,未 push `/system-skill`。若刻意保留请忽略;否则该页仅能直链访问。请确认是否符合预期。
3. **SystemMcp 沿用 #115 修复前旧版**(本 PR 独立分支):`FormModal.handleSubmit` 仍用本地化字符串比较跳步(`err.startsWith('服务标识')`),en-US 下 slug 校验失败会跳错步;缺 bearer sentinel;`systemMcp.json` 的 `form` 内 `sectionTools`/`sectionToolsDesc` 重复键。建议 #115 合入后 rebase 统一,或直接对齐其最终版。
4. **多处硬编码中文**(空态文案 / `+ 新增一条`),en-US 界面显示中文。

### 覆盖与盲区(需人工/后端联调)
- 未跑 marketplace 后端:skill 上传→parse 轮询→create、reupload、category 409、download presigned、以及 401→登出全链路,需真实 marketplace 环境手测。
- 新增写门控为客户端防御纵深;真实边界在服务端 `/admin/*` superAdmin,仍需服务端强制(已按契约设计,未在本仓验证)。

---

**Verdict: APPROVE。** 上一轮两个 P1 与两个 MEDIUM 均已实证修复,权限模型与生产路由回到正确状态。剩余 4 条均为非阻塞 LOW,建议清理但不阻塞合入。感谢快速迭代 👍

Jerry-Xin (dismissed):
The PR is relevant to octo-admin, but the newly exposed System Skill route contains a broken package-reupload workflow.

🔴 Blocking

- 🔴 Critical — Reuploading an existing skill never calls the reupload endpoint. [SkillFormModal.tsx:182](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/pages/SystemSkill/SkillFormModal.tsx:182) sends `parse_task_id`, version, and changelog through the ordinary PATCH API. However, `updateSkill` aliases `updateAdminSkill`, whose contract only updates metadata; the implemented package replacement endpoint is `POST /admin/skills/:id/reupload` in [skill.ts:456](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/api/skill.ts:456). Consequently, the UI can report success without replacing the archive, or the server can reject the unsupported fields. Use the reupload endpoint after parsing, then perform any remaining metadata update separately.

💬 Non-blocking

- 🟡 Warning — Skill-table action clicks bubble into the row click handler. Buttons such as download, edit, and delete at [SkillTab.tsx:202](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/pages/SkillMarket/SkillTab.tsx:202) also trigger the detail drawer registered at [SkillTab.tsx:283](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/pages/SkillMarket/SkillTab.tsx:283). Stop propagation on the actions container.
- 🟡 Warning — Category reorder controls only rearrange local state at [CategoryTab.tsx:87](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/pages/SystemSkill/CategoryTab.tsx:87); refreshing loses the order. Persist updated `sort_order` values or remove the controls.
- 🟡 Warning — Both navigation locale files contain duplicate `systemMcp` keys, so the first translation is silently overwritten: [en-US/nav.json:11](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/i18n/locales/en-US/nav.json:11) and [zh-CN/nav.json:11](/Users/mlamp/.openclaw/tmp/pr-review-zfd40tpg/repo/src/i18n/locales/zh-CN/nav.json:11).
- 🔵 Suggestion — Add API/component tests for skill creation, reupload, category mutation, and action propagation. Current new tests only cover MCP probe helpers.

✅ Highlights

- Production build and TypeScript compilation pass.
- All 14 new MCP probe-helper tests pass.
- Marketplace authentication remains tied to the current user token rather than exposing service credentials.
- The full suite reached 40 passing tests; nine unrelated feature-store tests failed because `localStorage` was unavailable under the local Node 26 runtime.

Jerry-Xin (changes requested):
The two authorization blockers from the previous review are now byte-verified fixed. However, this revision exposes a broken System Skill reupload workflow that blocks merge.

## 🔴 Blocking

- 🔴 **Critical — Reuploading a skill package never commits the replacement.** In `src/pages/SystemSkill/SkillFormModal.tsx`, the edit path correctly uploads + parses via `initReupload` → presigned PUT → `triggerParse` (produces a `parse_task_id`). But on submit (`SkillFormModal.tsx:182`), for an existing skill it calls `updateSkill(id, { parse_task_id, name, version, changelog, ... })`, and `updateSkill` is aliased to `updateAdminSkill` (`src/api/skill.ts:501`) → `PATCH /admin/skills/:id`. The PATCH contract type `PatchSkillParams` (`src/api/skill.ts:130-140`) is **metadata-only** and has no `parse_task_id` field — so the parsed archive is never committed. The dedicated `reuploadAdminSkill` → `POST /admin/skills/:id/reupload` (`src/api/skill.ts:457-476`), the only call that actually replaces the package using `parse_task_id`, is never invoked from the submit path (only its `initReupload` half is used). Net: an admin reuploads a new version, sees a success toast, but the package/version is not replaced (or the server rejects the unsupported field). Fix: after parse succeeds on an edit, call `reuploadAdminSkill` (POST `/reupload`) to commit the package, then do any remaining metadata PATCH separately.

## 💬 Non-blocking

- 🟡 **Duplicate `systemMcp` key in both nav locale files.** `src/i18n/locales/zh-CN/nav.json:11` ("系统 MCP") is silently overwritten by line 15 ("MCP 市场"); same in `en-US/nav.json:11`. The first, intended label is dead. Deduplicate (this was flagged in the prior review and is still unresolved).
- 🟡 **Skill-table action clicks bubble into the row/detail-drawer click handler** (`src/pages/SkillMarket/SkillTab.tsx:202` vs the row handler at `:283`). Stop propagation on the actions container so download/edit/delete don't also open the drawer.
- 🟡 **Category reorder only mutates local state** (`src/pages/SystemSkill/CategoryTab.tsx:87`); order is lost on refresh. Persist `sort_order` server-side or remove the controls.
- 🔵 **Tests:** new coverage only exercises MCP probe helpers. Consider adding tests for skill create/reupload, category mutation, and action-click propagation.

## ✅ Highlights

- **P1-① fixed:** the marketplace admin token is gone from the browser bundle. `src/api/mcp.ts` and `src/api/skill.ts` no longer read any `VITE_*_ADMIN_TOKEN`; both now attach the caller's own Octo login token (`token` header from the auth store) and rely on server-side `superAdmin` enforcement. `.env.example` no longer defines an admin-token var.
- **P1-② fixed:** `nginx.conf.template` now has a `/market/api/` reverse-proxy block symmetric to `/api/`, using a variable `proxy_pass http://$market_backend` (`${MARKETPLACE_BACKEND}`) with the `resolver 127.0.0.11` in place for variable-based proxy_pass, and the `/market/` prefix rewrite so requests land at the marketplace as `/api/v1/*`. Prod 502 resolved.
- **SystemSkill write-gating** is comprehensive: `canWrite` (skill.write capability) gates every mutating action across `index.tsx`, `CategoryTab`, `SkillTable`, `SkillFormModal`, and `DetailDrawer` — matching SkillMarket.
- 401 handling in both marketplace clients mirrors the shared interceptor (logout + redirect to /admin/login).
- Production build and TypeScript compilation pass.

Deploy note: this frontend requires the companion octo-marketplace service reachable at `${MARKETPLACE_BACKEND}` and the backend env wired for the per-user token verification path; ship them together.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #116 (octo-admin)
**Head:** `1f3a0418225da11482cc2124e73610a2b92b30e1`
**Scope:** feat: add MCP, skill market, and system skill admin (~4500 lines across 24 files)

## Summary

This PR adds three new admin surfaces to octo-admin: System MCP management (CRUD + probe), Skill Market admin (categories + skills CRUD with presigned-URL upload), and a legacy-compatible System Skill page. It introduces a separate axios client (`mcpApi` / `skillApi`) targeting the octo-marketplace backend via `/market/api/v1/*`, with matching nginx and vite dev proxy rewrites. Authorization is correctly gated via new `mcp.read/write` and `skill.read/write` manager capabilities. i18n covers both en-US and zh-CN for all new pages.

Overall the PR is well-structured with solid patterns (capability-gated routes, presigned URL upload flows, probe helper unit tests). No P0/P1 blockers found. Several P2 items and nits below.

## Verification

Static analysis only at head `1f3a041`; build and tests not executed in this environment.

- ✅ **Capability gating** — `mcp.read`/`skill.read` correctly guard routes in `src/App.tsx:183-210`; `mcp.write`/`skill.write` correctly gate write actions in all page components. `src/auth/capabilities.ts` properly extends `MANAGER_CAPABILITY_KEYS` and `firstManagerPath`.
- ✅ **Auth interceptor** — Both `mcpApi` and `skillApi` read token from `useAuthStore.getState().token` in request interceptors, consistent with the existing `api` client pattern. Error interceptors normalize both `{err:{...}}` and `{error:{...}}` envelopes.
- ✅ **Nginx ↔ Vite proxy parity** — `nginx.conf.template` rewrites `^/market/(.*) /$1 break` stripping `/market`; `vite.config.ts` rewrites `path.replace(/^\/market/, '')`. Both forward to the marketplace backend. Consistent.
- ✅ **Dockerfile** — Multi-stage build is correct; `MARKETPLACE_BACKEND` env var is declared with sensible default `octo-marketplace:8092`.
- ✅ **Probe helpers** — `probeHelpers.ts` correctly omits `authType` from wire payload (backend `DisallowUnknownFields`), ephemeral bearer token overrides `headers.Authorization` without persisting. Unit tests in `probeHelpers.test.ts` cover all edge cases including stdio-returns-null, empty-URL, header parsing, and i18n error resolution.
- ✅ **Presigned URL upload flows** — Both MCP icon upload (`src/api/mcp.ts:256-304`) and skill zip upload (`src/api/skill.ts:340-470`) follow the correct init→PUT→trigger→poll pattern with proper cleanup.
- ✅ **Stale response guards** — `SystemMcp/DetailDrawer.tsx` and `SystemSkill/DetailDrawer.tsx` both use `cancelled`/`stale` flags with cleanup on unmount.

## Findings

No P0/P1 issues. Four P2 items and two nits below.

### P2 — Hardcoded Chinese strings bypass i18n (`src/pages/SystemMcp/FormModal.tsx:912-960`)

The `DynamicListHeader`, `SimpleTextList`, and `FaqList` helper components at the bottom of FormModal.tsx contain hardcoded Chinese text that is rendered regardless of the active locale:

```ts
// line ~912 — DynamicListHeader
+ 新增一条

// line ~940 — SimpleTextList empty state
`还没有内容，点击右上角「${addLabel}」添加`
```

The locale files already provide proper translations (`form.addOne`, `form.exampleAdd`, etc.) and the main form body correctly uses `t('form.addOne')` for the same action. These helpers should accept an `addLabel` parameter (which they partially do but don't use for the button text) or call `t()` directly.

### P2 — Category reorder is never persisted (`src/pages/SystemSkill/CategoryTab.tsx:104-110`)

The `handleMove` function swaps items in local state only — there is no API call to persist the new order. The `CategoryItem` type has a `sort_order` field, but it is never updated server-side when the user clicks the up/down arrows. The reordered state is lost on page refresh, and the UI gives no indication the change is ephemeral. Either wire it to a batch-reorder endpoint or remove the arrows.

### P2 — Race condition in SkillDetailDrawer (`src/pages/SkillMarket/SkillDetailDrawer.tsx:24-46`)

The `useEffect` that fetches skill detail and SKILL.md content has no cleanup/stale-response guard. If a user opens drawer A, closes it, and quickly opens drawer B, a slower response for A could overwrite B's data. Compare with `SystemSkill/DetailDrawer.tsx:35-47` which correctly uses a `stale` flag with cleanup. Add the same pattern:

```ts
let stale = false
// ... fetch ...
return () => { stale = true }
```

### P2 — Hidden SystemSkill route with no sidebar entry (`src/App.tsx:203-210`)

Route `/system-skill` is registered and gated by `skill.read`, but `MainLayout.tsx` only adds `/skill-market` to the sidebar — not `/system-skill`. The SystemSkill page is reachable only by direct URL. This creates two parallel admin surfaces for the same data (marketplace categories are the same entities as system-skill categories, both hitting `/admin/skill_categories`). If SystemSkill is meant as a legacy fallback, consider documenting it or gating it behind a feature flag; if it's deprecated, consider removing it.

### Nit — Pagination labels reuse upload i18n keys (`src/pages/SystemSkill/SkillTable.tsx:177-182`)

The prev/next pagination buttons use `t('upload.prev')` and `t('upload.next')` — keys semantically intended for the upload wizard's step navigation. While the strings happen to read correctly, dedicated keys (e.g., `list.prev`/`list.next`) would be clearer and less fragile.

### Nit — `DynamicListHeader` button text doesn't use the `addLabel` prop (`src/pages/SystemMcp/FormModal.tsx:906-920`)

`DynamicListHeader` accepts an `onAdd` callback but hardcodes the button text as `+ 新增一条` instead of using the `addLabel` prop that callers (`SimpleTextList`, `FaqList`) already compute from i18n. Thread `addLabel` through and render it in the button.

- **Data flow: skill list → SkillEditModal** — `SkillTab` passes `SkillListItem` from `listAdminSkills` to `SkillEditModal`; `normalizeSkill` ensures `skill_id`, `description`, `tags` are populated. `updateAdminSkill(skill.skill_id, ...)` receives the correct ID.
- **Data flow: category list → filter dropdowns** — Both `SkillTab` and `SystemSkill/SkillTable` call `listSkillCategories()` (or its alias `listCategories`) and populate `<Select>` options using `c.skill_category_id`. Category filter correctly resets pagination.
- **Icon upload size/type validation** — `FormModal.tsx` validates `MAX_ICON_BYTES=2MB` and `ALLOWED_ICON_TYPES={png,jpeg,webp,gif}` before upload. `SkillFormModal.tsx` validates `MAX_SKILL_PACKAGE_SIZE=20MB` and `.zip/.skill` extension.
- **Bearer token never persisted** — `FormModal.tsx` resets `probeBearer` on modal open; `buildProbeRequest` puts it on `headers.Authorization` only for the probe call. The saved config keeps `SECRET_PLACEHOLDER`.
- **Last-row-on-page deletion** — `SystemMcp/index.tsx:handleDeleted` correctly handles the edge case where deleting the last item on a page decrements the page number.
- **i18n completeness** — New locale files (`systemMcp.json`, `skillMarket.json`, `systemSkill.json`) are present for both en-US and zh-CN with matching key structures. Nav keys added.

## Verdict: APPROVED

No P1 blockers. The PR delivers three well-gated admin surfaces with proper auth, i18n, and upload flows. The four P2 items (hardcoded Chinese in helper components, non-persisted category reorder, missing stale-response guard, hidden duplicate route) are non-blocking quality improvements. The race condition P2 in SkillDetailDrawer is the most actionable — a one-line fix mirroring the pattern already used in SystemSkill/DetailDrawer.

[Octo-Q] verdict: APPROVE — No P0/P1 found. Four P2 items (hardcoded i18n strings, ephemeral category reorder, missing stale guard in SkillDetailDrawer, hidden duplicate route) are non-blocking. Recommend addressing the stale-response guard before merge.

yujiawei (approved):
# Code Review — PR #116 (octo-admin)

Reviewed at head `1f3a041` against merge-base `3e42799`. Build (`tsc && vite build`) is green and all 49 vitest cases pass (including the 14 probe-helper tests). This PR was flagged security-sensitive, so the review focused on secret handling, auth/token flow, the MCP probe, the reverse-proxy config, and the presigned upload/download paths.

## 1. Spec compliance

**Spec: ✅**

The diff delivers exactly what the description lists — System MCP admin, Skill Market, System Skill admin, plus nav/permissions/i18n/upload-download/probe/dev-proxy wiring. Nothing out-of-scope. One structural note (not a spec violation): two skill-admin surfaces now coexist — `SkillMarket/` (wired into the nav) and `SystemSkill/` (route exists in `App.tsx` but is **not** in `MainLayout.tsx`). Both talk to the same `/admin/skills` backend. See the maintainability note below.

The headline change in the top commit — removing the build-time `VITE_MARKETPLACE_ADMIN_TOKEN` shared secret and switching to the operator's own Octo login token — is the right call and is implemented cleanly:

- No residual `ADMIN_TOKEN` / `X-Admin-Token` / `dev-admin-token` anywhere in `src/` or `.env.example`.
- `mcp.ts` / `skill.ts` now send the same `token` header the shared axios instance uses; marketplace verifies it and enforces `role=superAdmin` server-side, so no marketplace credential is baked into the shipped bundle.
- The ephemeral probe bearer (`probeBearer`) is correctly kept out of the persisted create/update payload (`FormModal.buildPayload`, verified — it is only ever attached to the transient probe request).
- Presigned PUTs use raw `fetch` rather than the axios instance, so the `token` / `Accept-Language` interceptors don't leak into third-party storage hosts.

No `dangerouslySetInnerHTML` / markdown-render sinks were introduced. External `window.open(url, '_blank')` calls use only server-issued presigned download URLs.

## 2. Code Quality

**Quality: Approved**

No P0/P1 found. The items below are advisory (P2) or things a human should confirm against the marketplace backend — none block merge.

### For human verification (backend-dependent, cannot confirm from this diff)

- **MCP probe is a server-side SSRF surface.** `probeSystemMcp` (`src/api/mcp.ts:252`) forwards an operator-supplied URL + arbitrary headers to `POST /admin/mcps/probe`. The frontend correctly cannot and does not attempt SSRF validation — but please confirm the marketplace probe handler blocks loopback / link-local / RFC1918 / `169.254.169.254` and non-http(s) schemes, and pins/validates redirect hops. This is an existing backend concern surfaced by the new UI, not a defect in this PR.
- **Presigned-upload content-type.** The new upload paths (`uploadAndParseSkillZip` at `src/api/skill.ts:423`, `uploadMcpIcon` at `src/api/mcp.ts:295`) PUT with `headers ?? {}`, letting the browser infer `Content-Type` from the `File` when the backend returns no header. The legacy `uploadToPresigned` (`skill.ts:526`) instead forces `application/octet-stream`. If the marketplace signs the presigned PUT *with* a fixed content-type in `SignedHeaders`, browser inference will cause a `403 SignatureDoesNotMatch`. Worth a quick confirm that skill-zip presigns don't sign content-type (icon uploads already send `content_type` on init, so those are fine).

### P2 — hardening / defense-in-depth

- **`VITE_MARKETPLACE_API_BASE` accepts an absolute cross-origin URL** (`mcp.ts:26`, `skill.ts:22`) while the request interceptor unconditionally attaches the user's token. A misconfigured/poisoned build (`https://evil/api/v1`) would send admin tokens off-origin. Consider rejecting absolute bases or pinning the same-origin `/market/api/v1` path. (Build-time trusted input, hence P2.)
- **nginx `/market/api/` location is broader than needed** (`nginx.conf.template:60`). It proxies and strips the whole `/market/api/*` namespace, not just `/v1/*`, so any future/undocumented marketplace route becomes reachable through the admin origin. Recommend scoping to `/market/api/v1/`. (Each routed endpoint must still authenticate independently — this is depth, not the boundary.)

### P2 — correctness / maintainability

- **Duplicate JSON keys in `nav.json`** (both `en-US` and `zh-CN`): `systemMcp` appears twice (`"System MCPs"` then `"MCP Market"`), and `skillMarket` / `systemSkill` overlap. JSON last-wins silently resolves it, so the nav renders `"MCP Market"` rather than the first value — not a crash, but it's an accidental defect and a lint smell. Please de-duplicate and keep one intended label per key.
- **Two coexisting skill-admin surfaces.** `SystemSkill/` is routed in `App.tsx` (`/system-skill`, capability-gated) but has no nav entry, so it's reachable only by direct URL while `SkillMarket/` is the linked one. If `SystemSkill/` is superseded, consider removing it to avoid two code paths drifting against the same backend.
- **Module-global `parseTaskByUploadId` Map** (`skill.ts:214`) is never cleared and is only used by the legacy `createSkill` compat path; a page refresh between parse and create loses the mapping. Minor (admin-only, small strings). The newer `SkillUploadModal` path threads `parseTaskId` explicitly and is unaffected.
- **Header-key case collision** (`probeHelpers.ts:54,97`): `parseHeaderBlock` preserves case, so a user-typed lowercase `authorization` plus the ephemeral bearer yields both `authorization` and `Authorization` in the JSON map. Harmless today but worth normalizing to avoid ambiguous forwarding.
- **Raw `fetch` rejections are `TypeError`, not `ApiError`** in the presigned PUT paths, so consumers pattern-matching on `ApiError.status`/`code` won't get structured info on network/CORS failures. Minor UX.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The prior blocker (shared admin token compiled into the browser bundle) is properly removed and replaced with the operator's own superAdmin-scoped Octo token. Remaining items are advisory; please skim the two human-verification points on the marketplace side before shipping to production.

## 4. Coverage note

This is a frontend-only review. I could not verify from this diff: marketplace-side `superAdmin` enforcement on `/admin/*`, probe SSRF/redirect defenses, presigned-URL signing (content-type/length constraints) and ownership checks, or server-side dedup/validation of uploaded skill archives. Two independent advisory passes were run: one line-level pass and one whole-PR pass. The whole-PR pass raised a "missing nginx resolver" concern that I refuted — `resolver 127.0.0.11 valid=10s;` is present at `nginx.conf.template:7` (the reviewer saw only the diff hunk, not the full file).

yujiawei (changes requested):
# Code Review — PR #116 (octo-admin)

**Updating my earlier verdict.** After a closer trace of the System Skill reupload path I found a confirmed functional blocker I missed on the first pass, so this supersedes my prior approval. Reviewed at head `1f3a041` against merge-base `3e42799`; `tsc && vite build` is green and all 49 vitest cases pass — but the green build actually hides the blocker (see below).

## 1. Spec compliance

**Spec: ✅**

The diff delivers what the description lists (System MCP admin, Skill Market, System Skill admin, plus nav/permissions/i18n/upload-download/probe/dev-proxy wiring); nothing out of scope. The headline change — removing the build-time `VITE_MARKETPLACE_ADMIN_TOKEN` shared secret and switching to the operator's own Octo login token — is correct and complete: no residual `ADMIN_TOKEN` / `X-Admin-Token` / `dev-admin-token` anywhere in `src/` or `.env.example`; both marketplace clients now send the same `token` header as the shared axios instance and rely on server-side `role=superAdmin`; the ephemeral `probeBearer` is kept out of the persisted payload; presigned PUTs use raw `fetch` so interceptors don't leak into storage hosts. No new HTML/markdown XSS sinks.

## 2. Code Quality

### 🔴 P1 — System Skill reupload never commits the new package

`src/pages/SystemSkill/SkillFormModal.tsx:181-195` (edit branch of `handleSubmit`): the reupload UI correctly runs `initReupload` → presigned PUT → `triggerParse`, producing a `parse_task_id`. But on submit it calls `updateSkill(activeEditSkill.id, { parse_task_id, name, version, changelog, ... })`, and `updateSkill` is aliased to `updateAdminSkill` (`src/api/skill.ts:501`) → **`PATCH /admin/skills/:id`**. The PATCH contract `PatchSkillParams` (`src/api/skill.ts:130-140`) is **metadata-only and has no `parse_task_id` field** — so the parsed archive is never committed. The only call that actually replaces a package, `reuploadAdminSkill` → `POST /admin/skills/:id/reupload` (`src/api/skill.ts:457-474`), is never invoked from any submit path — only its `initReupload` half is used.

Failure scenario: an admin opens System Skill → edits a skill → uploads a new `.zip`, parsing succeeds, submits, and sees the success toast — but the package and version are not replaced (the unsupported field is dropped server-side, or the request is rejected). A silent no-op on a data-mutating admin action.

Why the build didn't catch it: `parse_task_id` is passed inside a conditional spread (`...(parseTaskId ? {...} : {})`), and TypeScript excess-property checking does not fire through a spread — so `tsc` stays green even though the field isn't in `PatchSkillParams`.

Fix: after parse succeeds on an edit, call `reuploadAdminSkill` (POST `/reupload`) to commit the package, then issue any remaining metadata PATCH separately (or add `parse_task_id` to the reupload contract and route the edit-with-new-file case through it).

Scope note: this defect is confined to the `SystemSkill/` page, which is routed at `/system-skill` (capability-gated) but **not** linked from the sidebar (`MainLayout.tsx` only pushes `/skill-market`), so it is reachable by direct URL only. The nav-linked `SkillMarket/` surface uploads new skills via `createAdminSkill` (works) and edits metadata via `updateAdminSkill` (works) and has no package-reupload flow. The reupload feature is only in the unlinked page and is broken there.

### For human verification (backend-dependent)

- **MCP probe SSRF.** `probeSystemMcp` (`src/api/mcp.ts:252`) forwards an operator-supplied URL + arbitrary headers to `POST /admin/mcps/probe`. The frontend cannot and should not validate this; please confirm the marketplace probe handler blocks loopback / link-local / RFC1918 / `169.254.169.254` and non-http(s) schemes and validates redirect hops.
- **Presigned-upload content-type.** The new upload paths (`uploadAndParseSkillZip` at `skill.ts:423`, `uploadMcpIcon` at `mcp.ts:295`) PUT with `headers ?? {}`, letting the browser infer `Content-Type` from the `File` when the backend returns none; the legacy `uploadToPresigned` (`skill.ts:526`) forces `application/octet-stream`. If the marketplace signs the PUT with a fixed content-type, browser inference yields `403 SignatureDoesNotMatch`. Confirm skill-zip presigns don't sign content-type (icon init already sends `content_type`, so those are fine).

### P2 — non-blocking

- **`VITE_MARKETPLACE_API_BASE` accepts an absolute cross-origin URL** (`mcp.ts:26`, `skill.ts:22`) while the interceptor unconditionally attaches the user token; a poisoned build (`https://evil/api/v1`) exfiltrates admin tokens off-origin. Reject absolute bases or pin the same-origin path. (Build-time trusted input.)
- **nginx `/market/api/` is broader than needed** (`nginx.conf.template:60`) — it strips/proxies the whole `/market/api/*` namespace, not just `/v1/*`. Scope to `/market/api/v1/`. (Each endpoint must still auth independently.)
- **Duplicate JSON keys in `nav.json`** (en-US and zh-CN): `systemMcp` appears twice (line 11 then 15) so the nav silently renders the second label; `skillMarket`/`systemSkill` also overlap. De-duplicate.
- **`SkillMarket/SkillDetailDrawer.tsx:24-46`** fetch effect has no stale-response guard (unlike `SystemSkill/DetailDrawer.tsx`), so quickly reopening drawers can render stale data. One-line `stale` flag fixes it.
- **Action clicks bubble into the row/detail handler** in `SkillMarket/SkillTab.tsx` (download/edit/delete also open the drawer) — stop propagation on the action container.
- **Category reorder is local-only** (`SystemSkill/CategoryTab.tsx:87`, `handleMove`): `sort_order` is never persisted, so order is lost on refresh. Persist or remove the arrows.
- **Module-global `parseTaskByUploadId` Map** (`skill.ts:214`) is never cleared and breaks on refresh between parse and create; used only by the legacy `createSkill` compat path.
- **Header-key case collision** (`probeHelpers.ts:54,97`): a user-typed lowercase `authorization` plus the ephemeral bearer yields both `authorization` and `Authorization` in the JSON map. Normalize.
- **Hardcoded Chinese strings** in `SystemMcp/FormModal.tsx` helper components (`+ 新增一条`, empty-state text) render regardless of locale though i18n keys exist.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅ but Quality has a confirmed P1 (System Skill reupload silently fails to commit the package). The security posture is otherwise solid — the prior shared-token-in-bundle blocker is properly fixed. Once the reupload path is routed through `POST /reupload`, this is close to mergeable; the remaining items are non-blocking.

## 4. Coverage note

Frontend-only review. Not verifiable from this diff: marketplace-side `superAdmin` enforcement, probe SSRF/redirect defenses, presigned-URL signing constraints and ownership checks, and server-side handling of the `parse_task_id` field on PATCH vs POST /reupload (which determines whether the P1 above is a silent no-op or a hard 4xx — either way the reupload does not work as designed). Two independent advisory passes were run (one line-level, one whole-PR); the whole-PR pass raised a "missing nginx resolver" concern that I refuted — `resolver 127.0.0.11 valid=10s;` is present at `nginx.conf.template:7`.

lml2468 (dismissed):
## Re-review @ `537e17a0f277` — 独立本地复核（非转述）

上一轮我在 `1f3a0418` 已 APPROVE(2 个 P1 + 2 个 MEDIUM 修复确认)。此后 peers 在同一 head 发现一个**新 P1(SystemSkill 重传包只 PATCH 元数据、没真正换包)**。本轮针对新 commit `537e17a0 fix: commit system skill reuploads` 实证复核。**结论:该 P1 已修复 → 维持 APPROVE。**

### 验证记录（实际执行 @ 新 head）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3216 modules
- token 泄露回归:`VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED npm run build` → 打包产物 grep **无命中**(P1 保持已修)
- 全量 `vitest run` — **41 passed**(较上轮 +1,新增 `skill.test.ts` 通过);9 处失败仍全在未触碰的 `src/store/feature.test.ts`(既有,不阻塞)

---

### 本轮 P1 —— 核实已修复 ✅

**SystemSkill 重传技能包** — 已修复,证据充分:
- 新增 `commitAdminSkillReupload(skillId, {parse_task_id, version, changelog, tags})` → `POST /admin/skills/:id/reupload`(`skill.ts:306`)。
- `SkillFormModal.handleSubmit`(编辑态)现在**先**调 `commitAdminSkillReupload`(当 `parseTaskId` 存在,即真正提交新包/版本),**再**调 `updateSkill` 做元数据 PATCH。两步顺序 await,reupload 失败会中断、不会走到 PATCH。
- PATCH 不再夹带 `parse_task_id`(此前 `PatchSkillParams` 根本不收该字段,是空操作);version 由 reupload 落库,后续 PATCH 省略 version → 服务端保留 reupload 的版本,无覆盖。
- `reuploadAdminSkill` 也重构为复用同一函数,消除重复。
- **新增单测** `src/api/skill.test.ts`:断言 `commitAdminSkillReupload` 打到 `/admin/skills/skill-1/reupload` 且 body 正确 — 本地实跑 ✅ 通过。

之前所有阻塞项(密钥进 bundle / nginx `/market` 反代 / SystemSkill 写门控 / 401 处理)保持已修复,无回归。

---

### Gate 1 ✅ / Gate 2 ✅（无阻塞）

对抗性复核未发现新的 🔴/🟠。reupload 修复的顺序、错误传播、版本一致性均正确。

### 🟡 剩余非阻塞(建议 follow-up，不 gate merge)

1. **[MEDIUM] SystemSkill 分类 ↑/↓ 排序不持久化** — `CategoryTab.handleMove`(`SystemSkill/CategoryTab.tsx:87`)只 `setData(next)` 改本地 state,不调后端;且兼容层 `updateCategory`(`skill.ts:599`)**丢弃 `sort_order`**(只传 `name`)。管理员点 ↑/↓ 看到顺序变化,刷新后复原 —— 属"看似生效实则丢失"的误导性交互。修法:move 后调用带 `sort_order` 的 `updateSkillCategory` 持久化(SkillMarket 侧的 CategoryTab 走表单 `sort_order` 是持久化的,可对齐)。peer(Jerry-Xin/mochashanyao)亦提及。
2. **[LOW] nav.json 重复键**(carryover,未修)— en/zh 的 `nav.json` 各有两个 `systemMcp`(第 11、15 行),后者覆盖:System MCP 侧栏 label 实际显示「MCP Market / MCP 市场」。删第 11 行或改键名即可。
3. **[LOW] SystemMcp 沿用 #115 修复前旧版** — `FormModal` 本地化字符串比较跳步(en-US 下 slug 错步)、缺 bearer sentinel、`systemMcp.json` `form` 内 `sectionTools*` 重复键。建议 #115 合入后 rebase 统一。
4. **[LOW] 多处硬编码中文**,en-US 界面显示中文。

### 覆盖与盲区(需人工/后端联调)
- 未跑 marketplace 后端:reupload(init→parse→commit)、create、category CRUD、download presigned、401→登出全链路,需真实环境手测。
- probe 的 SSRF 防护(内网/非 http(s) 拦截)在服务端,属后端职责,本仓不可验证 —— peer 提的合理,建议后端确认。
- 部署:需与 octo-marketplace 伴随服务(可达 `${MARKETPLACE_BACKEND}`)+ per-user token 校验链路一同上线。

---

**Verdict: APPROVE。** 本轮新 P1(重传不换包)已实证修复并配单测,权限/生产路由/密钥各项保持正确。剩余为 1 条 MEDIUM(分类排序不持久化)+ 3 条 LOW,均非阻塞,建议清理。感谢持续快速迭代 👍

Jerry-Xin (dismissed):
Re-review at head `537e17a0`. The blocking issue from the previous review — the System Skill package re-upload not calling the re-upload endpoint — is now correctly fixed. Approving.

## ✅ Resolved (was the prior 🔴 blocker)

- **Skill re-upload now replaces the package.** `SkillFormModal.handleSubmit` (`src/pages/SystemSkill/SkillFormModal.tsx:181`) now calls `commitAdminSkillReupload(activeEditSkill.id, { parse_task_id, version, changelog, tags })` when a new package has been parsed, before the metadata `updateSkill` call. Previously the `parse_task_id` was passed through the metadata PATCH (`updateAdminSkill`), which only updated metadata and never actually swapped the package.
- New `commitAdminSkillReupload` (`src/api/skill.ts:306`) POSTs to `/admin/skills/{id}/reupload`, and `reuploadAdminSkill` was refactored to reuse it — single source of truth for the re-upload contract.
- New unit test `src/api/skill.test.ts` verifies the POST target and payload; it passes locally.

## ✅ Prior P1s confirmed still fixed (byte-verified, not regressed)

- **No marketplace secret in the browser bundle.** `nginx.conf.template` is byte-identical to the previously-approved head (`/market/api/` proxy block intact). No `VITE_MARKETPLACE_ADMIN_TOKEN` / `X-Admin-Token` anywhere in source. Regression build with a sentinel value (`VITE_MARKETPLACE_ADMIN_TOKEN=<sentinel> vite build`) produces a `dist/` bundle with **zero** occurrences of the sentinel. The marketplace client authenticates with the caller's own Octo login token at runtime (`src/api/skill.ts:29`), not a build-time secret.
- `src/api/mcp.ts` is byte-identical to the approved head — untouched.

## 💬 Non-blocking notes (pre-existing / low)

- Both `nav.json` locale files still carry duplicate `systemMcp` keys (`src/i18n/locales/en-US/nav.json:11`, `src/i18n/locales/zh-CN/nav.json:11`); the earlier translation is silently overwritten. Worth a cleanup.
- The `/system-skill` route (`src/App.tsx:248`) has no navigation entry and largely duplicates the Skill Market admin flow; consider consolidating to avoid divergence.
- System Skill parse polling uses an async `setInterval` (`src/pages/SystemSkill/SkillFormModal.tsx:127`); requests can overlap if a response exceeds the interval. Sequential `setTimeout` with an attempt/timeout cap would be safer.
- The Skill Market download uses `window.open()` after an awaited request (`src/pages/SkillMarket/SkillTab.tsx:105`); some browsers may block the popup since it's outside the original click gesture. Note this file is unchanged from the previously-approved head, so it is not a regression from this delta — flagging for a future cleanup (e.g. a synchronously-created `<a>` element).

## ✅ Highlights

- Marketplace credentials are not embedded in the frontend.
- Write controls are consistently capability-gated.
- Re-upload contract is now centralized and unit-tested.
- TypeScript compilation and production build succeed.

yujiawei (dismissed):
# Code Review — PR #116 (octo-admin)

Re-reviewed at head `537e17a` against merge-base `3e42799`. This is the third review pass; the two prior blocking findings (bundled marketplace admin token, and the System Skill reupload that never committed the replacement package) both have dedicated fix commits in this revision. All verification below was run independently against a fresh checkout of the head SHA.

**Verification (actually executed at head `537e17a`):**
- `pnpm install --frozen-lockfile` — ✅
- `tsc && vite build` — ✅ 3234 modules, bundle produced
- `vitest run` — ✅ 50/50 tests pass (incl. 14 probe-helper tests and the new `skill.test.ts`)
- Regression on the earlier token leak: `VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED_XYZ pnpm build` then grep the `dist/` bundle → **0 matches** (sentinel not present; the leak is gone)

## 1. Spec compliance

**Spec: ✅**

The diff delivers exactly what the PR description lists — System MCP admin (CRUD + live probe), Skill Market admin (categories + public skills with presigned upload/download), and System Skill admin — plus navigation, i18n (en/zh), capability wiring, and the dev/nginx `/market` proxy. No functionality beyond the stated scope; no requested piece missing.

## 2. Code quality

**Quality: Approved**

Prior blockers confirmed fixed:

- **Marketplace admin token no longer bundled.** Both `src/api/mcp.ts` and `src/api/skill.ts` now authenticate with the caller's own login token (`useAuthStore.getState().token` on the request interceptor), and there is no `VITE_MARKETPLACE_ADMIN_TOKEN` reference anywhere in `src/`, `.env.example`, `vite.config.ts`, or `nginx.conf.template`. Byte-verified gone from the production bundle (see regression above). Marketplace is documented to resolve the login token server-side and admit only `superAdmin` on `/admin/*`.

- **System Skill reupload now commits the replacement package.** `src/pages/SystemSkill/SkillFormModal.tsx:184` now calls the new `commitAdminSkillReupload(activeEditSkill.id, { parse_task_id, version, changelog, tags })` against `POST /admin/skills/:id/reupload` before the metadata `PATCH`, so an operator replacing a package actually persists the new bytes. The metadata `updateSkill` that follows sends only name/display_name/description/category/tags/visibility/icon — it does not carry `version`, so it cannot clobber the version the reupload just set. The extracted `commitAdminSkillReupload` is shared by both the modal and the legacy `reuploadAdminSkill` flow, and is covered by the new `src/api/skill.test.ts`.

Security-sensitive surfaces reviewed (this PR was flagged security-sensitive):

- **Capability gating is now consistent across all three new surfaces.** Routes are guarded by `CapabilityRoute` (`mcp.read` / `skill.read`) in `src/App.tsx`, and every write action is gated by `canWrite` derived from `hasManagerCapability(..., 'mcp.write' | 'skill.write')` — including `SystemSkill/index.tsx`, which was missing the gate in an earlier revision and now matches the SkillMarket/SystemMcp pattern (upload, edit, delete, category writes all gated). Note these are UI/UX gates; the authoritative enforcement is the marketplace `superAdmin` check server-side, which is correct.
- **MCP probe secret handling is sound.** The "试连密钥（不保存）" ephemeral bearer (`probeBearer`) is passed only into `buildProbeRequest` for the live probe and is never seeded from the editing record nor included in `buildPayload()`, so it is not persisted. `buildProbeRequest` correctly drops `authType` from the probe wire (the backend decodes with `DisallowUnknownFields`) while still delivering the token via the `Authorization` header.
- **Presigned upload/download** use short-lived server-issued URLs; `window.open(url, '_blank')` for downloads is standard. No credentials are embedded in these URLs by the client.
- **Reverse proxy** (`/market/api` → marketplace) strips the `/market` prefix consistently in dev (vite) and prod (nginx); no secrets in the config.

### Non-blocking notes (nits / suggestions — do not block merge)

- `dist/assets/index-*.js` is ~1.8 MB (562 kB gzip) and trips vite's 500 kB chunk warning. Not introduced by this PR alone, but the three new admin surfaces add to it; consider route-level `React.lazy` code-splitting for these admin pages in a follow-up.
- `probeHelpers.ts#parseHeaderBlock` is a deliberate duplicate of `FormModal.parseKV`; the comment already flags that the two copies must stay in sync — fine for now, but a shared util would remove the drift risk.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. Both previously-blocking findings are fixed and independently verified at head `537e17a` (green build, 50/50 tests, token leak byte-verified gone). Remaining items are non-blocking nits.

lml2468 (approved):
## Re-review @ `7cc0a867ce41` — 独立本地复核（非转述）

上一轮我在 `537e17a0` 已 APPROVE(重传 P1 已修),并留 1 条 MEDIUM(分类排序不持久化)+ LOW(nav.json 重复键)。本轮针对新 commit `7cc0a867 fix(system-skill): persist category ordering` 实证复核。**结论:维持 APPROVE,上轮两条 follow-up 均已修。**

### 验证记录（实际执行 @ 新 head）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3216 modules
- 全量 `vitest run` — **41 passed**(与上轮持平,无回归);9 处失败仍全在未触碰的 `src/store/feature.test.ts`(既有,不阻塞)

### 本轮改动 —— 逐条核实
**✅ [上轮 MEDIUM] 分类排序持久化** — 已修复。`SystemSkill/CategoryTab.handleMove` 现在本地交换后 `await Promise.all([updateCategory(swapped, {sort_order: index}), updateCategory(moved, {sort_order: target})])` 落库,随后 `fetchList()` 回读;失败也 `fetchList()` 回滚到后端真相。兼容层 `updateCategory` 签名补上 `sort_order` 并透传给 `updateSkillCategory`。加了 `if (!canWrite) return` 写门控兜底。

**✅ [上轮 LOW] nav.json 重复键** — 已修复。en/zh 均删掉重复的第二个 `systemMcp` 键(grep 现各 1 处),侧栏 System MCP label 恢复正确。

之前所有阻塞项(密钥进 bundle / nginx `/market` 反代 / SystemSkill 写门控 / 401 / 重传换包)保持已修复,无回归。

### Gate 1 ✅ / Gate 2 ✅（无阻塞）
对抗性复核无新 🔴/🟠。排序修复的交换/持久化/失败回读逻辑正确。

### 🟡 剩余非阻塞(LOW，不 gate merge)
1. **列表顺序依赖后端按 `sort_order` 返回** — `listSkillCategories` 不做客户端排序,持久化后的顺序能否在刷新后正确呈现,取决于后端 `GET /admin/skill_categories` 是否 `ORDER BY sort_order`。属契约假设,建议后端确认(human-verify)。另:交换只改两行的 `sort_order`=其位置下标,若历史数据 `sort_order` 非连续 0..n,极端下可能与邻居短暂重号,`fetchList` 回读可自愈,影响很小。
2. **SystemMcp 沿用 #115 修复前旧版**(carryover)— locale 字符串跳步 / 缺 bearer sentinel / `systemMcp.json` `form` 内 `sectionTools*` 重复键。建议 #115 合入后 rebase 统一。
3. **多处硬编码中文**(carryover),en-US 界面显示中文。
4. **peer 提及的既有非阻塞**:`/system-skill` 无侧栏入口、parse 轮询用 `setInterval`、SkillMarket 下载用 `window.open`(`SkillTab.tsx:105`,与已批准 head 字节一致、非本次增量)—— 均非回归,可 follow-up。

### 覆盖与盲区
- 未跑 marketplace 后端:category 排序回读、reupload、create、download presigned、401→登出全链路需真实环境手测。
- probe SSRF 防护属后端职责,建议后端确认。

---

**Verdict: APPROVE。** 上轮 MEDIUM + LOW 均已实证修复,无回归。剩余全为非阻塞 LOW。感谢持续快速迭代 👍

yujiawei (approved):
# Code Review — PR #116 (octo-admin)

Reviewed at head SHA `7cc0a86` against merge-base `3e42799`. Scope: admin console for platform-provided MCP management, Skill Market admin, and System Skill admin, plus nav / capabilities / i18n / upload-download / probe / dev-proxy wiring (40 files, ~9.9k additions).

Verification performed locally at this head: `tsc && vite build` ✅, `vitest run` ✅ (50 tests pass, incl. new `probeHelpers.test.ts` 14 cases and `skill.test.ts`).

## 1. Specification compliance

**Spec: ✅**

Every item in the PR summary is implemented and wired end to end:
- MCP admin: `src/api/mcp.ts`, `src/pages/SystemMcp/*` (list/create/edit/delete/probe/icon-upload).
- Skill Market admin: `src/api/skill.ts`, `src/pages/SkillMarket/*` (category + public skill CRUD, upload→parse→create).
- System Skill admin: `src/pages/SystemSkill/*`.
- Nav, capabilities (`mcp.read/write`, `skill.read/write` in `src/auth/capabilities.ts`), i18n (`src/i18n/index.ts` + locale bundles), and the `/market/api/` proxy in both `vite.config.ts` (dev) and `nginx.conf.template` (prod).

No missing deliverables, no functionality beyond the stated scope, no contradiction with the described approach.

Note on a prior state of this branch: an earlier commit shipped a bundled marketplace admin token; it was removed in `1f3a041` ("remove bundled admin token"). The current head correctly authenticates every marketplace call with the operator's own login token via the request interceptor — **no shared secret is present in the browser bundle at this head**. This was verified by scanning the full diff.

## 2. Code quality

**Quality: Approved**

Build and tests are green; no P0/P1 client-side blocking defect found. Data-flow tracing confirms the two most security-relevant claims in the code comments:
- The ephemeral probe bearer (`probeBearer`) is component-local state used **only** by `buildProbeRequest`; `buildPayload` never reads it, so it cannot be persisted into the create/update payload (`src/pages/SystemMcp/FormModal.tsx`, `probeHelpers.ts`).
- All user/server-controlled strings (skill `SKILL.md`, headers/env, descriptions, readme) render as escaped React text — no `dangerouslySetInnerHTML` is introduced anywhere in the diff.

### Items to verify manually before/at merge (security-sensitive)

These are not client-fixable in isolation — they depend on the marketplace backend — but on a security-sensitive surface they warrant explicit human confirmation:

- **[Verify — server] Probe SSRF.** `probeSystemMcp` forwards an operator-supplied URL (`src/pages/SystemMcp/FormModal.tsx` → `POST /admin/mcps/probe`) and the backend performs the outbound request. The endpoint is superAdmin-only, but the server must still reject loopback / link-local / RFC1918 / cloud-metadata / redirect destinations. Please confirm the marketplace probe handler has SSRF guards.
- **[Verify — server] Secret redaction on read.** `SystemMcp/DetailDrawer.tsx:206` renders `quickStart.headers` verbatim. This is correct **iff** the server always returns the `SECRET_PLACEHOLDER` sentinel (never a live `Authorization` value) on GET. Please confirm the read path redacts secrets; a client-side redaction of `Authorization`-like keys would add cheap defense-in-depth.
- **[Info] Write authz is UI-gated only.** `mcp.write` / `skill.write` only hide controls; the API layer has no guard. This is expected for a SPA (server enforces superAdmin), noted for completeness — confirm server-side authz covers create/edit/delete/reupload.

### P2 / suggestions (non-blocking)

- **Off-origin token exposure risk via config.** `src/api/mcp.ts` / `src/api/skill.ts` attach the login token to every request regardless of `VITE_MARKETPLACE_API_BASE`. Default (`/market/api/v1`) is same-origin and safe; a deploy that sets an absolute third-party origin would send the admin token off-origin. Consider asserting the base is same-origin/relative, or documenting that constraint next to the env var.
- **`window.open(url, '_blank')` without `noopener`.** `SkillMarket/SkillDetailDrawer.tsx:128` and `SkillTab.tsx:103` open a server-controlled download URL; add `'noopener'` to avoid reverse-tabnabbing. (Consistent with the pre-existing pattern in `Backup/index.tsx`, so this is a codebase-wide nit, not a regression.)
- **Dead / inconsistent helpers in `src/api/skill.ts`.** `reuploadAdminSkill` (line ~475) is exported but never called, and it inits a *new-skill* upload rather than the reupload-init endpoint — if it is ever wired up it would upload into the wrong workflow. Either remove it or fix it to use `initReupload`. Relatedly, `uploadToPresigned` (line ~530) hardcodes `PUT` while the init response carries a `method`; harmonize with the `fetch`-based flows that honor `init.method`.
- **Two near-identical skill surfaces.** `/skill-market` (`SkillMarket/`) and `/system-skill` (`SystemSkill/`) both target the same `/admin/skills` endpoints and both carry the "Skill Market" label. Only `/skill-market` is added to the sidebar and `firstManagerPath`; `/system-skill` has a route + gate but no nav entry, so it is effectively unreachable via normal navigation. Both are listed in the PR summary so this is in-scope, but consider consolidating or clearly differentiating them to avoid drift and operator confusion.
- **Header value parsing.** `parseKV(raw, ':')` splits on the first colon, so values containing colons (URLs) survive correctly — good. Just note `serializeKV` uses `': '` and `parseKV` trims, so a header value with leading/trailing spaces will not round-trip byte-for-byte; acceptable for this use.

## 3. Overall verdict

**APPROVED.** Spec is fully met, build and tests pass, and no P0/P1 client-side blocking defect was found. The remaining items are P2/nits plus server-dependent security checks that should be confirmed on the marketplace side given the security-sensitive classification — none of them block the frontend change from merging.

## 4. Reviewer's note on coverage

This review combined a full manual read of the diff with an independent adversarial pass. One additional automated reviewer leg was attempted but timed out with no output and is therefore treated as absent — it did **not** contribute a silent approval. Backend behavior (marketplace authorization, SSRF defenses, secret redaction, presigned-URL issuance, response-envelope contracts) is outside this repository and could not be verified here; it is flagged above for human/server-side confirmation.

Jerry-Xin (approved):
Re-review at head `7cc0a867`. Verdict: **APPROVE** (unchanged from the prior approval at `537e17a0`).

## Real delta
Single linear commit `fix(system-skill): persist category ordering` (ahead 1 / behind 0 — no rebase, no drift). Four files:
- `src/pages/SystemSkill/CategoryTab.tsx` — `handleMove` now persists order.
- `src/api/skill.ts` — `updateCategory` accepts optional `sort_order`.
- `src/i18n/locales/{en-US,zh-CN}/nav.json` — removed the duplicate `systemMcp` key.

## Review
- `handleMove` is now `async`, gated by `canWrite` (matches `handleDelete`/`handleSubmit`), does the optimistic swap, then persists both affected rows via `Promise.all([updateCategory(swapped.id,{sort_order:index}), updateCategory(moved.id,{sort_order:target})])` and re-syncs with `fetchList()`. On error it surfaces `message.error` and re-fetches to roll back the optimistic UI. Correct — this closes the prior "category sort not persisted" note.
- `updateCategory` cleanly threads the optional `sort_order` through to `updateSkillCategory`; no behavior change for existing callers.
- Removing the duplicate `systemMcp` nav key also closes one of the prior LOW items.

## No regression (byte-verified)
- `src/api/mcp.ts` and `src/pages/SystemSkill/SkillFormModal.tsx` are blob-identical to the approved head `537e17a0` — the secret-in-bundle fix and the skill re-upload endpoint fix (`POST /admin/skills/{id}/reupload` via `commitAdminSkillReupload`) are intact.
- `src/api/skill.ts` still routes the reupload flow through `/admin/skills/:id/reupload` (+ `/reupload/init`); no `VITE_MARKETPLACE_ADMIN_TOKEN` / `X-Admin-Token` references reintroduced.
- nginx config untouched; `/market/api/` proxy unaffected.

## Non-blocking (unchanged from prior review)
- `/system-skill` still has no sidebar entry.
- Ordering persistence issues two independent requests without a transaction; a partial failure could leave two rows sharing/gapping a `sort_order` until the next `fetchList`. Acceptable for an admin reorder action, but a single batch/atomic endpoint would be cleaner.

Nice, focused fix — approving.

yujiawei (approved):
# Code Review — PR #116 (octo-admin)

**Verdict: APPROVED** · head `7cc0a867ce41401bda1121550cf361d70870a5f6` · base `main` (merge-base `3e42799`)

This is a follow-up review. Two new commits landed on top of the previously-reviewed revision, and both resolve the blockers raised earlier. Build (`tsc && vite build`) is green and the full test suite passes (50/50).

## 1. Spec compliance

The delta since the prior revision does exactly what was asked, with no scope creep:

- `537e17a fix: commit system skill reuploads` — addresses the sole blocker.
- `7cc0a86 fix(system-skill): persist category ordering` — addresses the category-ordering gap plus a duplicated i18n nav key.

**Spec: ✅**

## 2. Code quality

### Blocker resolved — system-skill reupload now commits the package

Previously the edit-submit path in `SystemSkill/SkillFormModal.tsx` ran the reupload UI (init → presigned PUT → parse) but then only called the metadata PATCH (`PATCH /admin/skills/:id`), whose param type (`PatchSkillParams`) does not carry `parse_task_id`. The result was a success toast while the package was never replaced (silent no-op).

The fix (`src/api/skill.ts`, `src/pages/SystemSkill/SkillFormModal.tsx`) is correct and complete:

- New `commitAdminSkillReupload(skillId, params)` posts to `POST /admin/skills/:id/reupload` with `parse_task_id` — the endpoint that actually commits a package version.
- `SkillFormModal.handleSubmit` now, when a `parseTaskId` exists, calls `commitAdminSkillReupload` first, then runs the metadata PATCH separately. `parse_task_id` is no longer smuggled into the metadata PATCH.
- `reuploadAdminSkill` was refactored to reuse the same helper, so both call sites share one contract.
- A focused unit test (`src/api/skill.test.ts`) asserts the reupload endpoint and payload.

Verified `PatchSkillParams` (`skill.ts:131`) is metadata-only, and the edit path passes only metadata fields to `updateSkill` — the previous contract mismatch is gone.

### Category ordering now persisted

`SystemSkill/CategoryTab.tsx#handleMove` previously reordered only local state. It now issues two `updateCategory(..., { sort_order })` PATCHes (via `PATCH /admin/skill_categories/:id`) for the swapped rows and refetches. The `sort_order` field was threaded through `updateCategory` → `updateSkillCategory`, and the swap-index math is correct (each moved row gets its new position). A `canWrite` guard was also added, matching the write-gating pattern used elsewhere. The redundant `systemMcp` key duplicated in `nav.json` was removed at the same time.

**Quality: Approved.** No P0/P1 remain.

## Non-blocking suggestions (nits, not merge blockers)

- **Partial-failure on edit reupload:** in `SkillFormModal.handleSubmit`, the reupload commit and the metadata PATCH are two sequential awaits. If the reupload succeeds but the metadata PATCH throws, the package is replaced while metadata is not. This is acceptable (the package commit is the primary operation) but worth a follow-up if atomicity matters.
- **Optimistic reorder:** `handleMove` mutates local state before the PATCH resolves and refetches on error to recover — fine, but two rapid clicks can race. A pending-guard would tighten it.
- A few smaller items from the prior pass remain open and are all non-blocking: a stale-response guard on the SkillMarket detail drawer, an action-button click bubbling into the row handler, an un-cleared module-global parse-task map, and some hardcoded strings in form helpers.

## For manual verification (security-sensitive PR)

- Confirm the marketplace backend enforces `superAdmin` server-side on `POST /admin/skills/:id/reupload`, `PATCH /admin/skills/:id`, and `PATCH /admin/skill_categories/:id` — the admin console relies on the user's own token, with authorization enforced server-side (there is no bundled admin token in the bundle, which is correct).
- Confirm the reupload endpoint's `parse_task_id` is bound to the target skill server-side so a task parsed for skill B cannot be committed against skill A.

PR #118, opened 2026-07-21
fix(changelog): stop contributor avatar hover flicker and overflow

an9xyz:
## Summary

The contributor avatar stack on the Changelog page fanned out on hover by animating `margin`, which changed the group's layout width and re-flowed/wrapped the parent flex row **under the cursor** — producing a hover flicker loop. With many contributors the fanned stack also overflowed past the card's rounded corner.

This PR reworks the interaction so it is flicker-free and overflow-safe, on both desktop and touch.

## Changes

- **No flicker**: fan the stack with `transform: translateX()` (per-avatar index) instead of `margin`, so hover never triggers layout reflow. Fanned step (20px) stays below the avatar width (24px), so avatars remain gapless and the cursor never falls into a gap.
- **No overflow**: the group reserves its *fanned* width up front (via a CSS var), so the fan fills that box exactly and the whole group wraps as one unit instead of spilling over the card.
- **Bounded for large lists**: show at most 10 avatars plus a `+N` overflow badge; the badge's `title` lists the remaining contributor names.
- **Mobile**: under `@media (hover: none)` the fan/lift/name-tooltip are disabled and the reserved width falls back to `auto`, so a sticky tap-triggered `:hover` can't shift avatars or cause horizontal overflow.

Scope is limited to `src/pages/Changelog/index.tsx`.

## Test plan

- [x] `tsc --noEmit` passes
- [x] `vitest run` — 49/49 tests pass
- [x] Verified in a real browser (Playwright) against the test backend:
  - [x] Desktop: hover fans avatars out with no flicker; rightmost avatar stops at the content edge (no overflow past the card)
  - [x] Desktop: no page horizontal scrollbar
  - [x] Mobile (390px, touch): contributor group stays within the viewport (no overflow)

## Notes

- A pre-existing, unrelated horizontal overflow on mobile comes from the platform filter `Tabs` (6 icon tabs on narrow screens), not from the contributor avatars. Left out of this PR — happy to address separately.

### DIFF
--- a/src/pages/Changelog/index.tsx
+++ b/src/pages/Changelog/index.tsx
@@ -90,28 +90,68 @@ const css = `
 .contributor-group {
   display: flex;
   align-items: center;
+  width: var(--reserved-w, auto);
 }
 .contributor-avatar {
   position: relative;
-  transition: transform 0.2s ease, z-index 0s;
+  transition: transform 0.25s cubic-bezier(0.22, 1, 0.36, 1);
   z-index: 0;
 }
 .contributor-avatar:not(:first-child) {
   margin-left: -8px;
 }
-.contributor-group:hover .contributor-avatar {
-  margin-left: 4px;
+.contributor-avatar img {
+  transition: transform 0.2s ease;
 }
-.contributor-group:hover .contributor-avatar:first-child {
-  margin-left: 0;
+/* Fan the stack out on hover using transform only — no layout change, so the
+   parent flex row never re-flows/wraps under the cursor (which caused flicker).
+   Each avatar shifts right by index*4px, taking the per-avatar step from 16px to
+   20px. The group already reserves this fanned width (inline), so the fan fills
+   the reserved box exactly and never overflows the card. Avatars still overlap
+   (20 < 24), so there are no gaps for the cursor to fall into. */
+.contributor-group:hover .contributor-avatar {
+  transform: translateX(calc(var(--contributor-index, 0) * 4px));
 }
 .contributor-avatar:hover {
-  transform: translateY(-4px) scale(1.15);
   z-index: 10;
+}
+.contributor-avatar:hover img {
+  transform: translateY(-4px) scale(1.2);
 }
 .contributor-avatar:hover .contributor-name {
   opacity: 1;
   transform: translateX(-50%) translateY(0);
+}
+.contributor-more-badge {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 24px;
+  height: 24px;
+  border-radius: 50%;
+  border: 2px solid var(--bg-page);
+  background: var(--bg-hover);
+  color: var(--text-secondary);
+  font-size: 10px;
+  font-weight: 600;
+  line-height: 1;
+  cursor: default;
+}
+/* Touch devices have no reliable hover; disable the fan/lift so a sticky
+   tap-triggered :hover can't push avatars around or reveal the tooltip. */
+@media (hover: none) {
+  .contributor-group {
+    width: auto;
+  }
+  .contributor-group:hover .contributor-avatar {
+    transform: none;
+  }
+  .contributor-avatar:hover img {
+    transform: none;
+  }
+  .contributor-avatar:hover .contributor-name {
+    opacity: 0;
+  }
 }
 .contributor-name {
   position: absolute;
@@ -388,8 +428,23 @@ function Heatmap({ data }: { data: AppVersion[] }) {
   )
 }
 
+// Stacked contributor avatars. Collapsed (overlapped) at rest, gently fanned
+// out on hover. The group reserves its *fanned* width up front so the surrounding
+// flex row accounts for it — the fan happens inside that reserved box and can
+// never overflow the card (the whole group wraps to its own line if too wide).
+const MAX_VISIBLE_CONTRIBUTORS = 10
+const AVATAR_SIZE = 24
+// Per-avatar step: 16px at rest (8px overlap, via margin-left:-8px in CSS) →
+// 20px on hover (4px overlap). The group reserves the fanned width below.
+const FAN_STEP = 20
+
 function ContributorAvatars({ contributors, showLabel }: { contributors: Contributor[]; showLabel?: boolean }) {
   if (contributors.length === 0) return null
+
+  const visible = contributors.slice(0, MAX_VISIBLE_CONTRIBUTORS)
+  const overflow = contributors.slice(MAX_VISIBLE_CONTRIBUTORS)
+  const itemCount = visible.length + (overflow.length > 0 ? 1 : 0)
+  const reservedWidth = (itemCount - 1) * FAN_STEP + AVATAR_SIZE
 
   return (
     <div style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
@@ -399,9 +454,16 @@ function ContributorAvatars({ contributors, showLabel }: { contributors: Contrib
           贡献者
         </span>
       )}
-      <div className="contributor-group">
-        {contributors.map((c) => (
-          <div key={c.name} className="contributor-avatar">
+      <div
+        className="contributor-group"
+        style={{ ['--reserved-w' as string]: `${reservedWidth}px`, flexShrink: 0 } as React.CSSProperties}
+      >
+        {visible.map((c, index) => (
+          <div
+            key={c.name}
+            className="contributor-avatar"
+            style={{ ['--contributor-index' as string]: index } as React.CSSProperties}
+          >
             <img
               src={c.avatar}
               alt={c.name}
@@ -410,8 +472,8 @@ function ContributorAvatars({ contributors, showLabel }: { contributors: Contrib
                 event.currentTarget.src = c.fallbackAvatar
               }}
               style={{
-                width: 24,
-                height: 24,
+                width: AVATAR_SIZE,
+                height: AVATAR_SIZE,
                 borderRadius: '50%',
                 border: `2px solid var(--bg-page)`,
                 background: colors.surface.subtle,
@@ -422,6 +484,15 @@ function ContributorAvatars({ contributors, showLabel }: { contributors: Contrib
             <span className="contributor-name">{c.name}</span>
           </div>
         ))}
+        {overflow.length > 0 && (
+          <div
+            className="contributor-avatar contributor-more"
+            style={{ ['--contributor-index' as string]: visible.length } as React.CSSProperties}
+            title={overflow.map((c) => c.name).join('、')}
+          >
+            <span className="contributor-more-badge">+{overflow.length}</span>
+          </div>
+        )}
       </div>
     </div>
   )
--- a/src/pages/Changelog/index.tsx
+++ b/src/pages/Changelog/index.tsx
@@ -460,7 +460,7 @@ function ContributorAvatars({ contributors, showLabel }: { contributors: Contrib
       >
         {visible.map((c, index) => (
           <div
-            key={c.name}
+            key={`${c.name}-${index}`}
             className="contributor-avatar"
             style={{ ['--contributor-index' as string]: index } as React.CSSProperties}
           >
@@ -486,7 +486,7 @@ function ContributorAvatars({ contributors, showLabel }: { contributors: Contrib
         ))}
         {overflow.length > 0 && (
           <div
-            className="contributor-avatar contributor-more"
+            className="contributor-avatar"
             style={{ ['--contributor-index' as string]: visible.length } as React.CSSProperties}
             title={overflow.map((c) => c.name).join('、')}
           >


lml2468 (dismissed):
## Review @ `1d02b9653b28` — 独立本地复核（非转述）

从 merge-base `d75b1d86` 的本地 checkout 复核。单文件、纯前端交互修复。**结论:APPROVE。**

### 验证记录（实际执行）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 产物正常
- `src/pages/Changelog` 测试 — ✅ 通过
- 全量 `vitest` 的 9 处失败**全在 `src/store/feature.test.ts`**,且原因是我的 Node 测试环境 `localStorage is not available (--localstorage-file 未提供)` 的环境性报错,`blame` 核实该文件**本 PR 未触碰**(merge-base 即在)。作者环境有 localStorage 故看到 49/49 —— 属环境差异,非本 PR 回归,不阻塞。

### Gate 1 规格 ✅
意图:修复 Changelog 贡献者头像 hover 抖动 + 溢出卡片圆角。逐条对齐描述,无缺失/超范围/偏离,scope 严格限于 `Changelog/index.tsx`。

### Gate 2 代码质量 ✅（无阻塞）
- **消抖**正确:hover fan 从 `margin` 改为 `transform: translateX(index*4px)`,不触发 layout reflow → 父 flex 行不再在光标下 re-flow,根因消除。
- **防溢出**正确:group 预留 fanned 宽度 `--reserved-w = (itemCount-1)*20 + 24`,与 hover 后最右头像右边缘精确吻合(第 i 个 = i*20 起始 + 24 宽 = 预留值),fan 在预留框内发生,整组作为单元换行,不越卡片。
- **步距/间隙**:静止 16px、hover 20px,均 < 头像 24px → 始终重叠,光标不会掉进空隙(重启抖动的另一诱因),数学成立。
- **大列表有界**:最多 10 个 + `+N` badge,`title` 列出余下贡献者名,`--contributor-index` 给到 `visible.length` 使 badge 正确参与 fan。
- **移动端**:`@media (hover: none)` 关 fan/lift/tooltip 且宽度回落 `auto`,规避 sticky tap `:hover` 造成的位移/横向溢出。
- 对抗性复核未发现 🔴/🟠:`key={c.name}`(贡献者名唯一,合理);inline CSS var 用 `React.CSSProperties` 断言,tsc 通过;`onError` fallback 头像逻辑保留未动。

### 🟡 非阻塞(LOW，不 gate)
1. **`key={c.name}`** 若同名贡献者并存会 key 冲突;当前数据源大概率唯一,风险很低,可用 id/index 兜底。
2. 描述已诚实说明:移动端 `Tabs` 的既有横向溢出为 pre-existing、非本 PR 范围 —— 认可,建议单独跟进。

### 覆盖与盲区
- 视觉/hover 行为已由作者用 Playwright 在真机验证(desktop 无抖动/不溢出、mobile 390px 不溢出);我在无头环境未复现视觉,以代码+几何推导核实,建议合并前人工在浏览器再扫一眼 hover 手感。

---

**Verdict: APPROVE。** 聚焦、正确、几何自洽的交互修复,无回归,剩余仅 LOW。

Jerry-Xin (dismissed):
The PR is relevant to the Changelog UI and resolves the layout-driven hover flicker without introducing blocking issues.

The root cause of the old flicker was hover styling that changed **layout** (`margin-left: -8px` → `4px`), reflowing the flex row under the cursor so `:hover` dropped and re-triggered in a loop, and letting the expanded stack overflow the card. The fix replaces the margin-based expansion with a `transform: translateX()` fan-out (no reflow) and reserves the fanned width up front via inline `--reserved-w` with `flexShrink: 0`, so the fan happens inside a stable, pre-reserved box and can never overflow. The lift/scale moves to the inner `img` so the hover hit area stays stable, and an `@media (hover: none)` block disables the fan/lift/tooltip on touch to avoid sticky-tap displacement. This addresses the root cause rather than masking it.

💬 Non-blocking

- 🔵 Suggestion — Add a focused component or browser regression test covering 10+ contributors, reserved width, and touch behavior. The width calculation and media-query behavior currently lack automated coverage. (`src/pages/Changelog/index.tsx:431`)
- 🔵 Suggestion — The overflow contributors are exposed only through a `title` attribute, which is unavailable to keyboard and most touch users. Consider an accessible label or focusable disclosure. (`src/pages/Changelog/index.tsx:487`)

✅ Highlights

- Transform-based fan-out avoids layout reflow and the original hover loop.
- Reserved width correctly matches the 24px border-box avatars and 20px fanned step (verified: `(itemCount-1)*20 + 24`).
- Capping the rendered stack at 10 with a `+N` overflow badge bounds layout and rendering cost.
- Touch-specific rules prevent sticky-hover displacement.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

Dispatcher bridge markers (automated review leg):

yujiawei (dismissed):
# Code Review — PR #118 (octo-admin)

## Summary

This reworks the Changelog contributor-avatar hover interaction to eliminate the flicker loop and card overflow. The approach is correct and well-executed for a presentation-only change confined to `src/pages/Changelog/index.tsx`.

**Verdict: APPROVED.** No blocking (P0/P1) issues found. A couple of minor nits below are optional.

## Spec compliance

All four stated goals are implemented and hold up under inspection:

- **No flicker** — the fan is driven by `transform: translateX(calc(var(--contributor-index) * 4px))` instead of animating `margin`. Hover now only changes `transform`/`opacity`/`z-index`, none of which trigger layout; the group's width is fixed at rest and on hover, so the parent flex row can no longer re-wrap under the cursor. This directly addresses the root cause of the flicker loop.
- **No overflow** — the group reserves its *fanned* width up front via `--reserved-w`, and the fan happens inside that reserved box, so the whole group wraps as a unit rather than spilling past the card.
- **Bounded** — at most 10 avatars plus a `+N` badge; the badge participates in the same fan geometry.
- **Mobile** — `@media (hover: none)` disables the fan/lift/tooltip and resets the reserved width to `auto`.

## Correctness notes (verified)

- **Reserved-width math is exact.** Avatar `i`'s right edge on hover is `i*16` (rest step from `margin-left:-8px` on a 24px box) `+ i*4` (fan) `+ 24` = `i * FAN_STEP + AVATAR_SIZE`. The last item's index is `itemCount - 1`, so its right edge equals `reservedWidth = (itemCount - 1) * FAN_STEP + AVATAR_SIZE`. The `+N` badge (`--contributor-index = visible.length`) is correctly included in `itemCount`, so the reserved box matches the fanned content precisely — no residual overflow, no dead space beyond the last item.
- **`box-sizing: border-box` is global**, so the `2px` avatar border sits inside the 24px box; `AVATAR_SIZE = 24` correctly equals the on-screen footprint, keeping the math consistent.
- **The mobile override is wired correctly.** `width` is controlled by the stylesheet rule (`width: var(--reserved-w, auto)`) while the inline style only sets the `--reserved-w` custom property. That's what lets `@media (hover: none) { .contributor-group { width: auto } }` win (later source order, equal specificity). If the width had been set inline instead, the mobile fallback could not have overridden it — this is the right pattern.
- **Lift moved to the inner ``** so the container's `transform` is free for the fan; the two transforms no longer conflict. The absolutely-positioned name tooltip stays anchored to the container, which is correct.

Build/test claims confirmed locally at head SHA `1d02b96`: `tsc --noEmit` passes and `vitest run` reports 49/49 passing.

## Minor / non-blocking

- **Dead class**: the overflow badge div carries `contributor-more`, but there is no `.contributor-more` CSS rule (styling comes from `.contributor-more-badge` on the inner span). Harmless; drop the class or add a rule for clarity.
- **Overflow names via `title`**: the remaining contributors are surfaced only through the `title` attribute, which is hover-only and not exposed to keyboard/screen-reader users. Acceptable for this context, but a more accessible affordance could be a follow-up.
- **`key={c.name}`** (pre-existing, not introduced here) would collide if two contributors share a display name. Out of scope for this PR; noting for awareness.

Nice touch reserving the fanned width via a CSS var rather than inline `width` — that single decision is what makes both the no-reflow and the mobile fallback work.

lml2468 (approved):
## Re-review @ `a32c21bfcf06` — 独立本地复核（非转述）

我此前在 `1d02b965` 已 APPROVE。本轮新增 `a32c21bf refactor(changelog): address review nits`,针对 review nits 的两处小改。**结论:维持 APPROVE。**

### 真实增量（`1d02b965..a32c21bf`,仅 2 行）
1. **`key={c.name}` → `key={\`${c.name}-${index}\`}`** — 直接收口我上轮 LOW #1(同名贡献者 key 冲突风险)。✅
2. **删掉死类 `contributor-more`** — 该 class 无任何 CSS 规则(grep 核实:bare `.contributor-more` 命中 0;真正带样式的是 `.contributor-more-badge`,保留未动),移除纯清理,无行为变化。✅

### 验证记录（本轮实跑 @ 新 head）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 产物正常
- `src/pages/Changelog` 测试 — ✅ 通过

### 结论(承接 `1d02b965` 的完整审查,无回归)
消抖(transform 替代 margin、不 reflow)、防溢出(预留 `--reserved-w` 几何自洽)、大列表有界(10 + `+N` badge)、移动端 `@media (hover:none)` 兜底 —— 全部保持正确。本轮两处 nit 修复不触及这些逻辑。

### 🟡 剩余非阻塞(不变,不 gate)
- 溢出贡献者名仅在 `title` 属性,键盘/触屏读不到(可访问性小项);建议补 10+/预留宽/触屏回归测试。
- `check-sprint` 红是看板流程门(需 `Closes #<issue>` 关联),非代码/构建/测试失败,由 maintainer 补 issue 关联,非 review 阻断项。

---

**Verdict: APPROVE。** 两处 nit 修复干净、切题,无回归。

Jerry-Xin (dismissed):
The transform-based animation fixes hover reflow, but the maximum contributor stack can still overflow on touch-width timeline cards.

🔴 Blocking

- 🔴 Critical — Mobile overflow remains possible. At lines 143–145, touch devices switch the group to its collapsed width. With ten avatars plus the overflow badge, that width is `24 + 10 × 16 = 184px`. Including the label and gap, the non-wrapping wrapper at line 450 is approximately 244px wide. At a 390px viewport, the timeline content column is only about 230px wide after page padding and the fixed 112px time/rail columns (lines 772–808). Because the wrapper cannot wrap internally, it overflows the card. This contradicts the PR’s overflow-safety goal. Please make the complete contributor component responsive—for example, reduce the mobile visible count, allow the label and stack to wrap, or adjust the narrow-screen timeline layout.  
  `src/pages/Changelog/index.tsx:143`  
  `src/pages/Changelog/index.tsx:444-450`  
  `src/pages/Changelog/index.tsx:772-809`

💬 Non-blocking

- 🔵 Suggestion — Add a component or browser regression test covering the maximum visible count at a narrow viewport. Existing tests only cover contributor parsing and would not detect this layout regression.

✅ Highlights

- Using per-index transforms avoids layout-triggered hover flicker.
- Reserving the desktop fanned width correctly matches the 20px visual step.
- Bounding large contributor lists and disabling hover effects on touch devices are sensible improvements.

Jerry-Xin (approved):
Re-review at head a32c21bfcf06. The new commit cleanly addresses two of the non-blocking nits from the prior review, and the root-cause hover/overflow fix remains intact. Approving.

✅ Delta since last approve (single commit, 1 file)

- Avatar key changed from `key={c.name}` to `` key={`${c.name}-${index}`} `` — resolves the duplicate-display-name collision risk raised previously. Keys are now unique per rendered item.
- Dropped the unused `contributor-more` class from the overflow badge's outer `<div>`. Verified there is no `.contributor-more` selector in the stylesheet; the badge styling comes solely from `.contributor-more-badge` on the inner `<span>`, which is unchanged. No dead/orphaned CSS, no visual change.

✅ Root-cause fix still holds (unchanged this commit)

- Hover fan still uses `transform: translateX(...)` (no reflow), reserved width `--reserved-w = (itemCount-1)*FAN_STEP + AVATAR_SIZE` intact, lift/scale on the inner ``, and `@media (hover: none)` still disables the fan/tooltip on touch.

💬 Non-blocking (carried forward, follow-ups)

- Accessible disclosure for the "+N" overflow names (currently only in the `title` attribute) and a regression test for the max-visible-count layout are reasonable follow-ups, as the author noted.

Note on narrow-viewport layout: the contributor block is `flex-shrink: 0` but sits inside a row with `flexWrap: 'wrap'` under a `flex: 1; min-width: 0` content column, so on small screens the block wraps to its own line rather than forcing horizontal card overflow. Not a blocker for this change.

yujiawei (approved):
# Code Review — PR #118 (octo-admin)

## Summary

Reworks the Changelog contributor-avatar hover interaction to eliminate a hover flicker loop and prevent horizontal overflow. Scope is limited to `src/pages/Changelog/index.tsx`. The fix is correct and well-reasoned.

## Spec compliance

- **Missing:** none — every claim in the PR description is implemented (transform-based fan, reserved fanned width, `+N` overflow cap at 10, `@media (hover: none)` mobile fallback).
- **Over-build:** none — change stays inside the single stated file; no unrequested flags/fields/APIs.
- **Deviation:** none — implementation matches the described approach.

Verified locally at head SHA `a32c21b`:
- `tsc --noEmit` — passes (custom-property casts compile cleanly).
- `vitest run` — 49/49 pass.

## What's correct

- **Flicker root cause fixed.** The old code animated `margin-left`, which changed the group's layout width and re-flowed/wrapped the parent flex row under the cursor. The new code fans the stack with a paint-only `transform: translateX(index * 4px)`, so hover never triggers reflow.
- **Width math is exact.** `box-sizing: border-box` is global (`src/index.css:15`), so `AVATAR_SIZE = 24` equals the rendered avatar width including its 2px border. The rest step is 16px (`24 − 8px` overlap) and the fanned step is 20px, so the rightmost avatar's hover shift `(n−1)*4px` lands the visual right edge exactly at `reservedWidth = (itemCount−1)*20 + 24`. The fan fills the reserved box precisely — no spill, no gap for the cursor to fall into (`20 < 24`).
- **Wraps as a unit.** The group reserves its fanned width via `--reserved-w` with `flexShrink: 0`, and both call sites' parent rows use `flexWrap: 'wrap'` (`index.tsx:809`, `index.tsx:925`), so an oversized group wraps to its own line instead of overflowing the card.
- **Mobile-safe.** `@media (hover: none)` overrides `width` back to `auto` (source-later, equal-specificity rule wins) and disables the fan/lift/tooltip, so a sticky tap-triggered `:hover` can't shift avatars or cause horizontal overflow.
- **Overflow cap.** `MAX_VISIBLE_CONTRIBUTORS = 10` + `+N` badge with a `title` listing the remaining names bounds large lists; `itemCount` correctly includes the badge in the reserved-width calculation. Edge cases (exactly 10 → no badge; 11 → `+1`) behave correctly.

## Non-blocking suggestions (P2 / nits)

- **Coupled magic numbers.** The relationship between `AVATAR_SIZE` (24, JS), `FAN_STEP` (20, JS), the `4px` translateX and `-8px` overlap (CSS) is only linked by comments. If the avatar size ever changes, several spots must be updated in lockstep. Consider deriving the CSS step from the same constants (e.g. a CSS var for the fan delta) to keep them in sync. The inline comments do a good job documenting the invariant for now.
- **At-rest trailing space.** Because the group reserves the *fanned* width, at rest the avatars sit left-aligned inside a slightly wider box, leaving `(n−1)*4px` of empty space on the right (visible where the group is right-pushed, e.g. `index.tsx:939`'s `flex:1` spacer). This is the intended trade-off for flicker-free fanning; just noting it's a deliberate visual, not a bug.
- **Badge accessibility.** The `+N` badge conveys meaning via `title` only; a screen reader gets no accessible name beyond the raw text. An `aria-label` (e.g. "N more contributors") would be a small improvement. Not blocking.

## Verdict

No correctness, security, or build-breaking issues; the fix does exactly what it claims and is limited to one file. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #118 (octo-admin)

## Summary

This PR reworks the contributor avatar stack on the Changelog page to eliminate hover flicker (caused by `margin`-based fan animation triggering layout reflow under the cursor) and prevent overflow past the card's rounded corner. The approach is sound: fan the stack via `transform: translateX()` with per-avatar CSS custom properties, reserve the fanned width up front so the flex row accounts for it, cap visible avatars at 10 with a `+N` overflow badge, and disable all hover effects on touch devices via `@media (hover: none)`. Scope is confined to `src/pages/Changelog/index.tsx`.

## Verification

- ✅ **Data-flow trace** — `contributors` flows from `parseContributors(item.update_desc)` (`src/pages/Changelog/utils.ts:202`) through `ContributorAvatars` at `:861` and `:940`. The early-return guard at `:442` (`contributors.length === 0`) prevents any arithmetic on empty arrays. Slicing, `itemCount`, and `reservedWidth` are all computed after the guard — no reachable path produces NaN or negative width.
- ✅ **Reserved-width math** — `reservedWidth = (itemCount - 1) * FAN_STEP + AVATAR_SIZE` correctly matches the fanned-state rightmost extent. Verified boundary cases: 1 avatar (24px, no movement), 10 avatars (204px), 10+overflow (224px for 11 items). At-rest width `(N-1)*16+24` is strictly ≤ reserved width, so the fan always fills inside the reserved box.
- ✅ **Transform composition** — Fan (`translateX`) is on the `.contributor-avatar` container; lift (`translateY + scale`) is on the nested `img`. These are on different elements, so they compose without conflict. The old single-element `transform` (which mixed fan + lift) is correctly decomposed.
- ✅ **Touch-device fallback** — `@media (hover: none)` overrides reset `width: auto`, disable fan/lift transforms, and force `.contributor-name` opacity to 0. This correctly neutralizes sticky tap-triggered `:hover` on mobile.
- ✅ **Overflow badge** — Gets class `contributor-avatar` (participates in the fan) with `--contributor-index: visible.length`. Has no `` child, so `.contributor-avatar:hover img` selector does not match — no lift on the badge, which is the correct behavior for a non-interactive count indicator.
- ✅ **Transition continuity** — `z-index` is not in the new `transition` shorthand, so it still changes instantly on hover (same effective behavior as the old `z-index 0s`). No regression in stacking order timing.
- ✅ **Key stability** — ``key={`${c.name}-${index}`}`` disambiguates contributors with identical names. Acceptable for this render scope.

Static analysis only at head `a32c21bfcf06597c08732290061d5b7ae0221af2`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns. The math checks out, the data flow is clean, and the mobile fallback is correctly scoped.

- `parseContributors` returns `[]` for missing/malformed `@contributors:` lines — the early-return guard handles this.
- The `flexShrink: 0` inline style prevents the reserved-width group from being compressed by parent flex, which is the correct behavior.
- The `as React.CSSProperties` cast for CSS custom properties is standard React practice for properties not in the `CSSProperties` type.
- The overflow badge `title` attribute lists remaining contributor names with `、` separator, consistent with the Chinese UI context.
- `AVATAR_SIZE` constant (24) matches the CSS badge dimensions (24×24), the `border: 2px solid` matches the avatar border, and `FAN_STEP` (20) matches the CSS hover step math (16px rest + 4px fan = 20px).

## Verdict: APPROVED

Well-scoped UI fix that correctly addresses the root cause (layout reflow from margin animation) with a transform-based approach. The reserved-width strategy prevents overflow, the overflow badge caps visual complexity, and the touch-device fallback is thorough. No concerns.

PR #119, opened 2026-07-21
feat: add skill market and system skill admin

### LINKED ISSUE #120: 运营管理员管理系统 Skill 分类和公开跨 Space Skill
## Summary

运营管理员需要能够在 octo-admin 中管理系统级 Skill 分类，以及公开的、跨 Space 可见的 Skill。

该能力用于让平台侧统一维护 Skill 市场的公共内容，而不是依赖单个 Space 内的管理入口。

## Background

Skill 市场存在两类需要平台统一治理的数据：

- 系统级 Skill 分类：用于组织 Skill 市场的展示结构、排序和归类。
- 公开跨 Space Skill：对所有 Space 可见的公共 Skill，需要平台运营人员能够上传、编辑、下架和维护元数据。

当前需求面向运营管理员，不是普通 Space 管理员能力。

## Current behavior

运营管理员缺少一个明确的后台入口来完成以下操作：

- 管理系统 Skill 分类。
- 管理公开的跨 Space Skill。
- 上传或替换公共 Skill 包。
- 维护 Skill 的名称、描述、分类、标签、图标、版本等运营元数据。

## Expected behavior

在 octo-admin 中提供运营管理员可用的 Skill 管理能力：

- 运营管理员可以查看、创建、编辑、删除系统 Skill 分类。
- 运营管理员可以查看公开跨 Space Skill 列表和详情。
- 运营管理员可以上传新的公开 Skill。
- 运营管理员可以编辑公开 Skill 的运营元数据。
- 运营管理员可以替换已有 Skill 的包文件，并确保新包被实际提交生效。
- 普通无权限管理员不能访问或触发写操作。

## Scope

- In scope:
  - octo-admin 管理端页面、路由和导航入口。
  - Skill 分类管理。
  - 公开跨 Space Skill 的列表、详情、上传、编辑、删除或下架能力。
  - Skill 包上传、解析、重上传提交流程。
  - 管理员能力点 gating，例如 `skill.read` / `skill.write`。
  - 与 marketplace admin API 的前端接入。
  - 中英文 i18n 文案。

- Out of scope:
  - 普通 Space 内私有 Skill 管理。
  - marketplace 后端 API 的业务实现。
  - Skill 包解析器本身的规则变更。
  - 推荐、评分、搜索排序等 Skill 市场增长能力。

## Load-bearing impact

- Authz: 需要区分只读和写权限，前端 gating 不能替代 marketplace 后端的权威鉴权。
- Cross-Space visibility: 公开 Skill 会跨 Space 可见，错误写入可能影响所有 Space 用户。
- Artifact integrity: Skill 包上传和重上传必须真实提交新文件，不能只更新元数据。
- Category consistency: 分类删除、重命名、排序会影响 Skill 市场展示和筛选。
- API contract: 前端依赖 marketplace admin Skill/category/upload/reupload 接口的响应结构和错误 envelope。

## Acceptance criteria

- 具备权限的运营管理员可以进入 Skill 管理页面并看到系统 Skill 分类和公开 Skill 列表。
- 无 `skill.read` 权限的管理员访问路由时被重定向或拒绝。
- 无 `skill.write` 权限的管理员看不到上传、编辑、删除、分类写入等操作入口。
- 分类可以创建、重命名、排序和删除；删除被引用分类时有明确错误提示。
- 公开 Skill 可以通过 presigned upload + parse 流程创建成功。
- 编辑已有 Skill 上传新包后，前端必须调用 reupload commit 接口，确认新包和版本实际替换。
- 元数据编辑不会错误覆盖 reupload commit 设置的版本信息。
- 前端不打包 marketplace admin shared secret；请求使用当前登录管理员的 Octo token，由 marketplace 后端做权威 superAdmin 校验。
- 覆盖关键 API helper 的单元测试，至少包含 Skill reupload commit endpoint/payload 的回归测试。
- `npm test -- --run` 和 `npm run build` 通过。

LarsonLian:
## Summary

Adds the admin Skill surfaces that sit on top of the octo-marketplace admin API:

- **Skill Market admin** (`/skill-market`): list / detail / edit / delete public skills, manage skill categories, upload new skill archives through the marketplace presigned-upload + parse flow, and open short-lived presigned download URLs.
- **System Skill admin** (`/system-skill`): legacy/system-oriented skill management surface with category ordering, detail drawer, package upload/reupload, icon upload, and metadata editing.
- **Marketplace Skill client** (`src/api/skill.ts`): dedicated axios client for `/market/api/v1/admin/*` skill/category/upload endpoints, matching marketplace response envelopes and using the operator's own Octo login token at request time.
- **Admin wiring**: `skill.read` / `skill.write` capabilities, route guards, sidebar entry, en-US / zh-CN i18n, docs/env updates, and vite/nginx marketplace proxy alignment.

System MCP is already present on `upstream/main` via #115. This PR preserves that upstream implementation and only carries the remaining Skill Market / System Skill delta plus shared capability and marketplace-client integration needed by those pages.

## Linked Spec

- Closes #120.
- Current sprint: Sprint W30 (2026-07-21 + 7d).
- Marketplace API contract: `octo-marketplace/docs/api/mcp-v1.md` for the already-merged System MCP admin surface and the marketplace admin auth model.
- Skill marketplace contract: marketplace admin Skill/category/upload endpoints consumed by `src/api/skill.ts` (`/admin/skill_categories`, `/admin/skills`, `/admin/skill_uploads`, `/admin/skill_parse_tasks`, `/admin/skills/:id/reupload`).
- No `.octospec/tasks/<slug>/brief.md` exists in `octo-admin` for this PR. This is load-bearing admin UI work, so the body includes COMPREHENSION below and calls out the contract dependency explicitly.

## How verified

- `npm test -- --run` — 50 tests passed, including `src/api/skill.test.ts` for the reupload commit endpoint and 14 System MCP probe-helper tests inherited from upstream.
- `npm run build` — `tsc && vite build` passed; 3234 modules transformed. Vite still reports the existing large chunk warning.
- `rg -n "VITE_MARKETPLACE_ADMIN_TOKEN|MARKETPLACE_ADMIN_TOKEN|X-Admin-Token" src .env.example vite.config.ts nginx.conf.template Dockerfile README.md README.zh.md` — no matches.
- `rg -a -n "VITE_MARKETPLACE_ADMIN_TOKEN|MARKETPLACE_ADMIN_TOKEN|X-Admin-Token" dist` after production build — no matches.
- Rebased on `upstream/main`; current PR diff no longer includes `src/pages/Changelog/index.tsx`, so upstream #118 is not reverted.

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   - Adds admin UI paths for marketplace-backed Skill and category management. Operators with `skill.read` can view Skill Market / System Skill data; operators with `skill.write` see and can trigger create/edit/delete/upload/reupload/category mutations.
   - Introduces a Skill marketplace axios client parallel to the existing MCP client. It routes through `/market/api/v1`, normalizes marketplace error envelopes, sends `Accept-Language`, and authenticates with the current Octo login token from `useAuthStore.getState().token`. No browser-bundled marketplace admin token is used.
   - Fixes the System Skill reupload path so replacing an existing package commits parsed bytes through `POST /admin/skills/:id/reupload` before issuing a separate metadata `PATCH`. The follow-up metadata update does not carry `version`, so it cannot clobber the version set by the reupload commit.
   - Extends manager capability routing with `skill.read` / `skill.write`, sidebar navigation to `/skill-market`, and direct route access for `/system-skill`.

2. **What could break** because of it (dependents + failure mode)?

   - **Marketplace auth rollout**: octo-marketplace must accept the operator's Octo login token and enforce superAdmin server-side on `/admin/*`. If deployed against an older marketplace expecting `X-Admin-Token`, these pages will 401.
   - **Skill upload contract**: presigned upload/parse/reupload depends on marketplace returning the expected IDs, presigned PUT URL, headers, and parse task shape. If the backend signs a fixed `Content-Type` but omits that header from the presign response, direct PUT may fail with a signature mismatch.
   - **Capability data**: admin profiles must include `skill.read` / `skill.write` for intended users. Missing capabilities hide the sidebar entry or redirect direct URL access even if marketplace would accept the request.
   - **Bundle size**: the added admin surfaces increase the main chunk; build passes, but Vite reports the existing >500 kB chunk warning. Route-level code splitting is a follow-up optimization.
   - **Backend SSRF boundary**: MCP live probe remains a server-side marketplace responsibility. The frontend forwards operator-supplied probe URL/headers; marketplace must validate schemes, private ranges, link-local/metadata IPs, and redirect hops.

3. **How do you know it works** (specific test/repro/trace)?

   - The unit suite passed after the upstream rebase. `src/api/skill.test.ts` verifies `commitAdminSkillReupload('skill-1', ...)` POSTs to `/admin/skills/skill-1/reupload` with `parse_task_id`, `version`, `changelog`, and `tags`.
   - The production build passed after rebase and emitted `dist/assets/index-BcaN4ORT.js`; the marketplace admin-token grep against `dist` returned zero matches.
   - Source/config grep returned zero `VITE_MARKETPLACE_ADMIN_TOKEN`, `MARKETPLACE_ADMIN_TOKEN`, or `X-Admin-Token` matches in the frontend/config surfaces touched by this PR.
   - Manual reviewer smoke plan against a compatible marketplace backend:
     1. Log in as a super admin with `skill.read` / `skill.write`.
     2. Open `/admin/skill-market`; confirm categories and skills load through `/market/api/v1/admin/*` requests carrying the user's login token, not `X-Admin-Token`.
     3. Upload a skill archive; confirm presigned PUT, parse polling, and `POST /admin/skills` create a visible row.
     4. Edit metadata; confirm only metadata changes and the row refreshes.
     5. Open `/admin/system-skill`, edit an existing skill, upload a replacement archive, submit, and confirm the network sequence includes `POST /admin/skills/:id/reupload` before metadata `PATCH`.
     6. Log in without `skill.write`; confirm write controls are hidden and direct backend writes are rejected by marketplace server-side authorization.


Jerry-Xin (dismissed):
Approve. This is a narrower re-cut of the skill-market + system-skill admin work previously approved on `codex/admin-marketplace-surfaces` (the MCP-page portion is excluded here). Every one of the 32 changed files is byte-for-byte identical (blob-verified) to the already-approved revision, so the review below is a re-confirmation rather than a fresh diff.

## Security-critical re-verification (at head 7751ab42)
- **No build-time secret in bundle.** `src/api/skill.ts` sends the caller's runtime Octo login token via `config.headers.token = useAuthStore.getState().token`; there is no `VITE_MARKETPLACE_ADMIN_TOKEN` / `X-Admin-Token` build-time credential baked into the client.
- **API base + proxy consistent.** Client base is `/market/api/v1`; `nginx.conf.template` provides the matching `/market/api/` proxy for prod.
- **401 handling present.** Unauthorized responses redirect to `/admin/login` in both `src/api/skill.ts` and `src/api/mcp.ts`.
- **Skill re-upload flow intact.** `src/api/skill.ts` exposes `reuploadAdminSkill` posting to `POST /admin/skills/{id}/reupload` (with the presigned `reupload/init` step), invoked before metadata update; unit coverage in `src/api/skill.test.ts`.
- **skill.write gating** on the SystemSkill pages and **category sort persistence** logic are unchanged from the approved revision.

## Non-blocking notes (carried over, not merge blockers)
- `/system-skill` has no dedicated sidebar entry.
- Some list views use `setInterval` polling; downloads use `window.open`.
- A few hardcoded Chinese strings remain outside i18n.

## Gate
- `mergeable: true`.
- CI: the `check-sprint` failure is the "no linked issue" governance gate, not a code failure; all code/lint/label checks pass.

Highlights: clean separation of the marketplace client from build-time secrets, presigned re-upload flow with test coverage, and consistent 401 handling across the skill and mcp clients.

Jerry-Xin (dismissed):
Approve. This is a narrower re-cut of the skill-market + system-skill admin work previously approved on this branch (the MCP-page portion is excluded here). Every one of the 32 changed files is byte-for-byte identical (blob-verified) to the already-approved revision, so the review below is a re-confirmation rather than a fresh diff.

## Security-critical re-verification (at head 7751ab42)
- **No build-time secret in bundle.** `src/api/skill.ts` sends the caller's runtime Octo login token via `config.headers.token = useAuthStore.getState().token`; there is no `VITE_MARKETPLACE_ADMIN_TOKEN` / `X-Admin-Token` build-time credential baked into the client.
- **API base + proxy consistent.** Client base is `/market/api/v1`; `nginx.conf.template` provides the matching `/market/api/` proxy for prod.
- **401 handling present.** Unauthorized responses redirect to `/admin/login` in both `src/api/skill.ts` and `src/api/mcp.ts`.
- **Skill re-upload flow intact.** `src/api/skill.ts` exposes `reuploadAdminSkill` posting to `POST /admin/skills/{id}/reupload` (with the presigned `reupload/init` step), invoked before metadata update; unit coverage in `src/api/skill.test.ts`.
- **skill.write gating** on the SystemSkill pages and **category sort persistence** logic are unchanged from the approved revision.

## Non-blocking notes (carried over, not merge blockers)
- `/system-skill` has no dedicated sidebar entry.
- Some list views use `setInterval` polling; downloads use `window.open`.
- A few hardcoded Chinese strings remain outside i18n.

## Gate
- `mergeable: true`.
- CI: the `check-sprint` failure is the "no linked issue" governance gate, not a code failure; all code/lint/label checks pass.

Highlights: clean separation of the marketplace client from build-time secrets, presigned re-upload flow with test coverage, and consistent 401 handling across the skill and mcp clients.

lml2468 (dismissed):
## Review @ `7751ab424b57` — 独立本地复核（非转述）

从 merge-base `c8d72730`(= 已含 #115 的 `main`)的本地 checkout 复核。**结论:APPROVE。**

这是 #116/#117 系列的 **Skill Market + System Skill 部分**,rebase 到已合入 #115 的 main 之上,只携带 Skill 增量 + 共享能力/marketplace-client 集成。System MCP 已在 base(#115),本 PR **未重复**。

### 验证记录（实际执行）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3234 modules
- token 泄露回归:`VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED npm run build` → 打包产物 grep **零命中**
- 全量 `vitest run` — **41 passed**(含新增 `skill.test.ts`);9 处失败全在未触碰的 `src/store/feature.test.ts`,且为我 Node 环境 `localStorage 不可用` 的环境性报错(作者环境有,故其 50/50)—— 非本 PR 回归,不阻塞。

### 关键:sensitive 删除均为「注释-only」,未回退 #115
本 PR 对 `nginx.conf.template`(-1)、`src/api/mcp.ts`(-4)显示删除,逐行核实**全是注释删除**,功能代码完好:
- ✅ nginx `location /market/api/` 反代块**保留**(grep=1),rewrite + 变量 proxy_pass 完好。
- ✅ mcp.ts 的 `401 → logout() + 跳登录` 逻辑**保留**,仅删掉上方注释;auth 仍走 `useAuthStore` 用户 token,无 `ADMIN_TOKEN`/`X-Admin-Token`。

### Gate 1 规格 ✅ / Gate 2 代码质量 ✅（无阻塞）
#116 系列已实证收口的修复,在此 head 全部在位、无回归:
- ✅ **无密钥进 bundle** — `mcp.ts`/`skill.ts` 均用调用者 Octo 登录 token 走 `token` 头;源码零 `ADMIN_TOKEN`/`X-Admin-Token`;注入哨兵构建后产物零命中。
- ✅ **skill.ts 401 处理**(logout + 跳登录),兼容 `err`/`error` 双 envelope。
- ✅ **SystemSkill 写门控** `skill.write` 全覆盖(index/CategoryTab/SkillTable/DetailDrawer/SkillFormModal,25 处 `canWrite`)。
- ✅ **重传换包**:`commitAdminSkillReupload` → `POST /admin/skills/:id/reupload` 先于元数据 PATCH,version 不被 PATCH 覆盖,含单测。
- ✅ **分类排序持久化**:`handleMove` 落库 `sort_order` + 回读。
- ✅ **能力/路由**:`skill.read`/`skill.write` 入 enum,`/skill-market`+`/system-skill` 由 `CapabilityRoute capability="skill.read"` 守卫;`firstManagerPath` 兜底。nav.json 无重复键。

### 🟡 剩余非阻塞(LOW，不 gate merge)
1. **`/system-skill` 有路由无侧栏入口** — `MainLayout` 只 push 了 `/skill-market`,`/system-skill` 仅能直链访问。若刻意保留请忽略;否则建议补侧栏项。
2. **多处硬编码中文**(空态文案等),en-US 界面显示中文。
3. **既有非阻塞**(非回归):`+N` badge 缺 aria-label、magic number 分散、SkillMarket 下载 `window.open`、parse 轮询 `setInterval`。

### 覆盖与盲区(需人工/后端联调)
- 未跑 marketplace 后端:skill 上传→parse→create、reupload、category CRUD/排序回读、download presigned、401→登出全链路需真实环境手测。
- 分类刷新后顺序依赖后端 `GET /admin/skill_categories` 按 `sort_order` 返回;probe SSRF 防护属后端职责 —— 均建议后端确认。
- 部署:需与 octo-marketplace 伴随服务(可达 `${MARKETPLACE_BACKEND}`)+ per-user token 校验链路一同上线。

---

**Verdict: APPROVE。** 干净 rebase,只带 Skill 增量,#115 未被回退(删除均为注释),#116 系列全部修复在位、无回归,剩余全为非阻塞 LOW。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Code Review: octo-admin#119 — feat: add skill market and system skill admin
**Head SHA:** `7751ab424b572fa642e56a3efc4e46c5a1caab7c`
**Merge base:** `c8d727301488d676451d5792fc68fc85b8064bf1`
**Scope:** 20+ files, ~2500 lines added — new Skill Market admin surface + System Skill admin surface, shared API client, nginx/vite proxy, i18n, capability-gated routes.

---

### ✅ Verification Conclusions

1. **Auth model correct** — `skillApi` uses the logged-in user's Octo token via `useAuthStore.getState().token` (`src/api/skill.ts:30-33`). No bundled admin credential shipped to browser. 401 interceptor triggers logout + redirect (`src/api/skill.ts:42-45`). ✓
2. **Capability gating consistent** — Both `/skill-market` and `/system-skill` routes wrapped in `CapabilityRoute capability="skill.read"` (`src/App.tsx:196,210`). Write actions gated by `skill.write` in every component. ✓
3. **Proxy parity** — nginx `/market/api/` strips `/market/` prefix (`nginx.conf.template:57-63`), vite dev proxy does the same (`vite.config.ts:30-34`). Both target port 8092. ✓
4. **i18n key parity** — Verified all 4 new locale files (`systemSkill.json`, `skillMarket.json` × en-US/zh-CN) have matching key sets. No missing translations. ✓
5. **Data flow integrity** — Upload flow (init → presigned PUT → parse → poll → create) traced end-to-end through `uploadAndParseSkillZip` and `SkillFormModal.handleFileUpload`. Response envelopes correctly unwrapped via `normalizeSkill`/`normalizeCategory`. ✓
6. **Category delete safety** — SkillMarket `CategoryTab` handles 409 conflict with `skill_count` display (`src/pages/SkillMarket/CategoryTab.tsx:103-106`). SystemSkill `CategoryTab` disables delete button when `skill_count > 0` (`src/pages/SystemSkill/CategoryTab.tsx:131-135`). Both safe. ✓

---

### ⚠️ Findings (P2)

#### F1 — Polling interval leak on rapid re-upload (P2)
**File:** `src/pages/SystemSkill/SkillFormModal.tsx:113-137`
**Diff-scope:** New code (this PR).
`handleFileUpload` assigns `pollRef.current = setInterval(...)` without clearing a prior interval. If a user uploads file A (poll starts), then while polling is active uploads file B, the old interval reference is overwritten — the old `setInterval` continues running indefinitely, leaking resources and potentially calling `setParseStatus` with stale results that race the new poll.
**Fix:** Add `if (pollRef.current) clearInterval(pollRef.current)` before the new `setInterval` assignment.

#### F2 — Category reorder: partial failure leaves inconsistent sort_order (P2)
**File:** `src/pages/SystemSkill/CategoryTab.tsx:99-110`
**Diff-scope:** New code (this PR).
`handleMove` issues two parallel `updateCategory` calls via `Promise.all`. If one succeeds and the other fails (network blip, 500), the two swapped categories end up with the same `sort_order` value. The subsequent `fetchList()` will show the server state (one moved, one not), which may confuse users.
**Suggestion:** Use sequential calls or a single batch endpoint if the backend supports it. Alternatively, catch per-call and show a more specific error ("moved X but failed to move Y").

#### F3 — Unused `systemSkill` i18n nav key (P2/nit)
**Files:** `src/i18n/locales/en-US/nav.json`, `src/i18n/locales/zh-CN/nav.json`
**Diff-scope:** New code (this PR).
Both nav.json files add a `"systemSkill": "Skill Market"` / `"Skill 市场"` key, but this key is never referenced in `MainLayout.tsx` or any component. The `/system-skill` route exists but has no menu entry — only the `skillMarket` key is used. If this is intentional (hidden route for now), the unused key is dead weight; if the menu entry was accidentally omitted, it should be added.

#### F4 — No upload progress feedback in SkillUploadModal (P2/UX)
**File:** `src/pages/SkillMarket/SkillUploadModal.tsx:40`
**Diff-scope:** New code (this PR).
`uploadAndParseSkillZip(file)` is called without the optional `onProgress` callback. Users see the modal spinner but get no progress bar during the (potentially slow) upload + parse cycle. The SystemSkill `SkillFormModal` does show progress. Inconsistent UX between the two surfaces.

---

### 🔍 Data Flow Tracing

| Consumer | Upstream Source | Verified |
|---|---|---|
| `SkillTab` rows | `listAdminSkills` → `GET /admin/skills` → `resp.data.data` → `normalizeSkill` | ✓ |
| `SkillTab` categories | `listSkillCategories` → `GET /admin/skill_categories` → `resp.data.data` → `normalizeCategory` | ✓ |
| `SkillFormModal` parse pre-fill | `getParseStatus` → `pollAdminParseTask` → `resp.data.data.result.*` | ✓ |
| `SkillFormModal` submit (create) | `parseTaskId` from poll → `createSkill` → `POST /admin/skills { parse_task_id }` | ✓ |
| `SkillFormModal` submit (edit) | `parseTaskId` → `commitAdminSkillReupload` → `POST /admin/skills/:id/reupload` then `updateSkill` | ✓ |
| Token injection | `useAuthStore.getState().token` → request interceptor → `config.headers.token` | ✓ |
| Download URL | `getAdminSkillDownloadUrl` → `GET /admin/skills/:id/download?format=json` → `resp.data.data.download_url` → `window.open` | ✓ |
| Category filter in SkillTab | `listSkillCategories` → `setCategories` → `Select.Option` → `categoryFilter` state → `listAdminSkills({ category_id })` | ✓ |

---

### 🛡️ Blind Spot Checklist

| ID | Check | Result |
|---|---|---|
| C1 | Dual-path parity (SkillMarket ↔ SystemSkill) | **Clear** — Both use same `skillApi` client, same capability gates (`skill.read`/`skill.write`), same auth model. Different API endpoints (`/admin/skills` vs same) are correct for their respective contexts. |
| C2 | Control-flow ordering / nested reuse | **Clear** — No shared logic is called in conflicting order. `normalizeSkill`/`normalizeCategory` are pure functions called once per response. |
| C3 | Authorization boundary ≠ capability boundary | **Clear** — Frontend capability check is UX-only; real auth is server-side via token verification in marketplace backend. No endpoint is exposed without server-side auth. |
| C4 | Authorization lifecycle / container-member cascade | **N/A** — No container/member hierarchy in this PR. |
| C5 | Build/note parity | **Clear** — `Dockerfile` uses `npm ci` with lockfile, nginx proxy matches vite proxy, `MARKETPLACE_BACKEND` env var has sensible default (`octo-marketplace:8092`). |
| C6 | Governance doc self-consistency | **N/A** — No governance/policy docs changed. |

---

### 📋 Cross-Round Blocker Review

---

### Summary

Solid implementation. The dual-surface architecture (SkillMarket for marketplace admin, SystemSkill for system-level skills) is well-structured with consistent auth, capability gating, and i18n. The presigned upload flow is correctly implemented. Four P2 findings — none block production safety, but F1 (polling leak) is worth fixing before merge as it's a one-line change.

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. P2 items are code-quality and UX improvements that can be addressed in follow-up.

yujiawei (approved):
# Code Review — PR #119 (octo-admin)

Reviewed at head `7751ab4` against `main` (merge-base `c8d7273`). Build (`tsc && vite build`) is green and the full test suite passes (**50/50**, including the reupload contract test). This PR adds the Skill Market admin surface (nav-linked at `/skill-market`) and the legacy System Skill surface (direct-URL `/system-skill`), both talking to octo-marketplace via `/market/api/v1` using the operator's own Octo login token. Classified security-sensitive; reviewed accordingly.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Skill Market (categories + skills tabs, upload/edit/detail/download), System Skill admin, capability gating (`skill.read`/`skill.write`), i18n (en/zh), routing, and nav entry are all present and wired.
- **Extra:** none out of scope. The `mcp.ts` change is a comment-only removal; `vite.config.ts` adds an opt-in `VITE_PORT` override; `api/index.ts` extends `ApiError` with an optional `details` field consumed by the new category-in-use error path — all in direct support of the feature.
- **Deviation:** none. Auth model matches the established contract — no marketplace admin secret is shipped to the browser (verified: no `ADMIN_TOKEN`/`X-Admin-Token`/`dev-admin-token` anywhere in `src/` or `.env.example`); presigned uploads/downloads go through an interceptor-free client so the session token is never leaked to the storage host.

## 2. Code quality

**Quality: Approved**

Security crown-jewels verified clean:
- **No credential leakage to storage origins.** `uploadToPresigned` uses the bare `axios` default instance (no request interceptor) and `uploadAndParseSkillZip` uses `fetch` — neither attaches the `token`/`Accept-Language` headers that the authenticated `skillApi` instance adds. Presigned PUT/GET therefore carry no session token.
- **No XSS.** Skill README / SKILL.md is rendered as text inside a `<pre>` (React-escaped); no `dangerouslySetInnerHTML`, no raw HTML/markdown injection. `window.open(url, '_blank')` is only ever called on a server-issued presigned download URL.
- **Capability gating is comprehensive.** `skill.read`/`skill.write` are declared in `capabilities.ts`; every route is wrapped in `CapabilityRoute`; every mutating control is gated on `canWrite`, with handler-level `if (!canWrite) return` backstops. 401 → `logout()` + redirect in both API clients.

Non-blocking findings (P2 — fine to address in a follow-up, not merge-blocking):

- **P2 — `uploadToPresigned` hardcodes `axios.put`** (`src/api/skill.ts:537`) and `SkillFormModal` (`SystemSkill/SkillFormModal.tsx:121`) discards the `method` returned by the upload-init response. Works today because the backend issues PUT presigns; would break silently if the presign contract ever specifies another verb. Consider threading `method` through.
- **P2 — `icon_key` dropped in the category compatibility wrappers** (`src/api/skill.ts:589-607`). `createCategory`/`updateCategory` accept `icon_key` and `SystemSkill/CategoryTab.tsx:64` passes `'MoreHorizontal'`, but it is never forwarded to the API — the icon selection is a silent no-op on that legacy surface.
- **P2 — `SkillMarket/SkillDetailDrawer` has no stale-response guard** (`src/pages/SkillMarket/SkillDetailDrawer.tsx:20-39`). Rapidly switching rows can let an earlier `getAdminSkill` resolve after a later one and display stale detail. The analogous `SystemSkill/DetailDrawer` already guards this with a `stale` flag — mirror that pattern here.
- **P2 — List loads race** (`SkillMarket/SkillTab.tsx:45-66`, `SkillTable` in the System Skill page): concurrent search/filter/page requests are not cancelled or request-id checked, so a slower earlier response can overwrite the current one.
- **P2 — Poll interval can be orphaned** (`SystemSkill/SkillFormModal.tsx:127`): re-selecting a package while a previous parse poll is running assigns a new `pollRef.current` without clearing the previous interval; the old timer keeps firing until unmount/close. Clear `pollRef.current` at the top of `handleFileUpload` before starting a new poll.
- **P2 — `bumpPatch` no-ops on short versions** (`SystemSkill/SkillFormModal.tsx:33-39`): a `1.0` or `2` version returns un-bumped, which can collide with a backend version-uniqueness constraint on reupload. Minor; most packages carry full semver.
- **P2 (nit) — Duplicate nav label.** `nav.json` defines both `skillMarket` and `systemSkill` as `"Skill Market"`. Only `/skill-market` is nav-linked (`MainLayout.tsx`); `/system-skill` is direct-URL only, so there is no visible duplicate today, but the identical label is confusing — consider distinguishing them.
- **P2 — `parseTaskByUploadId` module-global `Map` is never cleared** (`src/api/skill.ts:214`). Bounded by uploads-per-session, but grows monotonically. Low impact.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. No P0/P1 issues: no auth/capability bypass, no token leakage to storage hosts, no XSS or SSRF, and no silent-no-op in the load-bearing create/reupload/edit paths (the reupload commit path correctly calls the dedicated `/reupload` endpoint and is covered by a unit test). The findings above are all P2 polish items.

## 4. Suggestions

- Add the `stale`-guard from `SystemSkill/DetailDrawer` to `SkillMarket/SkillDetailDrawer`, and clear `pollRef` before starting a new parse poll — these are the two most user-visible of the P2s.
- Thread the presigned `method` through instead of hardcoding PUT, to keep the client honest against the storage contract.

## 5. Additional notes (for manual verification)

Because this console authenticates with the operator's own token and relies on the marketplace to enforce authorization, a human should confirm on the backend that: (a) the `/admin/skills*` and `/admin/skill_categories*` endpoints enforce `role=superAdmin` server-side; and (b) a reupload `parse_task_id` is bound to its target skill server-side (a parse task for skill B cannot be committed against skill A). These cannot be settled from this frontend-only diff.

yujiawei (dismissed):
# Code Review — PR #119 (octo-admin), re-review at head 77af6912

Re-reviewing at head `77af6912` (my prior Approve was at `7751ab4`; a formal verdict does not carry across a new push, so this re-anchors it). Delta since the reviewed head is a single commit — `fix(skill): align admin upload endpoints` — touching only `src/api/skill.ts` (+43/-17) and `src/api/skill.test.ts`. Build (`tsc && vite build`) is green and the suite passes **52/52** (2 new tests added for the changed functions).

## 1. Spec compliance

**Spec: ✅**

The delta is a contract-alignment fix, in scope for the feature:
- `initReupload` no longer calls `/admin/skills/:id/reupload/init`; it reuses the shared `initAdminSkillUpload` (`/admin/skill_uploads`). The target-skill binding still happens server-side at the `/admin/skills/:id/reupload` commit step (unchanged), so this is the same trust model as first-time creation — correct, and now covered by a unit test.
- `uploadIcon` switches from a multipart `POST /admin/skill_icons` to a presigned flow: `POST /skill_icon_uploads` → `fetch` PUT to the returned presigned URL. Covered by a new test.

No new feature, no scope creep, no deviation from the established auth model.

## 2. Code quality

**Quality: Approved**

- The new `uploadIcon` presigned PUT uses bare `fetch` (not the authenticated `skillApi` instance), so the session token is **not** leaked to the storage host — consistent with the skill-package upload path and an improvement over the prior multipart-through-`skillApi` approach. ✓
- `initReupload`'s `skillId` param is now unused (renamed `_skillId`) but kept for call-site compatibility — acceptable.
- All prior crown-jewel verifications still hold at this head: no bundled admin token, interceptor-free presigned transfers, README rendered as escaped text (no XSS), comprehensive `skill.read`/`skill.write` gating, 401 → logout+redirect.

Non-blocking (P2):
- **P2 — icon endpoint path inconsistency.** `uploadIcon` posts to `/skill_icon_uploads`, while every other admin call is under `/admin/*` (`/admin/skills`, `/admin/skill_uploads`, `/admin/skill_categories`). The request still carries the user token (marketplace enforces authorization server-side), so this is not a security gap, but the missing `/admin/` prefix is worth a second look against the backend route table — confirm it's the intended admin icon-upload endpoint and not an unauthenticated/public one.
- Carry-over P2s from head `7751ab4` (unchanged, still non-blocking): orphaned poll interval on rapid re-select (`SystemSkill/SkillFormModal.tsx`), `SkillMarket/SkillDetailDrawer` missing the stale-response guard its `SystemSkill` sibling has, list-load races, `bumpPatch` no-op on short versions, duplicate `skillMarket`/`systemSkill` nav label, no progress feedback in `SkillMarket/SkillUploadModal` (UX inconsistency vs `SystemSkill/SkillFormModal`).

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved at head `77af6912`. No P0/P1: the endpoint-alignment delta is clean, adds no new security surface (icon upload is now interceptor-free presigned), and is test-covered. Remaining items are all P2 polish.

## 4. Additional notes (for manual verification)

Two backend-contract items a human should confirm (cannot be settled from this frontend-only diff): (a) `/skill_icon_uploads` is the intended admin icon-upload route and enforces `role=superAdmin` like the `/admin/*` endpoints; (b) the `/admin/skills/:id/reupload` commit binds the `parse_task_id` to its target skill server-side (a parse task for skill B cannot be committed against skill A) — this is the safety property that lets `initReupload` share the generic upload-init endpoint.

lml2468 (dismissed):
## Re-review @ `fbc6d2aba9a8` — 独立本地复核（非转述）

我此前在 `7751ab42` 已 APPROVE。本轮新增两条 commit(`440be6c align admin upload endpoints`、`fbc6d2a address admin skill ui regressions`),针对 peer 提的若干 nit 做加固。**结论:维持 APPROVE,质量净提升。**

### 真实增量（`7751ab42..fbc6d2ab`,4 文件 +173/-31）
1. **presigned PUT 增加 URL 安全校验** — 新 `assertSafeExternalUrl`:仅允许 `https:`,或 `http://localhost|127.0.0.1`(dev),其余 scheme/非法 URL 抛 `invalid_response` 502。所有 presigned PUT(`uploadAndParseSkillZip`/`uploadToPresigned`/`uploadIcon`)统一走 `putPresignedFile`,防止把文件 PUT 到非预期 scheme/目标。✅ 客户端防御纵深(真实 SSRF 边界仍在后端签发端)。
2. **upload 端点对齐** — `initReupload` 改用 marketplace 实际暴露的 `/admin/skill_uploads`(复用首传 init),reupload commit 时再绑定 parse task 到目标 skill;`uploadIcon` 改为 `/skill_icon_uploads` init → presigned PUT → 持久化 object_key(镜像 octo-web 图标流),含 malformed-response 前置校验。
3. **分类排序修复** — `updateCategory` 调用补 `name`(与 `sort_order` 一起传),避免 PATCH 只带 sort_order 时后端可能要求 name 的问题。
4. **SkillTab 行操作冒泡** — 操作列 `<Space onClick={stopPropagation}>`,修掉点操作按钮误触发行详情抽屉(peer 提及)。

### 验证记录（本轮实跑 @ 新 head）
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3234 modules
- `skill.test.ts` — ✅ **5 passed**(新增 initReupload / uploadIcon happy / malformed-response / unsafe-scheme 覆盖)
- 全量 `vitest run` — **45 passed**(较上轮 +4);9 处失败仍全在未触碰的 `src/store/feature.test.ts`(既有 localStorage 环境报错,不阻塞)

### 无回归(承接 `7751ab42` 完整审查)
安全关键路径保持:用户 token 鉴权、无 bundle 密钥(源码零 `ADMIN_TOKEN`/`X-Admin-Token`)、nginx `/market/api/` 反代、401→登出、重传换包端点、写门控。本轮改动均为加固,未触碰这些逻辑。

### 🟡 剩余非阻塞(LOW，不 gate)
1. **`uploadToPresigned` 丢失上传中进度粒度** — 原 axios `onUploadProgress` 逐帧回调改为 `fetch` 后,只剩 30→60 两跳。属 UX 细节回退(fetch 无原生上传进度),可接受;若要精细进度需 XHR。调用方 `SkillFormModal` 的进度条会从 30 直接跳 60。
2. **`assertSafeExternalUrl` 仅放行 localhost/127.0.0.1 的 http** — 若某些部署的 dev presigned 用 docker 服务名(如 `http://minio:9000`)会被拒;prod https 不受影响。建议确认 dev/staging 的 presigned host 落在放行集内。
3. carryover:`/system-skill` 无侧栏入口、硬编码中文、`+N` badge 缺 aria-label。

### 覆盖与盲区
- 未跑 marketplace 后端:upload→parse→create、reupload、icon 上传、category 排序回读全链路需真实环境手测。probe SSRF 仍属后端职责。

---

**Verdict: APPROVE。** 两条 commit 为纯加固(URL 校验 / 端点对齐 / 排序修复 / 冒泡修复)+ 扩测,无回归,剩余全为非阻塞 LOW。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #119 (octo-admin)

## Summary

This PR adds two admin surfaces for skill management: a **Skill Market** page (`/skill-market`) for managing public skill categories and listings against the octo-marketplace backend, and a **System Skill** page (`/system-skill`) for system-level skill CRUD with upload/reupload flows. The API layer (`src/api/skill.ts`) creates a dedicated axios instance pointed at `/market/api/v1` with its own response envelope, while reusing the operator's Octo login token for auth. Capability gating (`skill.read` / `skill.write`) is added to the auth system and wired into routes and UI write controls.

The security posture is solid: presigned URL uploads validate protocol via `assertSafeExternalUrl`, auth interceptors mirror the shared client pattern, and all routes are capability-gated. Five P2 items and one nit below — no production-blocking defects.

## Verification

- ✅ **Auth boundary** — both `/skill-market` and `/system-skill` routes use `CapabilityRoute` with `skill.read`; write operations in all four UI components check `canWrite` via `skill.write`. Auth interceptor in `src/api/skill.ts:28-33` attaches the user's Octo token, matching the shared client pattern in `src/api/index.ts`.
- ✅ **Presigned URL safety** — `assertSafeExternalUrl` at `src/api/skill.ts:276-290` rejects non-HTTPS URLs (except localhost HTTP for dev). Tests at `src/api/skill.test.ts:133-148` verify `javascript:` URLs are blocked before PUT.
- ✅ **nginx proxy unchanged** — `/market/api/` reverse proxy to `MARKETPLACE_BACKEND` is unchanged; only comments were cleaned up. No new exposure paths.
- ✅ **Data normalization** — `normalizeCategory` and `normalizeSkill` at `src/api/skill.ts:175-201` defensively handle dual-ID fields (`skill_category_id`/`id`, `skill_id`/`id`) and default all optional fields.
- ✅ **401 handling** — `skillApi` response interceptor at `src/api/skill.ts:47-49` logs out and redirects to `/admin/login` on 401, consistent with `src/api/index.ts` and `src/api/mcp.ts`.
- ✅ **Test coverage** — `src/api/skill.test.ts` covers `commitAdminSkillReupload`, `initReupload`, and `uploadIcon` including error paths (malformed response, unsafe URL scheme).

Static analysis only at head `fbc6d2ab`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Five P2 items and one nit below.

### P2 — Orphan route `/system-skill` duplicates `/skill-market` (`src/App.tsx:249`)

`App.tsx:249` registers a `/system-skill` route gated by `skill.read`, identical to the `/skill-market` route at `:221`. However, `MainLayout.tsx:128-130` only adds `/skill-market` to the sidebar — `/system-skill` is invisible but reachable via direct URL. Both `nav.json` locale files define `systemSkill` with the same label as `skillMarket` ("Skill Market" / "Skill 市場"), confirming this is a legacy alias.

Two URLs serving the same admin surface increases operator confusion and attack surface. Either wire `/system-skill` into the navigation with a distinct purpose, or remove the route and the unused `nav:systemSkill` i18n keys.

### P2 — Compatibility wrappers silently drop `icon_key` (`src/api/skill.ts:635`)

`createCategory` at `:630-639` accepts `{ name: string; icon_key: string }` but only forwards `{ name, sort_order: 0 }` to `createSkillCategory` — `icon_key` is never sent. The caller at `src/pages/SystemSkill/CategoryTab.tsx:64` passes `icon_key: 'MoreHorizontal'`, which is silently discarded. `updateCategory` at `:641-649` has the same pattern.

The type signature advertises `icon_key` as valid but the implementation ignores it. Either remove `icon_key` from the wrapper's parameter type, or forward it to the underlying call.

### P2 — Presigned upload has no timeout (`src/api/skill.ts:292`)

`putPresignedFile` at `:292-306` uses raw `fetch()` to PUT file bytes to a presigned URL with no AbortController and no timeout. If the object-storage endpoint stalls, the upload hangs indefinitely while the UI shows a progress indicator. The companion `uploadAndParseSkillZip` at `:451` caps parse polling at 120 seconds, but the PUT step itself is unbounded.

Add a default timeout via AbortController (e.g. 5 minutes for a 20 MB package), or accept a `signal` parameter so the modal's cancel path can abort the in-flight upload.

### P2 — Optimistic reorder with partial-failure divergence (`src/pages/SystemSkill/CategoryTab.tsx:88`)

`handleMove` at `:88-106` applies `setData(next)` immediately (optimistic reorder), then fires two parallel `updateCategory` calls with swapped `sort_order` values. If one succeeds and the other fails, the server ends up with a half-swapped sort order. The `catch` handler calls `fetchList()` to re-sync, but the resulting state may differ from both the optimistic and the original order.

Consider making the updates sequential, or batching into a single reorder API call if the backend supports it. At minimum, defer `setData` until both calls resolve.

### P2 — Parse-task polling has no upper bound (`src/pages/SystemSkill/SkillFormModal.tsx:117`)

The `setInterval` at `:117-143` polls `getParseStatus` every 2 seconds with no max-attempt limit. If the server-side parse task stays in `pending` or `parsing` indefinitely, this modal polls forever — leaking intervals and keeping the UI in a perpetual uploading state. Contrast with `uploadAndParseSkillZip` in `src/api/skill.ts:451` which caps at `maxAttempts = 60`.

Add a similar cap: after N polls with no status change, stop the interval and surface a timeout error.

### Nit — Dead `nav:systemSkill` i18n key (`src/i18n/locales/en-US/nav.json:15`)

Both locale files define `systemSkill` with the same label as `skillMarket`. `MainLayout.tsx` only references `nav:skillMarket`. Remove the dead key from both files.

## Human-verify

1. **Server-side auth enforcement** — the client-side capability gating is correct, but the review cannot verify that octo-marketplace's `/admin/*` endpoints independently validate the caller's Octo token and enforce superAdmin role (as documented in the `skill.ts` file header). This is **not** a merge blocker — flagging for cross-team confirmation.

2. **`/system-skill` vs `/skill-market` intent** — both routes exist in `App.tsx` but only one is wired to the sidebar. Confirm whether `/system-skill` is intentionally kept as a separate admin surface (perhaps for a future split) or should be removed. This is **not** a merge blocker.

- Dual-path parity (C1): Both `/skill-market` and `/system-skill` use identical `skill.read` capability gating. Both SkillMarket and SystemSkill UI components check `skill.write` for all write operations.
- Auth interceptor parity: `skillApi` in `skill.ts` mirrors the shared `api` instance — same token attachment, same 401 logout/redirect, same Accept-Language header.
- nginx proxy: `/market/api/` location is unchanged; only comments were cleaned. No new exposure paths.
- Dockerfile: `MARKETPLACE_BACKEND` env default and `API_BACKEND` are unchanged. Comment cleanup only.
- README additions correctly document the dual-proxy architecture and warn against shipping secrets as `VITE_*` variables — consistent with Dockerfile and nginx config.
- i18n: All new locale keys (`skillMarket`, `systemSkill`) are properly registered in `src/i18n/index.ts` for both `en-US` and `zh-CN`.
- `vite.config.ts`: `VITE_PORT` change is dev-server only, no production impact.

## Verdict: APPROVED

No production-blocking defects. The five P2 items are quality improvements — orphan route cleanup, honest API type contracts, upload timeout, optimistic-update safety, and polling bounds. None make a working path unavailable or produce user-visible wrong data. The security posture is solid for a new admin surface: proper auth gating, presigned URL validation, and consistent interceptor patterns.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Five P2 quality items and one nit. Auth boundary, presigned URL safety, and capability gating all verified correct. The orphan `/system-skill` route and missing upload timeout are worth addressing but not merge-blocking.

Jerry-Xin (dismissed):
APPROVE — re-review at head fbc6d2ab.

Verified the true delta since the prior approved head (7751ab42): compare reports `status: ahead`, ahead_by 2 / behind_by 0 (no rebase), touching only 4 files across 2 commits (`fix(skill): align admin upload endpoints`, `fix(skill): address admin skill ui regressions`). Reviewed the actual patches, not commit messages.

### What changed (byte-verified)
- `src/api/skill.ts`: introduced `assertSafeExternalUrl()` + `putPresignedFile()` helpers. Presigned PUT targets are now validated to be `https:` (or `http:` only for localhost/127.0.0.1) before any upload, rejecting `javascript:`/other schemes with an `invalid_response` ApiError. `initReupload` now initializes via `/admin/skill_uploads` (the endpoint marketplace exposes; the reupload commit binds the parsed task to the target skill server-side). `uploadIcon` moved from a multipart POST to `/admin/skill_icons` to an init (`/skill_icon_uploads`) + presigned-PUT flow, with a required-field guard before PUT. `uploadToPresigned`/`uploadAndParseSkillZip` refactored onto the shared helper.
- `src/api/skill.test.ts`: added coverage for `initReupload`, `uploadIcon` (happy path, malformed-response rejection, unsafe-scheme rejection before PUT).
- `src/pages/SystemSkill/CategoryTab.tsx`: reorder now forwards `name` alongside `sort_order` on the paired `updateCategory` calls.
- `src/pages/SkillMarket/SkillTab.tsx`: action-column `Space` stops row-click propagation so action buttons don't trigger the row detail handler.

### Peer notes assessed
- Category update / reorder name loss: fixed. `updateCategory` now forwards `name` with `sort_order`. Backend `updateSkillCategory` uses HTTP PATCH (partial update), so unsent fields are not wiped; forwarding `name` is a correct, defensive belt-and-suspenders and resolves the concern.
- icon_key not forwarded on category update: not a defect on this surface. There is no icon picker in the category form; `icon_key` is backend-defaulted, and PATCH partial semantics mean omission does not erase it.
- Polling interval not cleared on fast re-upload / package switch: not present at this head. No `setInterval`/`setTimeout` polling exists in the SkillMarket or SystemSkill components; nothing to leak.
- Stale-response guard on detail drawer, list request cancellation, and generic presigned method handling: non-blocking polish for an admin-only surface; none causes data corruption. The presigned helper now honors the server-provided `method` (defaulting to PUT) rather than hardcoding it.
- The `uploadToPresigned` refactor drops the granular `onUploadProgress` (30 to 60 step now jumps) — cosmetic only, non-blocking.

### Regression check on security-critical paths (all intact)
- User-token auth: request interceptor sets `config.headers.token` from the auth store (caller's own Octo login token). No admin credential is shipped to the browser; grep confirms no `X-Admin-Token` and no `VITE_*_TOKEN` in src.
- Marketplace proxy: base remains `/market/api/v1` (env override only), so requests continue through the intended proxy.
- 401 handling: interceptor logs out and redirects to `/admin/login`. Unchanged.
- Reupload endpoint: `POST /admin/skills/:id/reupload` (commit step) unchanged; only the init step was realigned to the exposed admin upload endpoint.
- write gating: `canWrite` still guards all category create/update/delete/reorder actions.
- New SSRF/scheme guard on presigned URLs is a net security improvement and is unit-tested.

No blocking issues. Nice hardening on the upload path.

yujiawei (dismissed):
# Code Review — PR #119 (octo-admin)

Reviewed at head SHA `fbc6d2aba9a86a04ccbf5866d058c9435c79c0de` against merge-base `c8d7273`. This is a security-sensitive change (marketplace admin auth, presigned uploads), so I verified the auth/token boundary and upload flow directly, and ran the test/build locally.

## Verification performed
- `pnpm test` → 54 tests pass (incl. `src/api/skill.test.ts`).
- `pnpm build` → passes; emits one large chunk (existing >500 kB warning, not introduced here).
- `rg VITE_MARKETPLACE_ADMIN_TOKEN|MARKETPLACE_ADMIN_TOKEN|X-Admin-Token` across `src/` and the built `dist/` → **zero matches**. No marketplace admin secret is compiled into the browser bundle.

## Security assessment (positive findings)
- **Login token is not leaked to the external presigned target.** `skillApi`'s request interceptor injects the Octo login token as `config.headers.token` (`src/api/skill.ts:29`), but the presigned PUT uses raw `fetch` with only backend-supplied headers (`src/api/skill.ts:394`). The axios interceptor never runs on that request, so the token is confined to the marketplace origin. Correct.
- **The presigned-URL scheme guard is applied to every PUT path.** `assertSafeExternalUrl` runs at the top of `putPresignedFile` (`src/api/skill.ts:393`), covering skill-zip upload (`:476`), the generic path (`:568`), and icon upload (`:624`). No unguarded PUT path exists. `javascript:`/`file:` schemes are rejected before any network call (covered by a unit test).
- **Capability gating is consistent.** Both routes require `skill.read` (`src/App.tsx`), all write controls are gated on `skill.write` (`canWrite`), and System Skill mutation handlers re-check `canWrite` inside the handler (`SkillFormModal.tsx:174`). This is client-side defense-in-depth/UX — the real authorization boundary is the marketplace `/admin/*` superAdmin check (see human-review note).
- **Reupload cannot clobber the committed version.** `commitAdminSkillReupload` carries `version` (`SkillFormModal.tsx:185`); the follow-up metadata `updateSkill` deliberately omits `version` (`SkillFormModal.tsx:192`), so the reupload-set version is preserved. Correct — and covered by a unit test.
- No `dangerouslySetInnerHTML` anywhere; `SKILL.md` is rendered as text in a `<pre>`, so no stored-XSS surface from parsed package content.

## 1. Spec compliance verdict
**Spec: ✅**
- Missing: none — the diff delivers the Skill Market and System Skill admin surfaces, the `skill.read`/`skill.write` capabilities, route guards, sidebar entry, marketplace client, i18n, and config/docs described in the PR body.
- Over-built: none — no unrequested flags, endpoints, or backdoors. System Skill is intentionally route-only (no sidebar entry), matching its "legacy/system-oriented" description.
- Divergence: none — auth model (operator login token, server-side superAdmin) matches the stated contract; no shared admin secret is shipped.

## 2. Code quality verdict
**Quality: Approved** — no P0/P1. The items below are P2/nit; none blocks merge, but the first is worth a fast follow-up.

### P2 — `uploading` can get stuck, disabling the upload control after reopen
`src/pages/SystemSkill/SkillFormModal.tsx:74-99`, `:127`
If the operator closes the modal while a package is still parsing, the cleanup clears the poll interval but never resets `uploading`. The reopen effect resets `parseTaskId`/`selectedFileName`/etc. but not `uploading`, so the `Dragger` stays `disabled={uploading}` and polling never resumes — recoverable only by a full page reload. Suggest adding `setUploading(false)` to the reopen (`if (open)`) branch.

### P2 — parse-status polling can overlap and reorder
`src/pages/SystemSkill/SkillFormModal.tsx:127`
The `setInterval(async …)` callback fires every 2s regardless of whether the previous `getParseStatus` resolved. If a call takes >2s, responses can arrive out of order and a late `pending` can overwrite a `success` state. Consider a self-scheduling `setTimeout` loop or an in-flight guard.

### P2 — stale list responses can overwrite newer filter/page selections
`src/pages/SystemSkill/SkillTable.tsx:55-70`
Rapid search/category/page changes issue concurrent `listSkills` calls with no cancellation or generation check; an older, slower response can replace the latest results. A request-generation counter (or AbortController) would fix it.

### P2 — recording the next cursor triggers a duplicate fetch for the current page
`src/pages/SystemSkill/SkillTable.tsx:63-73`
`cursors` is both mutated inside `fetchList` and listed in its dependency array, so appending `next_cursor` recreates the callback and re-runs the effect, fetching the same page twice. Keep the cursor cache in a ref, or drop `cursors` from the deps.

### P2 (hardening) — `window.open(url, '_blank')` without `noopener`
`src/pages/SkillMarket/SkillTab.tsx:103`, `src/pages/SkillMarket/SkillDetailDrawer.tsx:128`
Opening a backend-supplied download URL without `noopener,noreferrer` exposes the admin tab to reverse-tabnabbing in older browsers. Prefer `window.open(url, '_blank', 'noopener,noreferrer')`.

### Nits (non-blocking)
- `src/i18n/locales/*/nav.json`: `skillMarket` and `systemSkill` are both labeled "Skill Market" — the duplicate label is confusing if System Skill ever gets its own nav entry.
- `.env.example` dropped the comment explaining the login-token/`/v1/auth/verify` auth model. It's partly relocated to the README, but `.env.example` is the primary deploy-time discovery point; a one-line pointer there would reduce the risk of someone reintroducing a shared secret.

## 3. Human-review items (security-sensitive)
These are backend/runtime boundaries this frontend PR cannot prove on its own — please confirm before/at merge:
- **Server-side authorization:** octo-marketplace must verify the operator's Octo token via octo-server and enforce `role=superAdmin` on **every** `/admin/*` skill/category/upload/reupload endpoint. The client capability gate is UX only and is trivially bypassable via direct API calls.
- **Presigned-URL trust:** the marketplace must only ever return presigned URLs pointing at approved object storage. The frontend's HTTPS scheme check does not (and cannot) constrain the destination host, so an internal/attacker-controlled HTTPS host in a compromised response would still receive the PUT.
- **Deploy target compatibility:** these pages authenticate with the login token, not `X-Admin-Token`; deploying against an older marketplace that expects `X-Admin-Token` will 401.

## Advisory coverage
This review absorbed two independent advisory passes in addition to my own. A cross-family long-context pass flagged (a) that the docs warn against `VITE_*` secrets without code removing them and (b) that variable-based nginx `proxy_pass` needs a `resolver`; both were checked and **do not apply** — no such `VITE_*` variable exists in source or the built bundle, and `nginx.conf.template:7` already declares `resolver 127.0.0.11`. A line-level adversarial pass independently confirmed the token-boundary, guard-coverage, capability-gating, and reupload-version findings above, and surfaced the four P2 correctness items. Neither advisory could verify backend authorization, presigned-host ownership, or CORS without server access — see human-review items.

## Verdict
No P0/P1 issues; spec fully implemented with no over-build. **Approving.** Please treat the stuck-`uploading` P2 as a near-term follow-up and have a human confirm the three backend security items above (the PR already carries `needs-human-review`).

lml2468 (approved):
## Re-review @ `737a51bf730a` — 独立本地复核（非转述）

我此前在 `fbc6d2ab` 已 APPROVE。本轮唯一新增 `737a51bf chore(deps): bump vitest/vite/glob/esbuild to patch security advisories`。**结论:维持 APPROVE。**

### 真实增量（`fbc6d2ab..737a51bf`）
仅 3 文件:`package.json` + 两个 lockfile。**只动 devDependencies + overrides,无任何 runtime 依赖变更**:
- `vitest` `^2.1.9 → ^3.2.6`(major)
- `vite` `^6.2.0 → ^6.4.3`
- `@vitest/coverage-v8` `^2.1.9 → ^3.2.6`
- 新增 `overrides: { glob: ^10.5.0 }`

均为构建/测试工具链,不进产物;应用运行时依赖零改动。

### 验证记录（本轮实跑 @ 新 head，全新 `npm install`）
- `vitest --version` — 实际拉到 **3.2.7**,确认 major 升级生效
- `tsc --noEmit` — ✅ 0 错误
- `npm run build` — ✅ 3234 modules,产物正常
- token 泄露回归:`VITE_MARKETPLACE_ADMIN_TOKEN=PROVE_ME_LEAKED npm run build` → 打包产物**零命中**
- 全量 `vitest run`(在 v3 下) — **45 passed**,与升级前一致;9 处失败仍全在未触碰的 `src/store/feature.test.ts`(既有 localStorage 环境报错,vitest 3 下行为不变,不阻塞)
- `skill.test.ts` — ✅ 通过(v3 下)
- `npm audit` — **found 0 vulnerabilities**

### 结论(承接 `fbc6d2ab` 完整审查,无回归)
vitest major 升级未破坏测试:同样 45 passed、同一批 pre-existing 失败,行为一致。所有安全关键路径(用户 token 鉴权、无 bundle 密钥、nginx `/market/api/` 反代、401→登出、重传换包端点、写门控、上轮新增的 presigned URL 校验)保持不变。纯工具链安全补丁,净正向。

### 🟡 剩余非阻塞(carryover,不 gate)
- `uploadToPresigned` 上传进度粒度(30→60 跳);`assertSafeExternalUrl` 仅放行 localhost 的 http(docker 服务名 host 需确认);`/system-skill` 无侧栏入口;硬编码中文;`+N` badge 缺 aria-label。

### 覆盖与盲区
- devDep 升级的完整回归以 CI 全绿为准(我本地已跑通 tsc/build/unit);marketplace 后端全链路仍需真实环境手测。

---

**Verdict: APPROVE。** 纯 devDep 安全升级,runtime 零改动,vitest 3 下测试无回归,`npm audit` 干净。

Jerry-Xin (approved):
## Re-review — APPROVE ✅

Re-reviewed the incremental delta from the previously-approved head `fbc6d2ab` to the new head `737a51bf`. The new head is exactly **1 commit ahead** (linear, no rebase): `chore(deps): bump vitest/vite/glob/esbuild to patch security advisories`.

### Delta scope (byte-verified)
- **Only** `package.json`, `package-lock.json`, and `pnpm-lock.yaml` changed.
- `git diff fbc6d2ab..737a51bf -- src/` is **empty** — no runtime/source code changed since the approved head.
- Version moves (all devDependencies / build tooling):
  - `vite` `^6.2.0` → `^6.4.3`
  - `vitest` + `@vitest/coverage-v8` `^2.1.9` → `^3.2.6`
  - `esbuild` resolved to `0.25.12` (>= the fixed line for the dev-server request advisory)
  - `glob` pinned via new `overrides` to `^10.5.0` (resolved flat at `10.5.0`)

### Verification
- Lockfile/manifest sync: `npm install --package-lock-only --dry-run` reports **"up to date"** — `npm ci` would succeed; no desync.
- The `overrides` block in `package.json` is not mirrored as a top-level `overrides` key in `package-lock.json`, but npm treats it as satisfied because `glob` already resolves flat to `10.5.0`. Cosmetic only, non-blocking.

### Prior non-blocking nits
All previously-flagged nits (upload progress granularity, presigned-URL localhost dev concern, category `icon_key` forwarding, polling-timer cleanup on package switch/modal reopen, overlapping parse-status polling / out-of-order responses, `SkillDetailDrawer` stale-response protection, list-load request cancellation) live in `src/**`, which this delta does **not** touch. Their status is unchanged from the prior approval; none are regressed and none are newly addressed by this commit.

### Regression check
No new blockers. A dev-tooling bump with byte-identical source and a consistent lockfile carries no runtime regression risk. Vitest 2 → 3 is a test-runner major bump; the PR body states the 50-test suite still passes post-bump, which is the appropriate gate for this change.

Carrying forward the prior APPROVE. 👍

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #119 (octo-admin)

Reviewed at head `737a51bf730a088a4d774678f8be4bf5880e9ed1`. Scope: Skill Market and System Skill admin pages — new routes, API client, upload flow, category CRUD, i18n wiring.

## Summary

This PR adds two new admin sections: a Skill Market page (listing, upload, edit, category management) and a System Skill page (internal skill CRUD with zip upload and parse polling). The API layer in `src/api/skill.ts` introduces a separate axios instance pointed at `/market/api/v1` with its own auth interceptor, presigned-URL upload helpers, and a polling loop for async parse tasks. Route gating uses the existing `CapabilityRoute` pattern with new `skill.read` / `skill.write` capabilities. Overall the code is well-structured and consistent with the project's conventions.

## Verification

Static analysis only at head `737a51bf`; build and tests not executed in this environment.

- ✅ **Auth gating** — Both new routes in `src/App.tsx` are wrapped in `CapabilityRoute` with `capability="skill.read"`. Write actions (upload, edit, delete) in `SkillTab.tsx` and `CategoryTab.tsx` check `hasCapability('skill.write')` before rendering controls. Consistent with existing `CapabilityRoute` pattern.
- ✅ **Presigned URL safety** — `assertSafeExternalUrl()` at `src/api/skill.ts:244` rejects non-HTTPS URLs (except localhost HTTP for dev). Called before every `fetch()` PUT to a presigned URL in `putPresignedFile()` and `uploadIcon()`. Test coverage in `src/api/skill.test.ts` validates the `javascript:` scheme rejection.
- ✅ **Auth interceptor parity** — The marketplace axios instance (`src/api/skill.ts:18-46`) replicates the shared instance's request/response interceptors: token header injection and 401→logout redirect. The `Accept-Language` header uses `i18n.resolvedLanguage` with `FALLBACK_LANGUAGE` fallback, matching `src/api/index.ts`.
- ✅ **i18n wiring** — `src/i18n/index.ts` imports and registers both `skillMarket` and `systemSkill` namespaces for en-US and zh-CN. The `ns` array and `resources` objects are consistent.
- ✅ **Test coverage** — `src/api/skill.test.ts` covers `commitAdminSkillReupload`, `initReupload`, and `uploadIcon` including error paths (missing fields, unsafe URL scheme). Tests mock axios and fetch correctly.

## Findings

No P0/P1 issues. Eight P2 observations below.

### P2 — Hardcoded `visibility: 'public'` on edit (`src/pages/SystemSkill/SkillFormModal.tsx:173`)

The edit path forces `visibility: 'public'` in the update payload regardless of the skill's current visibility. If the backend supports `private` or `internal` visibility levels, this silently overwrites them on every edit. The create path does not set visibility at all (presumably server-defaulted). Consider either preserving the existing value or exposing a visibility selector.

### P2 — Arbitrary `icon_url` text input bypasses presigned upload (`src/pages/SkillMarket/SkillEditModal.tsx:82`)

The edit modal exposes a raw text input for `icon_url`, letting the admin paste any URL. This bypasses the `uploadIcon()` presigned-URL flow used elsewhere and could point to external, uncontrolled, or malicious URLs. The `SkillFormModal.tsx` in SystemSkill uses the proper `uploadIcon()` path. Worth either removing the raw input or adding URL validation/restriction.

### P2 — Non-atomic reupload-then-update sequence (`src/pages/SystemSkill/SkillFormModal.tsx:163`)

When editing a system skill with a new zip, the code calls `reuploadAdminSkill()` first, then `updateAdminSkill()` in sequence. If the second call fails (network error, validation), the skill has a new package version but stale metadata. There is no rollback or transactional guarantee. Consider either making this a single backend endpoint or at minimum surfacing a clear error to the user about the partial state.

### P2 — Category reorder lacks atomicity (`src/pages/SystemSkill/CategoryTab.tsx:101`)

`handleMove` fires concurrent `Promise.all` of individual `updateSkillCategory(id, { sort_order })` PATCHes. If one fails, the list ends up in a partially-reordered state. For a small number of categories this is low-risk, but a single `PATCH /categories/reorder` endpoint (or sequential updates with error rollback) would be more robust.

### P2 — Infinite polling with no attempt cap in SystemSkill upload (`src/pages/SystemSkill/SkillFormModal.tsx:122`)

The parse-task polling uses `setInterval` with no maximum attempt count. If the backend parse task hangs or the task ID is lost, the interval runs forever, leaking memory and keeping the modal in a loading state. The marketplace upload flow in `src/api/skill.ts:310` correctly caps at 60 attempts. Apply the same cap here.

### P2 — No client-side file size validation (`src/pages/SkillMarket/SkillUploadModal.tsx:40`)

The upload modal passes the file directly to `uploadAndParseSkillZip()` without checking file size. A very large file would be uploaded entirely before the server rejects it. The presigned-URL flow in `putPresignedFile()` has no `AbortController` either, so the user cannot cancel a large upload mid-flight. Consider adding a client-side size check (e.g. 50 MB max) and an abort mechanism.

### P2 — No upload timeout in `putPresignedFile` (`src/api/skill.ts:258`)

The raw `fetch()` PUT has no `AbortController` or timeout. If the presigned-URL host is unreachable or stalls, the upload hangs indefinitely with no user feedback. The function returns a bare `Response` — callers have no way to cancel. Adding an `AbortController` with a reasonable timeout (e.g. 5 minutes) would prevent stuck UI states.

### P2 — `createCategory` silently discards `icon_key` parameter (`src/api/skill.ts:518`)

`createSkillCategory` accepts an `icon_key` argument but the POST body at `:520` only sends `{ name, description }`. If the backend supports category icons, they are never set. Either include `icon_key` in the payload or remove the parameter from the function signature to avoid confusion.

## Human-verify

1. Server-side auth enforcement for the `/market/api/v1/admin/*` endpoints — the client-side capability gating (`skill.read`, `skill.write`) is correctly wired, but the marketplace backend must independently enforce role-based access. Not verifiable from this frontend-only repo. This is **not** a merge blocker for this PR.
2. Presigned-URL origin trust — `assertSafeExternalUrl()` validates scheme (HTTPS or localhost HTTP) but does not validate the hostname against an allowlist. If the marketplace backend can be tricked into issuing presigned URLs for arbitrary hosts, the client would PUT data to them. Confirming the backend only issues presigned URLs for its own storage buckets would close this. This is **not** a merge blocker for this PR.

- Route registration in `src/App.tsx` uses `lazy()` + `Suspense` consistent with other routes.
- `src/auth/capabilities.ts` correctly adds `skill.read` and `skill.write` to `MANAGER_CAPABILITY_KEYS` and maps `skill.read` in `firstManagerPath()`.
- Nav item in `src/layouts/MainLayout.tsx` uses `ShopOutlined` icon and is gated by `skill.read`.
- `parseTaskByUploadId` Map in `src/api/skill.ts` is module-scoped and only written/read by the upload flow — no concurrent-access concern.
- The legacy compatibility exports (`listSkills`, `createSkill`, etc.) in `src/api/skill.ts` correctly delegate to the new admin endpoints.
- Category CRUD modals in both `SkillMarket/CategoryTab.tsx` and `SystemSkill/CategoryTab.tsx` handle loading and error states consistently.
- `SkillDetailDrawer.tsx` download button uses the presigned URL from the API response with `assertSafeExternalUrl()` validation.
- Vite proxy config, nginx template, and Dockerfile changes are cosmetic (comments, env-var parameterization) with no behavioral impact.
- `package.json` dependency upgrades (`antd`, `@ant-design/icons`) are minor-version bumps.

## Verdict: APPROVED

Well-structured feature addition that follows the project's existing patterns for auth gating, API layering, and i18n. All eight findings are P2 (non-blocking quality improvements). No correctness, security, or build-breaking issues identified.

[Octo-Q] verdict: APPROVE — No P0/P1. Auth gating and presigned URL safety are correctly implemented. All findings are P2 quality improvements (atomicity gaps, missing caps/timeouts, parity inconsistencies). Safe to merge.

yujiawei (approved):
# Code Review — PR #119 (octo-admin)

Reviewed at head SHA `737a51bf730a088a4d774678f8be4bf5880e9ed1`. Independently verified build, tests, and the security-relevant claims; ran a second adversarial pass and a long-context cross-file pass and folded both in below.

## Verification performed

- `npm ci` + `npm test -- --run` → **54 tests pass** (incl. `src/api/skill.test.ts`).
- `npm run build` (`tsc && vite build`) → **passes**, 3234 modules, only the pre-existing >500 kB chunk-size warning.
- `rg -a "VITE_MARKETPLACE_ADMIN_TOKEN|MARKETPLACE_ADMIN_TOKEN|X-Admin-Token" dist` → **0 matches**; no admin secret bundled. Auth uses the operator's own Octo login token in the `token` header, mirroring the existing `src/api/mcp.ts` client; marketplace is expected to enforce `superAdmin` server-side.

## 1. Spec compliance

**Spec: ✅**

The diff implements the stated scope and nothing beyond it:
- Skill Market admin (`/skill-market`): list/detail/edit/delete/upload/category management via `/market/api/v1/admin/*`.
- System Skill admin (`/system-skill`): reupload commit (`POST /admin/skills/:id/reupload`) before a separate metadata `PATCH`; the metadata PATCH intentionally omits `version` so it cannot clobber the reupload's version.
- `skill.read` / `skill.write` capabilities, `CapabilityRoute` guards, sidebar entry (skill-market only; system-skill is direct-URL, both gated on `skill.read`), en/zh i18n, and proxy/config alignment.
- No unrequested flags, fields, or endpoints. No over-build.

## 2. Code quality

**Quality: Approved** — no P0/P1 after verification. The items below are all P2 (non-blocking); the first two are worth fixing before this sees heavy operator use.

### P2 — highest priority

- **Cross-skill upload contamination via orphaned poll interval** — `src/pages/SystemSkill/SkillFormModal.tsx:127`. If the modal is closed while an upload is still in flight (before `setInterval` is created), the effect cleanup has nothing to clear; when `triggerParse` later resolves it creates a poll interval that outlives the close. Reopening for a *different* skill can then write the first upload's `parseTaskId`/parse results into the second skill's form. Guard the async callbacks with a per-run "cancelled"/request-id token (or an `AbortController`) so a stale run cannot mutate state after cleanup. The form visibly repopulates and the reupload endpoint binds the parse task to the target skill server-side, which limits blast radius — **please have a human confirm the backend rejects a parse task bound to a different skill.**

- **SSRF guard is scheme-only, not host-constrained** — `src/api/skill.ts:366` (`assertSafeExternalUrl`). It admits any `https:` host (plus `http://localhost|127.0.0.1`). Note the PUT goes through native `fetch` with only backend-supplied headers, so the operator's Octo token is **not** attached — this is not a token-leak path. It is defense-in-depth against a compromised/misconfigured marketplace presign response. The real SSRF boundary is server-side; keep it there. Optionally restrict to an allowlisted storage host if one is known at build time.

### P2 — other

- **Submit enabled during `pending`/`parsing`** — `SkillFormModal.tsx:124` sets `parseTaskId` immediately after `triggerParse`, before parse success. Submitting mid-parse posts an unfinished task. Gate submit on `parseStatus === 'success'`.
- **No poll timeout in the modal** — `SkillFormModal.tsx:127` polls every 2s with no max-attempts cap, unlike `uploadAndParseSkillZip` in `skill.ts:486` (60-attempt cap → `Parse timed out`). A stuck backend task polls until the modal unmounts. Add a matching cap.
- **Non-atomic category reorder** — `src/pages/SystemSkill/CategoryTab.tsx:97` fires two independent `updateCategory` calls in `Promise.all`; a partial failure leaves the server with half a swap (duplicate/incorrect `sort_order`). Prefer a single reorder endpoint, or reconcile from the reload.
- **Stale-response races on rapid interaction** — `src/pages/SkillMarket/SkillDetailDrawer.tsx:27` (open A then B quickly; A resolves last and overwrites B) and `src/pages/SkillMarket/SkillTab.tsx:45` (rapid filter/sort changes; older list response overwrites newer). `SystemSkill/DetailDrawer.tsx` already uses a `stale` flag — apply the same pattern here.
- **Download URL not scheme-validated** — `src/api/skill.ts:339` returns the backend `download_url` opened via `window.open(url, '_blank')` (`SkillTab.tsx:103`, `SkillDetailDrawer.tsx:128`) without the `assertSafeExternalUrl` check used on uploads. Low risk (admin-only, backend-sourced), but asymmetric with the upload path.
- **Redundant second lockfile** — the PR adds a new `pnpm-lock.yaml` (3558 lines) while `Dockerfile`/CI use `npm ci` + `package-lock.json`. Both currently agree (vitest 3.2.7, vite 6.4.3) so there is no version desync today, but two lockfiles drift over time. Drop `pnpm-lock.yaml` unless the toolchain is intentionally moving to pnpm.

## 3. Security notes for human verification

This PR is auth/marketplace-adjacent. The frontend is correct in shape (no bundled secret; token-based auth; write controls gated on `skill.write`), but capability checks here are **UI-only** by design. A human should confirm on the marketplace backend:
- `/market/api/v1/admin/*` verifies the Octo token and admits only `superAdmin` (list, create, edit, delete, upload, reupload, category mutations).
- Presigned upload/parse/reupload validates schemes, private/link-local/metadata ranges, and redirect hops server-side; a parse task cannot be committed to a skill it wasn't bound to.
- `MARKETPLACE_BACKEND` in the deployed nginx template points at the real marketplace host.

## 4. Coverage / blind spots

Not verifiable from this frontend diff (all backend- or deploy-owned): marketplace `superAdmin` enforcement, parse-task lifecycle & binding validation, presigned host/header constraints, server-side SSRF validation, and runtime proxy config. Advisory legs: this review absorbed a line-level adversarial pass (confirmed the two race/SSRF items above, downgraded from P1 after verifying the token is not attached to the PUT) and a long-context cross-file pass (its lockfile/registry P1 was refuted — the `npmmirror` URLs are pre-existing baseline that this PR net-reduces, and `package.json` *is* bumped in the diff, contrary to that leg's claim). Both legs ran successfully.

## Verdict

No P0/P1 blockers. The P2 items — chiefly the upload-race guard and the submit-during-parse gate — are worth a fast follow-up but do not block merge, given admin-only access and server-side binding. Approving; please action the security-verification checklist in §3 before/at merge since this is flagged for human review.

PR #135, opened 2026-08-09
refactor(system-setting): make 60+ settings navigable and scannable

an9xyz:
## Why

`/system-setting` has grown to 60+ settings across ~10 categories, all rendered at once into a two-column CSS masonry. That layout stopped scaling:

- **No way to find a setting** — no search, and the category list itself was invisible.
- **Reading order fought the eye** — `column-count` fills the left column top-to-bottom before jumping right, and mixing a 1-item category with an 11-item one left ragged gaps.
- **Long descriptions blew up row heights** — the raw `description` was used verbatim as the row title, so the 40–80 character `botcard` / `app_bot` copy wrapped to 2–3 lines and pushed the control far from its label.
- **Badge noise** — nearly every one of the 60 rows carried a "default" badge, which buried the handful of rows that are actually overridden in the DB.
- **The save button scrolled away**, so editing something near the bottom meant scrolling back up.

## What changed

- **Category rail + single column.** Only the selected category renders, so the page stays about one screen tall. Two virtual groups sit above the real categories: **unsaved** (review cross-category edits before committing) and **overridden** (what this environment actually changes — more useful than "show everything").
- **Toolbar search** across all categories, matching `category.key` and the description, with results grouped by category and a match count.
- **First-sentence row titles.** Titles are cut at the first `；` / `;` / `。` outside brackets and kept to one line; the full text lives in a tooltip that opens on hover **and** keyboard focus. Every row is inline now, so heights are uniform and controls share a single right edge.
- **Quieter, more informative markers.** The "default" badge is gone (following the default is the norm). Overridden rows get a brand left rule + badge; rows with unsaved edits get a warning left rule + badge.
- **Floating save bar** with an unsaved count and a discard action, so edits can be committed from anywhere in the list.
- Numeric placeholders shortened to `default 60` — the long "follow default (current: …)" copy was being clipped by the narrow number box.

## Correctness fix required by the new layout

The payload now reads `form.getFieldsValue(true)` instead of the return value of `validateFields()`. antd only returns **mounted** field values otherwise; with a single category rendered, the previous call would have submitted every other category as an empty string and wiped it. Same reasoning for the unsaved-count computation, which no longer uses `onValuesChange`'s `allValues` argument.

Save semantics are otherwise unchanged — full payload, bool tri-state, encrypted mask — and the existing `valuesFromSettings` / `valuesToPayload` tests are untouched.

## Note for future work

The save bar is deliberately `position: fixed`, not `sticky`. `MainLayout`'s `<Content>` sets `overflow: auto`, which makes it the nearest scrollport for sticky children while never scrolling itself (the document scrolls), so `position: sticky` is a no-op inside any page. `.dashboard-anchor-nav` is in the same situation; not touched here.

## Test plan

- [x] `pnpm test` — 81 tests pass, including 21 new helper tests (28 in the file) covering bracket-aware summary extraction, query matching, and change detection
- [x] `pnpm build` (`tsc && vite build`) passes
- [x] End-to-end save against a stub backend with 60 settings: edited two different categories, saved, and asserted the POST payload carries **all 60 items** with both edits, untouched values preserved, follow-default rows still empty, and a blank encrypted field still sending the `****` mask
- [x] Cross-category edits both appear in the "unsaved" group (verifies the `getFieldsValue(true)` path)
- [x] Light / dark mode, and narrow viewport (<992px, rail degrades to a horizontal pill row)
- [x] Accessibility: the hidden save bar is `visibility: hidden` so its buttons stay out of the tab order; the truncated-description tooltip is reachable by keyboard focus
- [x] `zh-CN` and `en-US` verified in the running app and key-for-key in sync (52 keys each)


Jerry-Xin (dismissed):
This PR is in scope for `octo-admin` and the refactor is sound overall; I found no blocking issues.

💬 Non-blocking
- 🟡 Warning: The toolbar save button remains enabled even when the initial settings load has not completed or has failed, so `handleSave` can run with `settings === []` and submit an empty payload. If the backend treats the endpoint as replacement-style, this is risky. Consider disabling save while `loading` or until `settings.length > 0`. `src/pages/SystemSetting/index.tsx:213`, `src/pages/SystemSetting/index.tsx:135`
- 🔵 Suggestion: The new keyboard-focus tooltip covers sentence-truncated descriptions, but visually clipped one-sentence descriptions without separators only get the hover tooltip on the label. If keyboard visual access to all truncated text is a hard accessibility goal, consider making the info button appear whenever the rendered label can be clipped. `src/pages/SystemSetting/SettingRow.tsx:136`, `src/pages/SystemSetting/SettingRow.tsx:148`

✅ Highlights
- The `getFieldsValue(true)` path correctly addresses the hidden-category full-payload issue. `src/pages/SystemSetting/index.tsx:126`, `src/pages/SystemSetting/index.tsx:139`
- Helper coverage is strong for summaries, query matching, grouping, encrypted fields, and change detection. `src/pages/SystemSetting/helpers.test.ts:122`, `src/pages/SystemSetting/helpers.test.ts:216`
- The virtual changed/overridden groups and category rail are cleanly separated from row rendering. `src/pages/SystemSetting/CategoryNav.tsx:54`, `src/pages/SystemSetting/index.tsx:72`

Verification: focused SystemSetting helper tests passed, TypeScript passed, and Vite build passed. Full Vitest had unrelated local environment failures in `src/store/feature.test.ts:17` because `localStorage` was unavailable under this Node run.

yujiawei (dismissed):
# Code Review — PR #135 (octo-admin)

Reviewed at head SHA `498f58f79b9b94f18d878023493c5756d6e915d4`, diffed against merge-base `f3fd16f`. 8 files, +854/−108.

I read the full post-PR files (not just the diff), verified the antd/`rc-field-form` behaviour the correctness fix depends on against the installed `rc-field-form@2.7.1` source, and ran the build and test suite locally.

---

## 1. Scope / spec compliance

**Spec: ✅**

- **Nothing missing.** Every item in the PR description is present in the diff: category rail with two virtual groups (`CategoryNav.tsx:66-85`), toolbar search across all categories (`index.tsx:194-202`, `helpers.ts:114-121`), first-sentence row titles with tooltip (`helpers.ts:94-111`, `SettingRow.tsx:146-164`), badge cleanup (`badge.default` removed from both locale files), floating save bar with discard (`index.tsx:290-300`), shortened numeric placeholder (`SettingRow.tsx:84-88`).
- **Nothing extra.** All 8 touched files are `SystemSetting/*`, its two locale files, and the `System Setting` section of `admin.css`. No unrelated pages, no API/contract changes, no new dependencies. `.dashboard-anchor-nav` is indeed left alone as stated.
- **No divergence.** Dead CSS from the old layout (`.setting-masonry`, `.setting-row--inline`, `.setting-row--block`, `.setting-badge--default`) is fully removed, and `grep` finds no remaining references anywhere in `src/`. Both removed i18n keys (`badge.db`, `badge.default`) are likewise unreferenced.

### Test-plan claims, checked

| Claim | Result |
|---|---|
| `pnpm test` — 81 tests, 28 in `helpers.test.ts` | ✅ verified: 12 files, 81 passed, `helpers.test.ts` 28 |
| `pnpm build` (`tsc && vite build`) | ✅ verified, clean (only the pre-existing >500 kB chunk warning) |
| `zh-CN` / `en-US` key-for-key in sync, 52 keys each | ✅ verified: 52/52, zero one-sided keys, zero orphaned keys |
| Hidden save bar uses `visibility: hidden` so buttons leave the tab order | ✅ verified at `admin.css:2260` |
| Cross-category edits both land in the "unsaved" group | ✅ verified by construction (see below) |
| Truncated-description tooltip reachable by keyboard focus | ⚠️ partially — holds only when the description contains a separator (finding P2-1) |
| E2E save against a stub backend / light-dark / narrow viewport | ⛔ not reproducible here (no backend, no browser) — see coverage note |

---

## 2. Code quality

**Quality: Approved** — no P0 or P1. Six P2s below; none of them blocks merge.

### The correctness fix is right, and it was genuinely required

This is the load-bearing part of the PR, so I verified it at the library level rather than taking the description's word for it. All three links hold in `rc-field-form@2.7.1`:

1. `useForm.js:648` — `onValuesChange(changedValues, this.getFieldsValue())`. The second argument is the **mounted-only** projection, so the old `onValuesChange(_, allValues)` pattern would have reported every unrendered category as absent. Dropping it (`index.tsx:128`) is correct.
2. `useForm.js:226` — `if (mergedNameList === true && !mergedFilterFunc) return this.store`. `getFieldsValue(true)` returns the whole store, so `index.tsx:126` sees all 60 keys.
3. `useForm.js:527-530` — `isMergedPreserve(undefined)` resolves to `this.preserve ?? true`. The `<Form>` at `index.tsx:218` sets no `preserve`, so unmounting a category's fields does **not** delete their store values. This is the assumption `getFieldsValue(true)` rests on, and it holds.

Also confirmed: `setFieldsValue` (`useForm.js:651-658`) fires `notifyObservers`/`notifyWatch` but **not** `onValuesChange`, which is why the explicit `setChangedFields([])` in `fetchSettings` (`index.tsx:110`) and `handleDiscard` (`index.tsx:132`) is necessary rather than redundant. And in `updateValue` (`useForm.js:633`) `updateStore` runs *before* the `onValuesChange` callback, so the zero-argument `handleValuesChange` reading `getFieldsValue(true)` synchronously sees post-update state — no staleness.

Net: had the single-category render landed without this change, saving would have blanked ~59 of 60 settings. Good catch by the author, and the fix is complete rather than partial.

Diff-based `changedFieldNames` (`helpers.ts:130-146`) is also a real improvement over the old sticky `setDirty(true)`: edit-and-revert now correctly reports zero unsaved changes, and the `is empty for a freshly loaded form` test at `helpers.test.ts:252-262` pins the `valuesFromSettings` → `changedFieldNames` round-trip, which is exactly the invariant that would silently rot.

### P2 findings

**P2-1 — a visually clipped label with no separator has no keyboard-reachable tooltip.** `SettingRow.tsx:134-136`:

```tsx
const description = (item.description || '').trim()
const hasMoreDetail = description !== '' && description !== summary
```

`hasMoreDetail` detects *sentence extraction*, not *visual truncation*. The two are independent: `.setting-row-label` is `white-space: nowrap; overflow: hidden; text-overflow: ellipsis` (`admin.css:2164-2166`), so a long description with **no** `；`/`;`/`。` yields `summary === description`, renders no info button, and is still clipped on screen. Its only tooltip is the one on the bare `<span>` at line 148, which uses antd's default hover-only trigger.

Failure scenario: a setting described as `侧边栏最近会话列表按最后活跃时间倒序排列并按天分组显示的最大天数上限` (no separator) in a viewport where the text column is ~300px → row shows `…按天分组显示…` with an ellipsis, and a sighted keyboard-only user has no way to reveal the rest. Screen-reader users are fine (the full text is in the DOM, and the control carries `aria-label={settingLabel(item)}` = the full description), so this is narrower than it first looks — but it does mean the test-plan line "the truncated-description tooltip is reachable by keyboard focus" is true only for the separator case.

Suggested fix: base the info button on "is there text the user cannot see" rather than "did we cut a sentence" — either always render it when `description` is non-empty, or add `trigger={['hover', 'focus']}` plus `tabIndex={0}` to the label span.

**P2-2 — reverting an edit inside the "Unsaved changes" group yanks the row out from under the cursor.** `index.tsx:76` + `index.tsx:128`:

```tsx
if (activeKey === CHANGED_KEY) return [[CHANGED_KEY, changedItems]]
...
const handleValuesChange = () => setChangedFields(changedFieldNames(allValues(), settings))
```

The group's membership is recomputed on every keystroke. While viewing it, typing the value back to its saved state removes that row mid-edit — the input unmounts and focus is lost. If it was the last unsaved row, `changedItems` empties, the `stillValid` effect (`index.tsx:88-99`) fires, and `activeKey` jumps to `__overridden__`, so the user is navigated to a different list while typing.

Repro: open **Unsaved changes** with exactly one edit where `support.email` was saved as `a@b.com` and is now `a@b.comX`; press Backspace → the row and then the whole group vanish.

Suggested fix: freeze the group's membership on entry (snapshot the name list when `CHANGED_KEY` is selected and keep rendering those rows, marking reverted ones as no-longer-changed instead of removing them). At minimum, don't auto-navigate away while an input inside the group has focus.

**P2-3 — `validateFields()` and the submitted payload cover different field sets.** `index.tsx:136-139`:

```tsx
await form.validateFields()
...
await updateSystemSettings(valuesToPayload(allValues(), settings))
```

`validateFields()` walks mounted field entities only; `allValues()` is the whole store. Today this is latent — no setting `Form.Item` in `SettingRow.tsx` declares `rules`, so validation cannot reject anything — but it is a trap primed to fire: the first time someone adds a `rules` constraint (a port range, a URL format), a value edited in category A and submitted from category B will bypass it silently. Worth either a comment marking the hazard, or switching to `validateFields(Object.keys(allValues()))` so the two sets are the same by construction.

Separately, `await form.validateFields()` sits outside the `try`, so a rejection would surface as an unhandled promise rejection with no user-facing message rather than a toast. Pre-existing (same shape before this PR) and unreachable while there are no rules, but it compounds the above.

**P2-4 — `settingSummary` bracket tracking counts depth without matching types.** `helpers.ts:100-109`:

```ts
if (BRACKETS_OPEN.includes(char)) depth += 1
else if (BRACKETS_CLOSE.includes(char)) depth = Math.max(0, depth - 1)
```

`）`, `)`, `】`, `]` are interchangeable, so `标题（说明]；后续` has `]` close `（` and the summary is cut at `；`; conversely an unmatched `（` suppresses every later separator and returns the whole string. Both degrade gracefully (ellipsis + tooltip), and `Math.max(0, …)` correctly stops depth going negative on a stray closer, so this is a robustness nit rather than a bug — but `helpers.test.ts:136-144` only covers the well-formed case, and a malformed-bracket test would pin the intended behaviour.

**P2-5 — `SENTENCE_ENDINGS` omits `.`, `?`, `!` and their full-width forms.** `helpers.ts:84`:

```ts
const SENTENCE_ENDINGS = '；;。'
```

An English or question-form description (`Whether to enable QR login. Disabled by default`, `是否开启？关闭后…`) is returned whole and falls back to CSS ellipsis. Note this is a genuine trade-off, not a clear miss: adding `.` would cut `octo/v1`, `admin@example.com`, `1.2.3`, and `OCTO_CARD_MESSAGE_ENABLED`-style copy at the wrong place, so the conservative set may well be the right call. If you keep it, the doc comment at `helpers.ts:88-92` already states the exact separator set, which is honest — consider adding `？`/`！`/`?`/`!` (safe, unlike `.`) and renaming the `handles a full stop and an ASCII semicolon` test (`helpers.test.ts:146`), which currently exercises `。` and `;` but not an ASCII full stop.

**P2-6 — the fixed save bar has no narrow-viewport guard.** `admin.css:2243-2262` sets `position: fixed; right: 28px; bottom: 24px` with no `left`, `max-width`, or wrapping, and the `max-width: 991.98px` block (`admin.css:2279-2320`) doesn't touch it. At a 320px viewport the available width before `right: 28px` is 292px; the `en-US` content ("`3 unsaved`" + "`Discard changes`" + "`Save Settings`" + 26px padding + two 10px gaps) exceeds that, so the pill grows off the left edge of the viewport and clips. `zh-CN` is shorter and fits. Low priority for an admin console, but a `left: 16px` (or `max-width: calc(100vw - 32px)` + `flex-wrap: wrap`) in the existing media query closes it cheaply.

The `position: fixed`-over-`sticky` rationale in the PR description checks out, incidentally: `MainLayout.tsx:346` does set `overflow: 'auto'` on `<Content>`. And `fixed` is the safe choice regardless of which element ends up being the scrollport, so the note is right for a slightly more general reason than stated.

### Things I checked that are clean

- **CSS tokens** — all 15 custom properties used by the new rules (`--a-warning`, `--a-radius-pill`, `--a-bg-elevated`, `--a-shadow-lg`, `--a-bg-subtle`, `--a-brand-subtle-bg/-fg`, `--a-text-quaternary`, …) are defined in all three theme blocks, so light/dark parity holds without hardcoded colours. `color-mix(in srgb, …)` at `admin.css:2215` matches 16 existing uses in the same file, so it introduces no new browser-support floor.
- **Control widths** — `.setting-row-control` targets `.ant-input-affix-wrapper` as well as `> .ant-input` (`admin.css:2146-2150`), which is what makes `allowClear`/`Input.Password` (both of which wrap the input in an affix wrapper) size correctly alongside a bare `Input`. Easy thing to get wrong; it's right here.
- **Flex shrink** — `.setting-row-label` has no explicit `min-width: 0`, but its `overflow: hidden` sets the automatic minimum size to 0, so ellipsis works inside the flex row. Fine as written.
- **`onValuesChange` closure freshness** — antd re-registers callbacks each render, so the `settings` captured by `handleValuesChange` is never stale.
- **Tri-state / encrypted semantics** — `valuesToPayload` and `valuesFromSettings` are untouched, and `changedFieldNames` handles encrypted (always-blank load ⇒ any input is a change), legacy bool (`'true'` vs `'1'` normalised, `helpers.test.ts:229`), and `int` (`5` vs `'5'` compared as strings, `helpers.test.ts:241`) correctly.
- **List keys** — rows keyed on `settingFormName(category, key)` and cards on the group key; no collision path, since the virtual keys are `__changed__`/`__overridden__`.

---

## 3. Verdict

**APPROVE.** Scope is clean, the one behavioural change that could have destroyed data is both necessary and correctly implemented (verified against the form library, not just asserted), build and tests are green, and i18n is at parity. The six P2s are follow-up material — P2-1 and P2-2 are the two I'd actually spend time on, since both sit inside the features this PR is adding.

## 4. Suggested follow-ups

1. Make the info button track *visual* truncation rather than sentence extraction (P2-1) — this is the one that makes the accessibility claim fully true.
2. Snapshot the "Unsaved changes" membership on entry so rows don't disappear mid-keystroke (P2-2).
3. Align `validateFields()` with the submitted key set, or leave a comment at `index.tsx:136` recording that rules on setting fields are currently unenforceable (P2-3).
4. `left: 16px` on `.setting-save-bar` inside the existing narrow-viewport media query (P2-6).

## 5. Additional observations (outside the diff's remit)

- **Refresh silently discards unsaved edits.** `index.tsx:210` calls `fetchSettings()`, which overwrites the form and clears `changedFields` with no confirmation. Behaviourally identical to before this PR, so not a regression — but this PR promotes "unsaved changes" to a first-class concept (count in the button, dedicated nav group, explicit Discard action), which makes a Refresh button that quietly throws the same edits away feel inconsistent. A confirm when `dirty` would be cheap.
- **One-frame Empty flash on a pristine environment.** `activeKey` initialises to `OVERRIDDEN_KEY` (`index.tsx:56`); if nothing is overridden, the first post-fetch render shows the "No settings in this category" `Empty` before the `stillValid` effect switches to the first real category. Cosmetic, and only on installs with zero overrides.
- **Search doesn't match localised category titles.** `matchesSettingQuery` (`helpers.ts:114`) matches `category.key` and `description`, so `support` finds the email settings but `Email Service` / `邮件服务` does not. The placeholder says "name or key", which a user may reasonably read as the displayed category name.

## 6. Coverage note

What I could **not** verify, in case another reviewer wants to close the gap:

- The end-to-end save against a real or stub backend (no backend available here). I verified the payload mechanism at the `rc-field-form` level and via `valuesToPayload` unit tests, but not an actual 60-item POST.
- Rendered appearance: light/dark, the <992px pill-row rail degradation, and actual label truncation points. P2-1 and P2-6 are reasoned from the CSS and would benefit from a screenshot check.
- Real backend `description` strings. Whether P2-1 and P2-5 bite in practice depends entirely on how long and how punctuated the actual 60 descriptions are — worth eyeballing the live page once.
- The accessibility tree as exposed by an actual screen reader (antd `Tooltip`'s `aria-describedby` wiring on the info button in particular).
- Merge-order interaction with any sibling PR touching `admin.css` — `MERGEABLE` on this PR alone does not guarantee a green `main` after both land.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #135 (octo-admin)

## Summary

This PR refactors the System Settings page from a two-column masonry of all categories into a category rail plus single-category panel, adding cross-category search, two virtual groups ("Unsaved", "Overridden"), per-row unsaved/overridden markers, first-sentence row titles with full-description tooltips, and a floating save bar with an accurate unsaved count. Change detection moves from a sticky `dirty` boolean to `changedFieldNames()`, which diffs the whole form store against the last fetched server values. The refactor is careful: the new helpers are well tested, i18n key parity holds across both locales, and the CSS migration leaves no orphaned selectors. No P0/P1/P2 issues found; one cosmetic nit.

## Verification

Static analysis only at head `498f58f7`; build and tests not executed in this environment (dependencies not installed). Verified by tracing changed code:

- ✅ **Save payload stays complete under single-category rendering** — only the active category mounts `Form.Item`s, but `fetchSettings` seeds the entire store via `form.setFieldsValue(valuesFromSettings(items))` (`src/pages/SystemSetting/index.tsx:111`) and save reads `form.getFieldsValue(true)` (`:126`), which in antd v5 returns the whole store including unmounted fields — the documented pattern for partially-rendered forms. `valuesToPayload` then maps over all `settings`, so the POST body covers every item exactly as before the refactor.
- ✅ **Change detection round-trips** — `changedFieldNames(valuesFromSettings(items), items) === []` is locked by a test (`src/pages/SystemSetting/helpers.test.ts:256`); legacy bool normalisation (`'true'` vs `'1'`), encrypted-blank semantics and int string-coercion are each covered.
- ✅ **i18n contract intact** — en-US/zh-CN `systemSetting.json` key sets are identical (checked programmatically, zero diff both directions); every new key (`nav.*`, `group.*`, `search.*`, `stat.summary`, `action.saveWithCount`, `action.discard`, `unsavedCount`) exists in both locales, and `common:action.refresh` already exists in `common.json` for both languages (same pattern as Dashboard/Users pages).
- ✅ **Selection validity** — the effect at `src/pages/SystemSetting/index.tsx:87` re-keys when a virtual group empties or a refresh drops the active category; nav counts and panel contents derive from the same `settings` array, so they cannot drift.

## Findings

No P0/P1/P2 issues. One nit.

### Nit — Empty-state flash on first load and after category-invalidating refreshes (`src/pages/SystemSetting/index.tsx:248`)

`loading` is initialised to `false` (`:49`), so on the very first render — before the mount-time `fetchSettings()` effect flips it — `matchCount === 0 && !loading` is true and the `group.empty` Empty card paints for one frame. The same one-frame flash occurs when a refresh drops the active category: `visibleGroups` renders `[[activeKey, []]]` before the selection-validity effect (`:87`) re-keys. Initialise `loading` to `true` (or suppress the Empty state until the first fetch settles) to remove the flash.

- **Form-store lifecycle pairs**: every state setter has a clearer — fetch resets `changedFields` and re-seeds the store, discard re-seeds from server values, save refetches with `keepSavedHint`; `savedAt` is cleared on manual refresh. Unmounted fields keep their values (`preserve` default), so edits survive category switches.
- **No validation regression**: `await form.validateFields()` now only sees mounted fields, but no field declares `rules`, so nothing that was previously enforced is skipped.
- **Encrypted fields**: blank input still maps to the `****` keep-mask in `valuesToPayload` (`src/pages/SystemSetting/helpers.ts:44`); a cleared-then-retyped encrypted field behaves exactly as pre-PR.
- **Test-email action**: still reachable via the support category card and via search results (search groups by real category, so `categoryActions['support']` is preserved); virtual groups intentionally carry no header action.
- **`settingSummary` parser**: bracket-depth tracking ignores separators inside `（）()【】`; leading separators are stripped; empty/blank descriptions fall back to the key; a truncated title always exposes the full text via a keyboard-focusable tooltip button.
- **CSS migration**: no remaining references to the removed `.setting-masonry` / `.setting-row--inline` / `.setting-row--block` / `.setting-badge--default` selectors; the responsive breakpoint is preserved (rail becomes a scrollable pill row under 992px); the fixed save bar uses `visibility` (not just opacity) so hidden buttons leave the tab order, and the layout reserves 76px bottom padding while it shows.
- **No auth, endpoint, notification, or persistence-backend changes** in this diff — it is a pure UI reorganisation over the existing `/v1/manager/common/system_setting` endpoints.

## Verdict: APPROVED

The refactor preserves every runtime contract the old page had — most critically the completeness of the save payload under single-category rendering — and adds solid helper-level test coverage. The single nit is a one-frame cosmetic flash, non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1/P2; only one cosmetic nit (empty-state flash on first paint). Save-path data flow verified end to end.

yujiawei:
Follow-up correction to my review — verdict unchanged (still approved), but one observation I made was narrower than the actual behaviour, and the corrected version is worth having on record.

In §5 I described the empty-state flash as cosmetic and **"only on installs with zero overrides"**. That understated it: the flash is **unconditional and happens on every visit** to the page.

`loading` is initialised to `false` (`index.tsx:49`), and the mount-time `fetchSettings()` lives in a `useEffect`, which runs *after* the first paint. So on the very first render `settings` is `[]`, `matchCount` is `0`, and `loading` is still `false` — which means the `matchCount === 0 && !loading` branch at `index.tsx:248` wins and paints the "No settings in this category" `Empty` card, not the `Card loading` skeleton that the rest of the page is built around.

I checked this rather than reasoning about it, because I had it wrong the first time. Rendering the page with `renderToString` models the pre-effect paint exactly (`useEffect` does not run on the server):

```
EMPTY_ON_FIRST_PAINT: true
```

So the user sees an explicit "there is nothing here" message on every page load before the data arrives — which reads as a real empty result rather than a loading state. The zero-overrides case I originally described is a *second*, distinct flash that happens later, after the fetch settles and before the selection-validity effect re-keys `activeKey`.

One-line fix that closes both: initialise `loading` to `true` (`useState(true)`), so the skeleton path owns the pre-fetch render and the `Empty` card only ever means "genuinely nothing to show".

Still non-blocking, and it doesn't change the approval — the page self-corrects within a frame. Flagging it because "cosmetic, only in a rare configuration" and "an empty-state message on every page load" are different bug reports, and I gave you the wrong one.

an9xyz:
Thanks both — all six findings addressed in 2eab840, with a note on the two I left as-is.

**P2-2 / unsaved group yanking rows (the one that actually bit users).** Confirmed exactly as described: membership was derived live and recomputed per keystroke, so typing a value back to its saved state unmounted the input mid-edit, and reverting the last one navigated away via the `stillValid` effect. The group's membership is now frozen on entry and grow-only while open — reverted rows stay in place and just lose their marker, and an edit made in another category still appears. Verified in a browser: row stays rendered, input keeps focus, panel does not navigate, badge clears, and the rail drops the group once nothing is unsaved.

**Save enabled during initial load.** Fixed — `disabled={loading || settings.length === 0}`. Worth noting the payload was `{items: []}` rather than a destructive one, so against a per-item upsert the only damage was a misleading success toast; disabled either way.

**P2-6 / narrow save bar.** Pinned on both edges inside the existing `max-width: 991.98px` block, plus `flex-wrap`. Measured at 320px en-US: `x=16`, right edge `304`. Sidenote from testing it — at 320px the shell as a whole is unusable regardless of this page, since `MainLayout`'s `<Sider>` has no `breakpoint` and keeps its 220px. Out of scope here, but that's the real blocker at that width.

**P2-1 / keyboard access to clipped titles.** You're right that my test-plan line was only true for the separator case, and that `hasMoreDetail` detects sentence extraction rather than visual truncation. Rather than put an ⓘ on every row — which would restore exactly the badge noise this PR removed — the label and its ⓘ are now a single tooltip trigger and a single tab stop, with the ⓘ decorative (`aria-hidden`). That covers both truncation causes without doubling the tab order. Verified: a genuinely clipped no-separator label (`sticker.allowed_mime` at 1100px) opens its tooltip on keyboard focus, and the ⓘ contributes zero tab stops.

**P2-5 / sentence endings.** Added `？！?!`. Agreed on keeping `.` out, and there's now a test pinning that `octo/v1.2` and `admin@example.com` survive intact, so the trade-off is documented in code rather than just in a comment.

**P2-4 / bracket depth.** Left counted rather than type-matched, since every malformed case degrades to a longer title (ellipsis + tooltip) and never a wrong one — but added the tests you noted were missing: mismatched closer, unclosed opener, and a stray closer that must not push depth negative.

**P2-3 / `validateFields()` vs payload key sets.** Left as-is with a comment at the call site recording the hazard and the fix (`validateFields(Object.keys(allValues()))`). Unreachable today since no setting declares `rules`; I'd rather not add the coupling before there's a rule to enforce.

**On the observations outside the diff:** the Refresh-discards-edits inconsistency is a fair point now that unsaved state is first-class — I'd rather do that as its own change than widen this one. The one-frame `Empty` flash on a zero-override install is real and cosmetic. On search not matching localised category names, I'd push back mildly: the placeholder's "名称" refers to the setting's own name (its description), which *is* matched — the category title isn't a setting name.

Tests are 83 now (30 in `helpers.test.ts`), build is clean, and the end-to-end save check still passes with all 60 items and both cross-category edits intact.

an9xyz:
@mochashanyao — your nit and @yujiawei's "one-frame Empty flash" observation are the same thing, and it fired on *every* page load, not just zero-override installs: `loading` started `false`, so the first render matched `matchCount === 0 && !loading` before the mount effect could flip it. Fixed in 884895d by starting in the loading state, so the card skeleton paints instead.

Verified rather than assumed — polled the DOM from first paint via `requestAnimationFrame` with the settings request held open, and `.ant-empty` never appears at any point during load while `.ant-skeleton` does.

The narrower half of that finding is still there and I've left it: if a refresh drops the active category, `visibleGroups` renders it empty for one frame before the selection-validity effect re-keys. Removing that one needs the fallback to move out of the effect and into the derivation itself, which is more restructuring than a one-frame flash on a backend-removes-a-category path justifies. Noted in the commit message so it isn't lost.

Also confirming your save-path trace matches what I measured against a stub backend: the POST carries all 60 items with cross-category edits intact, untouched values preserved, and a blank encrypted field still sending the `****` mask.

Jerry-Xin (dismissed):
The PR is in scope for `octo-admin`: it refactors the existing `/system-setting` page and keeps the system-setting API/form semantics intact. I found no blocking bugs.

💬 Non-blocking

🟡 Warning: I could not complete local verification because `pnpm test` and `pnpm build` were blocked before scripts ran by the environment’s pnpm ignored-builds policy for `esbuild@0.25.0`. This looks environmental, not caused by the PR, but CI should be trusted for final confirmation.

🔵 Suggestion: [src/pages/SystemSetting/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-623j7bfl/repo/src/pages/SystemSetting/index.tsx:154) notes that `validateFields()` currently validates only mounted fields. That is fine while rows have no validation rules, but if rules are added later, this should change to validate all setting field names so hidden categories cannot bypass validation.

✅ Highlights

- [src/pages/SystemSetting/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-623j7bfl/repo/src/pages/SystemSetting/index.tsx:130) correctly uses `getFieldsValue(true)` for full payload generation, avoiding data loss from unmounted category fields.
- [src/pages/SystemSetting/helpers.ts](/Users/mlamp/.openclaw/tmp/pr-review-623j7bfl/repo/src/pages/SystemSetting/helpers.ts:34) preserves existing bool tri-state and encrypted-mask behavior.
- [src/pages/SystemSetting/helpers.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-623j7bfl/repo/src/pages/SystemSetting/helpers.test.ts:118) adds focused coverage for summary extraction, search matching, grouping, and changed-field detection.
- [src/pages/SystemSetting/index.tsx](/Users/mlamp/.openclaw/tmp/pr-review-623j7bfl/repo/src/pages/SystemSetting/index.tsx:135) recomputes changed state from the full form store, which matches the new single-category rendering model.

Jerry-Xin (approved):
The PR is in scope for `octo-admin`: it refactors the existing `/system-setting` page and keeps the system-setting API/form semantics intact. I found no blocking bugs.

💬 Non-blocking

🟡 Warning: I could not complete local verification because `pnpm test` and `pnpm build` were blocked before scripts ran by the environment’s pnpm ignored-builds policy for `esbuild@0.25.0`. This looks environmental, not caused by the PR, but CI should be trusted for final confirmation.

🔵 Suggestion: [src/pages/SystemSetting/index.tsx](src/pages/SystemSetting/index.tsx:154) notes that `validateFields()` currently validates only mounted fields. That is fine while rows have no validation rules, but if rules are added later, this should change to validate all setting field names so hidden categories cannot bypass validation.

✅ Highlights

- [src/pages/SystemSetting/index.tsx](src/pages/SystemSetting/index.tsx:130) correctly uses `getFieldsValue(true)` for full payload generation, avoiding data loss from unmounted category fields.
- [src/pages/SystemSetting/helpers.ts](src/pages/SystemSetting/helpers.ts:34) preserves existing bool tri-state and encrypted-mask behavior.
- [src/pages/SystemSetting/helpers.test.ts](src/pages/SystemSetting/helpers.test.ts:118) adds focused coverage for summary extraction, search matching, grouping, and changed-field detection.
- [src/pages/SystemSetting/index.tsx](src/pages/SystemSetting/index.tsx:135) recomputes changed state from the full form store, which matches the new single-category rendering model.

yujiawei (approved):
# Code Review — PR #135 (octo-admin)

Reviewed at head SHA `884895d3ed7d59ee0639ceb1c8699b61e9004285`, diffed against merge-base `f3fd16f`. 8 files, +934/−110. This is a re-review after the two follow-up commits (`2eab840`, `884895d`); the earlier approvals were auto-dismissed by those pushes, so this verdict is issued fresh against the current head rather than inherited.

I read the full post-change files (not only the diff), re-verified the form-library assumptions the correctness fix rests on against the installed `rc-field-form@2.7.1`, ran the test suite and the production build locally, and did the layout arithmetic against `MainLayout`'s actual sider/content geometry.

---

## 1. Spec compliance

**Spec: ✅**

- **Nothing missing.** Every item claimed in the PR description is present: category rail with the two virtual groups (`src/pages/SystemSetting/CategoryNav.tsx:66-88`), cross-category toolbar search (`src/pages/SystemSetting/index.tsx:217-224`, `src/pages/SystemSetting/helpers.ts:119-126`), first-sentence titles with a full-text tooltip (`helpers.ts:99-116`, `SettingRow.tsx:152-159`), badge cleanup (`badge.default` gone from both locales), floating save bar with discard (`index.tsx:323-333`), shortened numeric placeholder (`SettingRow.tsx:84-88`).
- **Nothing extra.** All 8 touched paths are `SystemSetting/*`, its two locale files, and the System Setting section of `admin.css`. No API/contract change, no new dependency, no unrelated page. `.dashboard-anchor-nav` is left alone exactly as the description says.
- **No divergence.** Dead selectors from the old layout (`.setting-masonry`, `.setting-row--inline`, `.setting-row--block`, `.setting-badge--default`) are fully removed and `grep` finds no residual reference anywhere under `src/`. Both removed i18n keys (`badge.db`, `badge.default`) are likewise unreferenced.

### The six previously-raised findings, re-checked at this head

| Finding | Status at `884895d` |
|---|---|
| Save enabled before first load lands | ✅ fixed — `disabled={loading \|\| settings.length === 0}` (`index.tsx:242`) |
| Unsaved group yanked rows out mid-keystroke | ✅ fixed — membership frozen on entry, grow-only (`index.tsx:60`, `:141-145`, `:261-267`) |
| Clipped title not keyboard-reachable without a separator | ✅ fixed — the whole title is one tooltip trigger and one tab stop, ⓘ made decorative (`SettingRow.tsx:152-158`) |
| `SENTENCE_ENDINGS` missing `？！?!` | ✅ fixed, with a test pinning that `.` stays out (`helpers.ts:86`, `helpers.test.ts:157-165`) |
| Save bar off-screen below 992px | ✅ fixed — `left: 16px; right: 16px; flex-wrap: wrap` (`admin.css:2336-2342`) |
| Malformed-bracket degradation untested | ✅ fixed — `helpers.test.ts:167-177`, including the stray-closer case |
| `validateFields()` vs. whole-store payload | ⚠️ deliberately left as-is with a hazard comment (`index.tsx:155-157`) — acceptable, see P2-3 |

### Test-plan claims, checked

| Claim | Result |
|---|---|
| `pnpm test` green | ✅ 12 files, **83** passed (30 in `helpers.test.ts`) — the description's "81 / 28" predates the follow-up commits |
| `pnpm build` (`tsc && vite build`) | ✅ clean; only the pre-existing >500 kB chunk warning |
| zh-CN / en-US key-for-key in sync | ✅ **51/51**, zero one-sided keys, zero orphaned `t()` keys (description says 52 — off by one after `2eab840` dropped a key; parity itself holds) |
| Hidden save bar leaves the tab order | ✅ `visibility: hidden` at `admin.css:2272`, not just opacity |
| E2E 60-item save / light-dark / narrow viewport | ⛔ not reproducible here (no backend, no browser) — see the coverage note |

---

## 2. Code quality

**Quality: Approved** — no P0 and no P1. Six P2s below; none of them is a reason to hold the merge.

### The load-bearing correctness fix is right, and re-verified

I did not take the description's word for the `getFieldsValue(true)` change, because it is the one edit in this PR that could destroy 60 settings if it were wrong. All three assumptions hold in the installed `rc-field-form@2.7.1`:

1. `useForm.js:226` — `if (mergedNameList === true && !mergedFilterFunc) return this.store`. So `getFieldsValue(true)` at `index.tsx:133` really does return the whole store, not the mounted projection.
2. `useForm.js:527` — `isMergedPreserve` resolves to `preserve ?? true`, and the `<Form>` at `index.tsx:248` sets no `preserve`. Unmounting a category therefore does **not** delete its store values. This is the assumption the entire single-category design rests on.
3. `useForm.js:644-647` — `onValuesChange(changedValues, this.getFieldsValue())`: the second argument is the mounted-only projection, so dropping it and reading the store inside `handleValuesChange` (`index.tsx:136`) is required, not stylistic. `updateStore` runs before the callback, so the zero-argument handler sees post-update state — no staleness.

Net: without this change, saving from one category would have submitted every other category as an empty string. The fix is complete rather than partial, and `valuesToPayload` maps over `settings` (not over `values`), so the POST body still covers exactly the item set the server returned.

The freeze logic added in `2eab840` also holds up under inspection: `changedFieldNames` returns a fresh array each keystroke, and the grow-only reducer returns the *previous* reference when nothing was added (`index.tsx:144`), which keeps `changedGroupItems` → `visibleGroups` referentially stable. Combined with rows keyed on `settingFormName`, that is what stops the focused input from remounting mid-edit. Worth noting because it is easy to break later with an innocuous-looking refactor.

### P2 findings

**P2-1 — the desktop layout is genuinely cramped between ~992px and ~1150px, and the key line has no overflow guard.** `admin.css:2115-2152` + `admin.css:2199-2206` + `admin.css:2291`.

Rows are now always inline, the string/encrypted control is pinned at `width: 340px` (`admin.css:2149`, `flex: none`), and the single-column breakpoint is still `max-width: 991.98px` measured against the **viewport** — while `MainLayout` reserves a 220px `<Sider>` (`MainLayout.tsx:195`) and 24px `<Content>` padding (`MainLayout.tsx:343`). Working the geometry through at an expanded sider:

| Viewport | Panel | Row | Label + key column |
|---|---|---|---|
| 1024px | 526px | 486px | **132px** |
| 1152px | 654px | 614px | 260px |
| 1280px | 782px | 742px | 388px |

`.setting-row-label` ellipsizes cleanly, but `.setting-row-key` is a plain `display: block` `<code>` with **no** `overflow`/`text-overflow` (`admin.css:2199-2206`), so a key such as `sidebar.recent_filter_group_days` (~205px at 11px monospace) overflows its 132px column and paints into the control band. The old masonry did not have this problem in the same band: `column-count: 2` gave each card ~370px and string/encrypted rows used the `--block` layout (label above a full-width control), so this is a regression confined to that width range rather than a pre-existing wart.

Cheap fixes, any one of which closes it: raise the media-query breakpoint to ~1200px, switch to a container query on `.setting-panel`, make the 340px a `max-width`/`flex-basis` that can shrink, or add `overflow: hidden; text-overflow: ellipsis; white-space: nowrap` to `.setting-row-key`. This is the one I would spend time on first.

**P2-2 — reverting the last unsaved edit leaves a "phantom" Unsaved group.** `CategoryNav.tsx:67-75` + `index.tsx:82`.

The rail renders the Unsaved item only `if (changedCount > 0)`, but the panel keeps rendering it while `activeKey === CHANGED_KEY` and `changedGroupItems` is non-empty — and that group is deliberately grow-only, so it does not empty when the edit is reverted. Revert your last change while viewing the group and you get: no highlighted item anywhere in the rail, a card still titled "Unsaved changes · 1 items", and a count that contradicts the (now hidden) toolbar state. It is recoverable — clicking any category leaves — but the rail and the panel are describing different things.

This is the direct trade-off of the `2eab840` fix, and the fix is a one-liner in the opposite direction: keep the item mounted while it is selected, e.g. `{(changedCount > 0 || activeKey === CHANGED_KEY) && <NavItem … count={changedGroupCount} />}`, and title the card from the frozen membership rather than implying all of it is unsaved.

**P2-3 — `validateFields()` and the submitted payload still cover different field sets.** `index.tsx:155-158`.

`validateFields()` walks mounted field entities; `valuesToPayload(allValues(), …)` submits the whole store. Unreachable today — no setting `Form.Item` in `SettingRow.tsx` declares `rules` — and the in-code comment now records the hazard, which is a reasonable resolution for this PR. Two things still worth doing when someone adds the first rule: switch to `validateFields(Object.keys(allValues()))`, and move the `await` inside the `try`. As written it sits outside (`index.tsx:158`), so a rejection surfaces as an unhandled promise rejection with no toast. Same shape as before this PR, so not a regression.

**P2-4 — the one-frame Empty flash on first paint is not fully closed.** `index.tsx:94-105` + `index.tsx:281`.

`884895d` fixed the unconditional flash by starting `loading` at `true`, which is right. The narrower case remains and is correctly described in that commit message: `activeKey` initialises to `OVERRIDDEN_KEY` (`index.tsx:62`), the post-fetch render batches `setSettings` + `setLoading(false)` together, and `visibleGroups` yields `[[OVERRIDDEN_KEY, []]]` for one frame before the selection-validity effect re-keys. So on **any** environment with zero overrides, the very first thing painted after the fetch is "No settings in this category". Same one frame after a refresh that drops the active category, and after Discard/Save while sitting in the Unsaved group. The structural fix is to pick the fallback in the derivation instead of in an effect — e.g. have `visibleGroups` fall back to `settings[0].category` when the resolved group is empty and `settings.length > 0` — which also lets the effect go away.

**P2-5 — the category override dot is pointer-only.** `CategoryNav.tsx:36-44`.

"This category has overridden settings" is conveyed by a 5px coloured `<span>` whose only text is a `title` attribute. `title` on a non-interactive span is not keyboard-reachable and is inconsistently surfaced by assistive tech, so the information is effectively mouse-only. The button already has a text label and a count; appending visually-hidden text (or folding the state into the button's `aria-label`) costs one span. Related: the warning dot on the Unsaved item is passed `dot="warning"` with no `dotTitle` (`CategoryNav.tsx:68-74`), so it has no accessible name at all.

**P2-6 — two Save buttons, two different guards; and every row title is now a tab stop.** `index.tsx:242` vs `index.tsx:330`; `SettingRow.tsx:153`.

The toolbar Save carries `disabled={loading || settings.length === 0}`; the floating bar's Save carries neither. In practice the bar is only reachable when `dirty` (which implies a loaded `settings`) and is `visibility: hidden` otherwise, so this is not exploitable today — but it does mean clicking Save in the bar during an in-flight Refresh races the refetch. Worth making the two buttons share one `disabled` expression so the invariant is stated once.

Separately, `tabIndex={description ? 0 : undefined}` makes **every** row title focusable, not just clipped ones, so a category adds ~10 tab stops whose tooltip repeats text that is already fully visible; and because the trigger set includes `focus`, a mouse click on a title focuses the span and leaves the tooltip open until focus moves elsewhere. Gating on `hasMoreDetail` would cover the sentence-extraction case that prompted the change while dropping the redundant stops — at the cost of still missing the CSS-only truncation case, which is the trade the current code deliberately makes. Flagging so the trade is a choice rather than an oversight.

### Things I checked that are clean

- **Save-path semantics** — bool tri-state, the encrypted `****` keep-mask, and int stringification are untouched (`helpers.ts:34-60`) and covered by tests; `changedFieldNames` round-trips against `valuesFromSettings` with a test pinning it (`helpers.test.ts:278-288`), which is exactly the invariant that would rot silently.
- **`settingSummary`** — leading-separator strip, bracket-depth clamp at zero, empty/blank fallback to the key, and the deliberate omission of `.` are all now test-pinned (`helpers.test.ts:122-194`). I re-derived the three malformed-bracket expectations by hand; they are correct.
- **CSS tokens** — all 15 custom properties the new rules use are defined in every theme block, so light/dark parity holds with no hardcoded colours. `color-mix(in srgb, …)` at `admin.css:2227` matches 16 pre-existing uses in the same file, so no new browser-support floor.
- **Control widths** — `.setting-row-control` targets `.ant-input-affix-wrapper` and `.ant-input-password` as well as `> .ant-input` (`admin.css:2146-2150`), which is what makes `allowClear` and `Input.Password` size correctly next to a bare `Input`. Easy to get wrong; it is right.
- **`position: fixed` over `sticky`** — the rationale checks out: `MainLayout.tsx:346` does set `overflow: 'auto'` on `<Content>`, and `fixed` is the safe choice regardless of which element ends up being the scrollport.
- **List keys / collisions** — rows keyed on `settingFormName`, cards on the group key; the virtual keys are `__changed__` / `__overridden__`, which cannot collide with a real `category`.
- **Self-containment** — nothing outside `src/pages/SystemSetting/` imports these helpers or `SettingRow`, so the removed props and renamed classes have no external blast radius.
- **Effect termination** — the selection-validity effect always falls back to a key that satisfies its own predicate, so it cannot ping-pong.

---

## 3. Overall verdict

**APPROVE.**

Scope is clean in both directions, the one change that could have caused data loss is necessary and correctly implemented (verified at the library level, not asserted), all six findings from the previous round landed or were explicitly and defensibly deferred, and build plus tests are green locally. The six P2s are follow-up material: P2-1 is a real visual regression in a common window-width band and is the one I would fix next; P2-2 is the small inconsistency the previous fix traded for a much worse one.

## 4. Suggested follow-ups

1. Move the single-column breakpoint to ~1200px (or use a container query on `.setting-panel`), and give `.setting-row-key` an ellipsis — P2-1.
2. Keep the Unsaved rail item mounted while it is the selected group — P2-2.
3. Resolve the group fallback during derivation rather than in an effect, which removes both the residual Empty flash and the effect — P2-4.
4. Give the override dot an accessible name; pass `dotTitle` for the warning dot — P2-5.
5. Share one `disabled` expression between the two Save buttons — P2-6.

## 5. Additional observations (outside this diff's remit)

- **An encrypted override cannot be cleared from this UI.** `valuesToPayload` maps a blank encrypted field to the `****` keep-mask whenever the item is configured (`helpers.ts:40-47`), and `changedFieldNames` treats blank as unchanged (`helpers.ts:143`), so there is no gesture that means "delete this secret override and follow the default". That is the documented intent of the "Leave blank to keep the current value" placeholder and it is **pre-existing** — untouched by this PR — but this PR gives bool fields an explicit "Reset to default" affordance, which makes the asymmetry more visible than it was. Worth a product decision, not a change here.
- **Refresh silently discards unsaved edits.** `index.tsx:232` calls `fetchSettings()`, which overwrites the store and clears `changedFields` with no confirmation. Identical to pre-PR behaviour, but this PR promotes "unsaved changes" to a first-class concept (count in the button, dedicated group, explicit Discard), so a Refresh that quietly throws the same edits away now reads as inconsistent.
- **A failed initial fetch shows "No settings in this category".** With `settings === []` and `loading === false`, the Empty branch (`index.tsx:281-287`) renders the category-empty copy rather than an error state; the only error signal is a transient toast. A retry affordance would be cheap.
- **Search does not match localised category titles.** `matchesSettingQuery` (`helpers.ts:119`) matches `category.key` and `description`, so `support` finds the email settings but "Email Service" / "邮件服务" does not, while the placeholder says "name or key".
- **The Test Email action is less discoverable.** `categoryActions` is keyed by real category (`index.tsx:201-203`, consumed at `:299`), so the button no longer appears on the default landing view (the Overridden group) — only when the Email Service category is selected, or in search results. Intentional given virtual groups carry no header action, but it is a discoverability change from the all-cards-visible layout.
- **Narrow-viewport rail has no scroll-into-view.** Below 992px the rail becomes a horizontally scrollable pill row (`admin.css:2294-2303`); the active pill can sit off-screen after a programmatic re-key with nothing scrolling it back.

## 6. Coverage note

What I could **not** verify, in case someone wants to close the gap:

- **Rendered appearance.** No browser was available, so P2-1's overlap is derived from the CSS plus `MainLayout`'s measured geometry, not from a screenshot. A single 1024px-wide screenshot of the Email Service category would confirm or kill it outright, and is the highest-value manual check on this PR.
- **The end-to-end 60-item save against a real backend.** I verified the mechanism at the `rc-field-form` level and via the `valuesToPayload` unit tests, but not an actual POST.
- **Light/dark rendering and the <992px pill-row degradation.**
- **The real backend `description` strings.** Whether the truncation-heavy title design reads well, and whether P2-1 bites, depends entirely on the actual length and punctuation of the 60 descriptions. Worth eyeballing the live page once.
- **The accessibility tree as an actual screen reader exposes it** — in particular antd `Tooltip`'s `aria-describedby` wiring on the now-focusable title span, and whether the tooltip text is announced at all.
- **Merge-order interaction with any sibling PR touching `admin.css` or `SystemSetting/*`.** `MERGEABLE` on this PR alone does not guarantee a green `main` after two such PRs land.
- **Concurrency:** the Refresh-then-Save-from-the-floating-bar race in P2-6 is reasoned from the code, not exercised.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #135 (octo-admin)

## Summary

This PR reworks the System Setting page from a two-column masonry of all 60+ settings into a category rail + single-panel layout: only the selected category renders, a toolbar search spans all categories, and two virtual groups ("Unsaved" / "Overridden") sit above the real categories. It adds row-level change markers, a floating save bar with discard, first-sentence row titles with tooltips, and helper tests. The risky part of such a refactor is the form-store contract (one category mounted at a time vs. a whole-store payload) and the unsaved-tracking state machine; both hold up under tracing.

## Verification

Static analysis only at head `884895d`; build and tests not executed in this environment.

- ✅ **Whole-store save payload** — only one category is mounted at a time, so `handleSave` now validates mounted fields but submits `form.getFieldsValue(true)` (`src/pages/SystemSetting/index.tsx:133`, `:161`). The full store is populated by `form.setFieldsValue(valuesFromSettings(items))` on every fetch (`:115`), and `valuesToPayload` iterates `items` (not the store), so unmounted categories are fully covered and stale store keys cannot leak into the payload.
- ✅ **Change detection round-trip** — `changedFieldNames` compares `formValueToString(store[name])` against `item.value` under the declared contract `value: string` (`src/api/system-setting.ts:9`); bool values are normalised on both sides (`'true'` vs `'1'`), encrypted fields count any non-empty input as changed, ints compare as strings. The new tests lock exactly these cases, including the freshly-loaded round-trip returning `[]` (`src/pages/SystemSetting/helpers.test.ts:277`).
- ✅ **No validation gap today** — `validateFields()` covers mounted fields only while the payload is the whole store; grep confirms no setting `Form.Item` declares `rules` (the only `rules` in the file belong to the test-email modal form, `src/pages/SystemSetting/index.tsx:357`), and the caveat is documented inline (`:155`).
- ✅ **Selection stays valid** — the validity effect (`src/pages/SystemSetting/index.tsx:96`) falls back to Overridden / first category when a virtual group empties after save or a refresh drops a category.
- ✅ **i18n parity** — en-US and zh-CN key sets are identical (checked programmatically); every `t()` key used exists in both; removed keys `badge.db` / `badge.default` have zero remaining references.
- ✅ **Hidden save bar is inert** — `visibility: hidden` + `pointer-events: none` (`src/styles/admin.css:2265` area) removes it from tab order and hit testing; `.setting-layout--with-save-bar` adds 76px bottom padding so the fixed bar never covers the last row; the narrow-screen variant pins both edges.

## Findings

No P0/P1 issues; one non-blocking P2 below.

### P2 — Empty-state copy after a failed fetch says "no settings" (`src/pages/SystemSetting/index.tsx:288`)

When `getSystemSettings()` fails, `settings` stays `[]` and `loading` becomes `false`, so the panel renders the new `Empty` state with `t('group.empty')` ("No settings in this category") even though the real cause is the fetch error already shown in the toast. The Empty state is new in this PR, so the misleading copy is introduced here; impact is minor because the toast carries the actual error. Track a fetch-error flag (or reuse the catch branch) so the failure state renders error copy / a retry affordance instead.

## Data-flow traceability

- `changedFields` ← `changedFieldNames(allValues(), settings)` ← form store populated by `valuesFromSettings(items)` from the API response; consumed by the save-bar count, row badges, nav count and `dirty`. Fetch/discard/save all reset it explicitly; `onValuesChange` (user edits only, not `setFieldsValue`) recomputes it. Verified the data really flows: round-trip equality proven by tests, no early-return/gate skips the update.
- `changedGroupItems` ← frozen `changedGroup` ∩ current `settings`; re-frozen on every entry into the Unsaved group (`onSelect`, `src/pages/SystemSetting/index.tsx:262`), grow-only while open (avoids unmounting the input being typed into), cleared on save/discard/refresh. Intentional, commented.
- `visibleGroups` ← the same `settings` array on all three branches (query filter / virtual group / category filter) — no second source is dropped, so search and category views cannot diverge from the fetched data.
- `item.value` / `configured` / `effective_value` ← `SystemSettingItem` (server contract, `value: string`); consumed with matching normalisation on both comparison sides (see Verification). No numeric/boolean type mismatch under the declared contract.
- `categoryActions[key]` ← map keyed by real categories only; virtual-group keys resolve to `undefined` → no card-extra rendered there, which is the intended behavior.

## Blind-spot checklist (C1–C6)

- **C1 dual-path parity** — checked: save↔discard both restore the store and clear `changedFields`/`changedGroup`; save-bar show↔hide toggles both visibility and pointer-events; search set↔clear via `onSelect` and the clear button; `CategoryNav` buttons set `type="button"` inside the `<Form>`, so no accidental submit path. Clear.
- **C2 control-flow ordering / reuse** — `handleSave` is reachable from two buttons (toolbar + save bar) but is one handler guarded by the `saving` flag; validate→payload ordering preserved. Clear.
- **C3 authorization boundary** — N/A: no auth/jail/tool/credential changes; the page and its `/v1/manager/...` endpoints are unchanged.
- **C4 authorization lifecycle / cascade** — N/A: no authorization code touched.
- **C5 build ≠ runtime** — per review constraints no build/tests were executed; JSX class names were cross-checked one-by-one against the changed CSS rules, and the fixed save-bar / grid layout were reasoned about at both breakpoints. Note stands: runtime not exercised here.
- **C6 governance/docs** — N/A: no governance or policy docs in the diff.
- **R3 distributed side effects** — N/A: purely client-side UI state; no notifications/events/counters emitted.

## Cross-round blocker re-check (R6)

N/A — first review round for this PR; no prior blockers to re-verify.

## Verdict: APPROVED

The refactor preserves the form-store/save contract that the masonry layout used to provide implicitly: whole-store payload, whole-store dirty tracking with correct bool/encrypted/int normalisation, and explicit reset paths on fetch/save/discard. No correctness, security, or data-integrity defects found; the single P2 (empty-state copy on fetch failure) is non-blocking and can be addressed in a follow-up.

[Octo-Q] verdict: APPROVE — 无 P0/P1：整表保存载荷（getFieldsValue(true) + valuesToPayload 按 items 迭代）与变更检测（changedFieldNames 字符串契约 + bool/encrypted/int 归一化）经数据流回溯与测试交叉验证成立；唯一 P2 为拉取失败后 Empty 文案误导，非阻塞。

PR #140, opened 2026-08-20
feat(changelog): give desktop releases their own platform lane

an9xyz:
## Why

`webPlatforms` folded `windows` / `macos` / `linux` into the `web` set, so the desktop 1.0.0 release rendered on What's New as a **Web** card titled "08-20 更新":

- the version number was dropped — a first stable release showed no `v1.0.0` and no severity tag;
- the Windows and macOS builds merged into one per-day card;
- that card carried a single Download button pointing at whichever installer came first in the feed, so a Mac visitor was handed the `.exe`;
- there was no desktop entry in the tab row at all.

Separately, "Windows 桌面端 1.0.0 版本发布" fell into 其他, because the category rules only match a prefix and this line names the platform first.

## What changed

**Desktop is its own lane.** A 桌面端 tab filtering `windows`/`macos`/`linux`, per-OS labels and icons, and a `--platform-desktop-*` token trio (teal) defined for light, dark and `prefers-color-scheme: dark`.

**One card per version, not per OS.** All builds of a version share a card — `v1.0.0` with a badge and a download link per installer — so a release that ships two binaries reads as one release. Grouping lives in `groupReleases()` (`Changelog/utils.ts`): desktop groups by version, web by day, everything else stands alone. It is keyed by map rather than "compare with the previous entry", which also stops a release landing between two web deploys from splitting that day into two cards. Cards are ordered by the date each one claims, and which binary a card links is decided by comparing timestamps — neither depends on the order the feed arrives in, since the API sorts by `updated_at`.

**Per-OS notes are shown, never merged.** `parseUpdateDesc()` is stateful — a 【…】 / `##` / `**bold**` heading owns every line below it — so unioning two OSes' notes into one document files a line only macOS wrote under whatever heading Windows ended on. `noteBlocks()` returns one block per distinct set of notes, labelled with the platforms sharing it. A single block is left unlabelled only when the platforms that filed notes account for every build: one block can also mean the others said nothing, and presenting one platform's notes as release-wide is the same misattribution.

**The card leads with the build you can run.** `detectViewerOS()` sniffs the UA; that installer becomes the only filled button and the rest stay quiet links. Mobile, ChromeOS and iPadOS-claiming-to-be-a-Mac all decline to guess, and where the OS is unknown every build is offered with equal weight. Adds `--brand-solid` / `--brand-contrast` so the filled button clears AA in both themes.

[...]

PR #141, opened 2026-08-20
feat(changelog): offer the desktop installers above the timeline

an9xyz:
Split out of #140 at review request, so this gets read on its own terms rather than as a late addition to the platform-lane change. #140 has merged; this now targets `main` and its diff is only the strip.

## Why

The installers are reachable only from the release that shipped them, so they are findable exactly as long as that release stays at the top of the page. Web ships several times a day, so a desktop release sinks below a week of day-cards within days — and a visitor who came for the app has to know to filter by 桌面端 to find it again.

## What it does

A strip above the page title offering the newest usable installer per desktop OS:

- **one button per platform that has a build** — one if only one does, none at all if none do (the component renders nothing rather than an empty shell);
- **the visitor's own platform is promoted** to the filled button, by the same `detectViewerOS()` rule the release cards use. Where the OS is unknown, or the visitor's platform has no build, every installer is offered with equal weight rather than promoting one they cannot run;
- **read from the whole feed, not the filtered view** — switching to the iOS tab should not take the installer away;
- **picked by a total order** — tier (stable, then prerelease, then a version string that cannot be read), then version, then build number, then upload time, then URL. A hotfix row for an older version added later cannot displace a newer build; a prerelease never displaces a stable release, whatever it is numbered, since a prerelease is normally numbered ahead of the stable it precedes; and the offer does not depend on the order the feed arrives in, which matters because the API sorts by `updated_at`. Only recognised suffixes count as prereleases — `1.2.3-x64` is an arch, and ranking it below a stable would offer an older installer than it names;
- **unusable URLs are skipped** via the existing `safeDownloadUrl()` guard, so a bad row leaves no dead button;
- **the version is shown only when every offered installer says the same thing** — an identical version string modulo a `v` prefix, or a shared formatted version when they differ and none is a prerelease. Otherwise no version line: `formatVersion()` drops a `-suffix`, so labelling a mixed set would badge a release candidate as the stable release above the button that hands it over.

When the newest release is the one the strip already offers, the spotlight drops its own download row instead of repeating the same buttons half a screen apart. `offeredAbove()` decides that, keyed by platform as well as URL so two platforms sharing one URL cannot cover for each other.

It is sized as a call to action rather than a caption — a 48px icon tile, a two-line block naming the product and what is on offer, full-size buttons on a brand-tinted band. The secondary installer is a bordered button at this size rather than an underlined link: underline is the right affordance inline, but reads as unfinished next to a filled button. Its links carry a `:focus-visible` outline, and the band is a labelled `<section>` since they are the first focusable things on the page.

## Verification

`npx tsc --noEmit`, `npm test` (175 passing across 15 files), `npm run build`. Coverage added here: the per-platform offer, partial and empty feeds, version-over-upload-time, prerelease-never-displaces-stable in both feed orders, prerelease-only feeds, the recognised-suffix rule, shuffled arrival orders, unsafe-URL skipping, the dedup predicate including the shared-URL case, the label contract in both platform orders, and a component test rendering the band itself. Checked in a browser with both installers, with only one, with none at all, and in light and dark themes.

## For a reviewer

`download_url` is admin-entered and this puts it in an `href` at the top of a public, unauthenticated page. Every render site goes through `safeDownloadUrl()`, which allows only `http(s)` with an authority or a path-relative value. The server still applies no scheme validation on write, and `app_version` is accepted as free text with only a `required` rule — the page is inferring release semantics from a string nobody validates. Both belong in a write-side follow-up rather than in more heuristics here.

https://claude.ai/code/session_01VUmovPkkCwhbQNRKUYAbia

### DIFF
--- a/src/pages/Changelog/DesktopDownloadBar.test.tsx
+++ b/src/pages/Changelog/DesktopDownloadBar.test.tsx
@@ -1,0 +1,90 @@
+import { act } from 'react'
+import { createRoot, type Root } from 'react-dom/client'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import type { DesktopDownload } from './utils'
+
+// The page imports antd for its shell; the band itself uses none of it.
+vi.mock('antd', async () => {
+  const React = await vi.importActual<typeof import('react')>('react')
+  const Passthrough = ({ children }: { children?: React.ReactNode }) => React.createElement('div', null, children)
+  return {
+    Tabs: Passthrough,
+    Spin: Passthrough,
+    Empty: Passthrough,
+    Tag: Passthrough,
+    Tooltip: Passthrough,
+    ConfigProvider: Passthrough,
+    theme: { darkAlgorithm: {}, defaultAlgorithm: {} },
+  }
+})
+
+const { DesktopDownloadBar } = await import('./index')
+
+const build = (os: string, app_version: string, download_url = `https://cdn.example.com/${os}`): DesktopDownload => ({
+  os,
+  app_version,
+  download_url,
+  created_at: '2026-08-20 16:00:00',
+  is_force: 0,
+  update_desc: '',
+})
+
+describe('DesktopDownloadBar', () => {
+  let host: HTMLDivElement
+  let root: Root
+
+  beforeEach(() => {
+    ;(globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true
+    host = document.createElement('div')
+    document.body.appendChild(host)
+    root = createRoot(host)
+  })
+
+  afterEach(() => {
+    act(() => root.unmount())
+    host.remove()
+  })
+
+  const render = (downloads: DesktopDownload[]) => {
+    act(() => root.render(<DesktopDownloadBar downloads={downloads} />))
+    return host.textContent ?? ''
+  }
+
+  it('labels a version every offered installer shares', () => {
+    expect(render([build('windows', '1.0.0'), build('macos', '1.0.0')])).toContain('v1.0.0 · 支持')
+  })
+
+  it('says no version when one platform is on a prerelease and another is not', () => {
+    // Both format to 1.0.0, so a formatted comparison would badge the RC as stable
+    // above the button that hands it over.
+    for (const offers of [
+      [build('windows', '1.0.0'), build('macos', '1.0.0-rc1')],
+      [build('windows', '1.0.0-rc1'), build('macos', '1.0.0')],
+    ]) {
+      const text = render(offers)
+      expect(text).not.toContain('v1.0.0')
+      expect(text).toContain('支持')
+    }
+  })
+
+  it('keeps a lone prerelease verbatim', () => {
+    expect(render([build('macos', '1.0.1-rc1')])).toContain('1.0.1-rc1 · 支持 macOS')
+  })
+
+  it('renders one link per installer, and nothing at all without one', () => {
+    render([build('windows', '1.0.0'), build('macos', '1.0.0')])
+    expect(Array.from(host.querySelectorAll('a')).map((a) => a.getAttribute('href'))).toEqual([
+      'https://cdn.example.com/windows',
+      'https://cdn.example.com/macos',
+    ])
+
+    expect(render([])).toBe('')
+  })
+
+  it('drops an installer whose URL should never reach an href', () => {
+    render([build('windows', '1.0.0', 'javascript:alert(1)'), build('macos', '1.0.0')])
+    expect(Array.from(host.querySelectorAll('a')).map((a) => a.getAttribute('href'))).toEqual([
+      'https://cdn.example.com/macos',
+    ])
+  })
+})
--- a/src/pages/Changelog/index.tsx
+++ b/src/pages/Changelog/index.tsx
@@ -79,6 +79,11 @@ const tabItems = [
 ]
 
 const css = `
+.download-link:focus-visible {
+  outline: 2px solid var(--brand-solid);
+  outline-offset: 2px;
+  border-radius: 6px;
+}
 @keyframes fadeSlideIn {
   from { opacity: 0; transform: translateY(12px); }
   to   { opacity: 1; transform: translateY(0); }
@@ -583,7 +588,7 @@ function DesktopDownloads({ builds, compact, large }: { builds: DesktopBuild[]; 
               : { ...base, color: 'var(--brand-solid)', border: '1px solid var(--brand-solid)', padding: large ? '9px 20px' : '4px 12px', borderRadius: large ? radius.md : radius.sm }
 
         return (
-          <a key={build.os} href={build.href} target="_blank" rel="noopener noreferrer" style={style}>
+          <a key={build.os} className="download-link" href={build.href} target="_blank" rel="noopener noreferrer" style={style}>
             <DownloadOutlined />
             下载 {platformConfig[build.os]?.label ?? build.os}
           </a>
@@ -742,11 +747,13 @@ function StructuredChanges({ desc }: { desc: string }) {
  * deploys within days, so the installer cannot only live where its release sits —
  * one button per OS that has a build, the visitor's own promoted.
  */
-function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
+export function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
   if (downloads.length === 0) return null
 
   const versionLabel = offerVersionLabel(downloads)
-  const platforms = downloads.map((build) => platformConfig[build.os]?.label ?? build.os).join('、')
+  // Same order the buttons take, so the line does not read "Windows、macOS" above a
+  // macOS-first row.
+  const platforms = orderBuilds(downloads, viewerOS).map((build) => platformConfig[build.os]?.label ?? build.os).join('、')
 
   return (
     // Labelled: these are the first focusable things on the page, ahead of its <h1>.
@@ -776,7 +783,8 @@ function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
       </span>
 
       <div style={{ minWidth: 0 }}>
-        <div style={{
+        <h2 style={{
+          margin: 0,
           fontSize: font.size.lg,
           fontWeight: font.weight.bold,
           color: colors.text.primary,
@@ -784,7 +792,7 @@ function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
           lineHeight: 1.3,
         }}>
           Octo 桌面端
-        </div>
+        </h2>
         <div style={{ fontSize: font.size.base, color: colors.text.secondary, marginTop: 2 }}>
           {versionLabel && `${versionLabel} · `}支持 {platforms}
         </div>
@@ -1130,6 +1138,7 @@ function LatestReleaseSpotlight({ item, severity, hideDownloads }: { item: Relea
         <ReleaseNotes blocks={blocks} />
       </div>
 
+      {(stats.added > 0 || stats.fixed > 0 || stats.changed > 0 || !hideDownloads) && (
       <div style={{
         display: 'flex',
         alignItems: 'center',
@@ -1183,6 +1192,7 @@ function LatestReleaseSpotlight({ item, severity, hideDownloads }: { item: Relea
           </a>
         ) : null}
       </div>
+      )}
     </div>
   )
 }
--- a/src/pages/Changelog/index.tsx
+++ b/src/pages/Changelog/index.tsx
@@ -30,8 +30,8 @@ import 'dayjs/locale/zh-cn'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import api from '../../api'
 import { colors, radius, space, font } from '../../styles/tokens'
-import { parseUpdateDesc, getVersionSeverity, formatVersion, parseContributors, detectViewerOS, groupReleases, orderBuilds, noteBlocks, safeDownloadUrl, desktopPlatforms } from './utils'
-import type { VersionSeverity, ChangeCategory, Contributor, ViewerOS, AppVersion, DesktopBuild, NoteBlock, ReleaseEntry } from './utils'
+import { parseUpdateDesc, getVersionSeverity, formatVersion, parseContributors, detectViewerOS, groupReleases, orderBuilds, noteBlocks, latestDesktopDownloads, offeredAbove, safeDownloadUrl, desktopPlatforms } from './utils'
+import type { VersionSeverity, ChangeCategory, Contributor, ViewerOS, AppVersion, DesktopBuild, DesktopDownload, NoteBlock, ReleaseEntry } from './utils'
 import type { PlatformKey } from '../../styles/tokens'
 import { AnalyticsPanel } from './AnalyticsPanel'
 import { useTheme } from '../../hooks/useTheme'
@@ -541,7 +541,7 @@ function EntryBadges({ entry }: { entry: ReleaseEntry }) {
  * leads and is the only filled button; otherwise every build is offered with
  * equal weight rather than guessing at them.
  */
-function DesktopDownloads({ builds, compact }: { builds: DesktopBuild[]; compact?: boolean }) {
+function DesktopDownloads({ builds, compact, large }: { builds: DesktopBuild[]; compact?: boolean; large?: boolean }) {
   const ordered = orderBuilds(builds, viewerOS)
     .map((build) => ({ ...build, href: safeDownloadUrl(build.download_url) }))
     .filter((build): build is typeof build & { href: string } => build.href !== null)
@@ -561,22 +561,26 @@ function DesktopDownloads({ builds, compact }: { builds: DesktopBuild[]; compact
         const base: React.CSSProperties = {
           display: 'inline-flex',
           alignItems: 'center',
-          gap: 4,
-          fontSize: font.size.sm,
+          gap: large ? 6 : 4,
+          fontSize: large ? font.size.base : font.size.sm,
           fontWeight: isLead ? font.weight.semibold : font.weight.medium,
           textDecoration: 'none',
           whiteSpace: 'nowrap',
         }
-        const quiet: React.CSSProperties = { color: colors.text.secondary, textDecoration: 'underline' }
+        // Underline carries the affordance inline; at button size a border reads
+        // better next to the filled one.
+        const quiet: React.CSSProperties = large
+          ? { color: colors.text.secondary, background: colors.surface.card, border: `1px solid ${colors.surface.border}`, borderRadius: radius.md }
+          : { color: colors.text.secondary, textDecoration: 'underline' }
         const style: React.CSSProperties = compact
           ? isLead || !knowsViewer
             ? { ...base, color: 'var(--brand-solid)' }
             : { ...base, ...quiet }
           : isLead
-            ? { ...base, color: 'var(--brand-contrast)', background: 'var(--brand-solid)', border: '1px solid var(--brand-solid)', padding: '5px 14px', borderRadius: radius.sm }
+            ? { ...base, color: 'var(--brand-contrast)', background: 'var(--brand-solid)', border: '1px solid var(--brand-solid)', padding: large ? '10px 22px' : '5px 14px', borderRadius: large ? radius.md : radius.sm }
             : knowsViewer
-              ? { ...base, ...quiet, padding: '5px 10px' }
-              : { ...base, color: 'var(--brand-solid)', border: '1px solid var(--brand-solid)', padding: '4px 12px', borderRadius: radius.sm }
+              ? { ...base, ...quiet, padding: large ? '10px 14px' : '5px 10px' }
+              : { ...base, color: 'var(--brand-solid)', border: '1px solid var(--brand-solid)', padding: large ? '9px 20px' : '4px 12px', borderRadius: large ? radius.md : radius.sm }
 
         return (
           <a key={build.os} href={build.href} target="_blank" rel="noopener noreferrer" style={style}>
@@ -729,6 +733,65 @@ function StructuredChanges({ desc }: { desc: string }) {
           </div>
         )
       })}
+    </div>
+  )
+}
+
+/**
+ * Get-the-app strip above the timeline. The desktop card sinks below a week of web
+ * deploys within days, so the installer cannot only live where its release sits —
+ * one button per OS that has a build, the visitor's own promoted.
+ */
+function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
+  if (downloads.length === 0) return null
+
+  const versions = Array.from(new Set(downloads.map((build) => build.app_version)))
+  const sharedVersion = versions.length === 1 ? versions[0] : null
+  const platforms = downloads.map((build) => platformConfig[build.os]?.label ?? build.os).join('、')
+
+  return (
+    <div style={{
+      display: 'flex',
+      alignItems: 'center',
+      gap: space[5],
+      flexWrap: 'wrap',
+      padding: `${space[6]}px ${space[6]}px`,
+      marginBottom: space[8],
+      background: 'var(--brand-bg)',
+      border: '1px solid color-mix(in srgb, var(--brand) 24%, transparent)',
+      borderRadius: radius.xl,
+    }}>
+      <span style={{
+        display: 'inline-flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        width: 48,
+        height: 48,
+        flexShrink: 0,
+        borderRadius: radius.lg,
+        background: colors.surface.card,
+        border: '1px solid color-mix(in srgb, var(--brand) 18%, transparent)',
+      }}>
+        <DesktopOutlined style={{ fontSize: 22, color: 'var(--brand-text-on-bg)' }} />
+      </span>
+
+      <div style={{ minWidth: 0 }}>
+        <div style={{
+          fontSize: font.size.lg,
+          fontWeight: font.weight.bold,
+          color: colors.text.primary,
+          letterSpacing: '-0.01em',
+          lineHeight: 1.3,
+        }}>
+          Octo 桌面端
+        </div>
+        <div style={{ fontSize: font.size.base, color: colors.text.secondary, marginTop: 2 }}>
+          {sharedVersion && `v${formatVersion(sharedVersion)} · `}支持 {platforms}
+        </div>
+      </div>
+
+      <span style={{ flex: 1, minWidth: space[4] }} />
+      <DesktopDownloads builds={downloads} large />
     </div>
   )
 }
@@ -990,7 +1053,7 @@ function ChangelogItem({ item, isFirst, prevVersion, prevTimeLabel }: { item: Re
   )
 }
 
-function LatestReleaseSpotlight({ item, severity }: { item: ReleaseEntry; severity: VersionSeverity }) {
+function LatestReleaseSpotlight({ item, severity, hideDownloads }: { item: ReleaseEntry; severity: VersionSeverity; hideDownloads?: boolean }) {
   const isWeb = item.os === 'web'
   const dateObj = dayjs(item.created_at)
   const blocks = useMemo(() => noteBlocks(item, viewerOS), [item])
@@ -1094,7 +1157,7 @@ function LatestReleaseSpotlight({ item, severity }: { item: ReleaseEntry; severi
           </span>
         )}
         <span style={{ flex: 1 }} />
-        {item.builds ? (
+        {hideDownloads ? null : item.builds ? (
           <DesktopDownloads builds={item.builds} />
         ) : downloadHref ? (
           <a
@@ -1153,8 +1216,19 @@ export default function Changelog() {
     return groupReleases(raw)
   }, [data, activePlatform])
 
+  // Deliberately from `data`, not `filtered`: switching to the iOS tab should not
+  // take the desktop installer away.
+  const desktopDownloads = useMemo(() => latestDesktopDownloads(data), [data])
+
   const latestItem = filtered[0] ?? null
   const restItems = filtered.slice(1)
+
+  // When the newest release is the one the bar above is already offering, the card
+  // would repeat the same two buttons half a screen apart.
+  const spotlightOfferedAbove = useMemo(
+    () => (latestItem ? offeredAbove(latestItem, desktopDownloads) : false),
+    [latestItem, desktopDownloads],
+  )
 
   const prevVersionMap = useMemo(() => {
     const map = new Map<number, string>()
@@ -1238,6 +1312,8 @@ export default function Changelog() {
 
         <main style={{ width: '100%', maxWidth: 1040, margin: '0 auto', padding: `${space[8]}px ${space[6]}px ${space[8]}px`, boxSizing: 'border-box' }}>
 
+          {!loading && <DesktopDownloadBar downloads={desktopDownloads} />}
+
           <div style={{ marginBottom: space[6] }}>
             <h1 style={{
               fontSize: font.size['2xl'],
@@ -1258,6 +1334,7 @@ export default function Changelog() {
             <LatestReleaseSpotlight
               item={latestItem}
               severity={getVersionSeverity(latestItem.app_version, prevVersionMap.get(0))}
+              hideDownloads={spotlightOfferedAbove}
             />
           )}
 
--- a/src/pages/Changelog/index.tsx
+++ b/src/pages/Changelog/index.tsx
@@ -30,7 +30,7 @@ import 'dayjs/locale/zh-cn'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import api from '../../api'
 import { colors, radius, space, font } from '../../styles/tokens'
-import { parseUpdateDesc, getVersionSeverity, formatVersion, parseContributors, detectViewerOS, groupReleases, orderBuilds, noteBlocks, latestDesktopDownloads, offeredAbove, safeDownloadUrl, desktopPlatforms } from './utils'
+import { parseUpdateDesc, getVersionSeverity, formatVersion, parseContributors, detectViewerOS, groupReleases, orderBuilds, noteBlocks, latestDesktopDownloads, offerVersionLabel, offeredAbove, safeDownloadUrl, desktopPlatforms } from './utils'
 import type { VersionSeverity, ChangeCategory, Contributor, ViewerOS, AppVersion, DesktopBuild, DesktopDownload, NoteBlock, ReleaseEntry } from './utils'
 import type { PlatformKey } from '../../styles/tokens'
 import { AnalyticsPanel } from './AnalyticsPanel'
@@ -745,17 +745,17 @@ function StructuredChanges({ desc }: { desc: string }) {
 function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
   if (downloads.length === 0) return null
 
-  const versions = Array.from(new Set(downloads.map((build) => build.app_version)))
-  const sharedVersion = versions.length === 1 ? versions[0] : null
+  const versionLabel = offerVersionLabel(downloads)
   const platforms = downloads.map((build) => platformConfig[build.os]?.label ?? build.os).join('、')
 
   return (
-    <div style={{
+    // Labelled: these are the first focusable things on the page, ahead of its <h1>.
+    <section aria-label="下载 Octo 桌面端" style={{
       display: 'flex',
       alignItems: 'center',
       gap: space[5],
       flexWrap: 'wrap',
-      padding: `${space[6]}px ${space[6]}px`,
+      padding: space[6],
       marginBottom: space[8],
       background: 'var(--brand-bg)',
       border: '1px solid color-mix(in srgb, var(--brand) 24%, transparent)',
@@ -786,13 +786,13 @@ function DesktopDownloadBar({ downloads }: { downloads: DesktopDownload[] }) {
           Octo 桌面端
         </div>
         <div style={{ fontSize: font.size.base, color: colors.text.secondary, marginTop: 2 }}>
-          {sharedVersion && `v${formatVersion(sharedVersion)} · `}支持 {platforms}
+          {versionLabel && `${versionLabel} · `}支持 {platforms}
         </div>
       </div>
 
       <span style={{ flex: 1, minWidth: space[4] }} />
       <DesktopDownloads builds={downloads} large />
-    </div>
+    </section>
   )
 }
 
--- a/src/pages/Changelog/utils.test.ts
+++ b/src/pages/Changelog/utils.test.ts
@@ -385,6 +385,20 @@ describe('latestDesktopDownloads', () => {
     expect(offers.map((build) => build.app_version)).toEqual(['1.1.0'])
   })
 
+  it('does not let a prerelease displace the stable release it shares a number with', () => {
+    const stable = release({ os: 'windows', app_version: '2.0.0', download_url: 'https://x/stable.exe', created_at: '2026-09-01 10:00:00' })
+    const beta = release({ os: 'windows', app_version: '2.0.0-beta', download_url: 'https://x/beta.exe', created_at: '2026-09-10 10:00:00' })
+
+    for (const feed of [[beta, stable], [stable, beta]]) {
+      expect(latestDesktopDownloads(feed).map((build) => build.app_version)).toEqual(['2.0.0'])
+    }
+  })
+
+  it('still offers a prerelease when it is the only build there is', () => {
+    const offers = latestDesktopDownloads([release({ os: 'macos', app_version: '1.0.0-rc1', download_url: 'https://x/rc.dmg' })])
+    expect(offers.map((build) => build.app_version)).toEqual(['1.0.0-rc1'])
+  })
+
   it('skips a build whose URL should never reach an href, rather than offering a dead button', () => {
     const offers = latestDesktopDownloads([
       release({ os: 'windows', app_version: '1.1.0', download_url: 'javascript:alert(1)', created_at: '2026-09-01 10:00:00' }),
--- a/src/pages/Changelog/utils.test.ts
+++ b/src/pages/Changelog/utils.test.ts
@@ -439,6 +439,24 @@ describe('latestDesktopDownloads', () => {
     }
   })
 
+  it('ranks only a recognised suffix as a prerelease', () => {
+    // 1.2.3-x64 is an arch, not a release candidate; ranking it below a stable would
+    // offer the older 1.2.2 installer instead of the newer build it names.
+    const offers = latestDesktopDownloads([
+      release({ os: 'windows', app_version: '1.2.3-x64', download_url: 'https://x/new.exe', created_at: '2026-09-01 00:00:00' }),
+      release({ os: 'windows', app_version: '1.2.2', download_url: 'https://x/old.exe', created_at: '2026-01-01 00:00:00' }),
+    ])
+
+    expect(offers.map((build) => build.download_url)).toEqual(['https://x/new.exe'])
+  })
+
+  it('settles rows alike down to the second on something other than arrival order', () => {
+    const a = release({ os: 'windows', app_version: '1.0.0', download_url: 'https://x/a.exe', created_at: '2026-01-01 00:00:00' })
+    const b = release({ os: 'windows', app_version: '1.0.0', download_url: 'https://x/b.exe', created_at: '2026-01-01 00:00:00' })
+
+    expect(latestDesktopDownloads([a, b])).toEqual(latestDesktopDownloads([b, a]))
+  })
+
   it('still offers a prerelease when it is the only build there is', () => {
     const offers = latestDesktopDownloads([release({ os: 'macos', app_version: '1.0.0-rc1', download_url: 'https://x/rc.dmg' })])
     expect(offers.map((build) => build.app_version)).toEqual(['1.0.0-rc1'])
@@ -495,6 +513,17 @@ describe('offerVersionLabel', () => {
     // formatVersion drops the suffix, so this would otherwise read "v1.0.0" above a
     // button handing over the release candidate.
     expect(offerVersionLabel([offer('1.0.0-rc1')])).toBe('1.0.0-rc1')
+  })
+
+  it('says nothing when one platform is on a prerelease and another is not', () => {
+    // Both format to 1.0.0, so a formatted comparison would badge the RC as stable.
+    // Asserted in both platform orders, since the label reads from the offer list.
+    expect(offerVersionLabel([offer('1.0.0'), { ...offer('1.0.0-rc1'), os: 'macos' }])).toBeNull()
+    expect(offerVersionLabel([offer('1.0.0-rc1'), { ...offer('1.0.0'), os: 'macos' }])).toBeNull()
+  })
+
+  it('says nothing when two platforms are on different prereleases', () => {
+    expect(offerVersionLabel([offer('1.0.0-rc1'), { ...offer('1.0.0-rc2'), os: 'macos' }])).toBeNull()
   })
 
   it('says nothing when the installers do not share a version', () => {
--- a/src/pages/Changelog/utils.test.ts
+++ b/src/pages/Changelog/utils.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { detectViewerOS, getVersionSeverity, groupReleases, latestDesktopDownloads, noteBlocks, offeredAbove, orderBuilds, parseContributors, parseUpdateDesc, safeDownloadUrl } from './utils'
+import { detectViewerOS, getVersionSeverity, groupReleases, latestDesktopDownloads, noteBlocks, offerVersionLabel, offeredAbove, orderBuilds, parseContributors, parseUpdateDesc, safeDownloadUrl } from './utils'
 import type { AppVersion } from './utils'
 
 describe('parseContributors', () => {
@@ -385,6 +385,51 @@ describe('latestDesktopDownloads', () => {
     expect(offers.map((build) => build.app_version)).toEqual(['1.1.0'])
   })
 
+  it('does not let a prerelease displace a stable release, whatever it is numbered', () => {
+    // A prerelease is normally numbered ahead of the stable it precedes — 2.0.1-beta
+    // exists before 2.0.1 does — so comparing numbers first hands out a beta for as
+    // long as it is the highest-numbered row.
+    const stable = release({ os: 'windows', app_version: '2.0.0', download_url: 'https://x/stable.exe', created_at: '2026-01-01 00:00:00' })
+    const laterBeta = release({ os: 'windows', app_version: '2.0.1-beta', download_url: 'https://x/beta.exe', created_at: '2026-01-02 00:00:00' })
+
+    for (const feed of [[stable, laterBeta], [laterBeta, stable]]) {
+      expect(latestDesktopDownloads(feed).map((build) => build.app_version)).toEqual(['2.0.0'])
+    }
+  })
+
+  it('reads a prefixed version the same way for the triple and the suffix', () => {
+    const offers = latestDesktopDownloads([
+      release({ os: 'macos', app_version: 'Octo 2.0.0', download_url: 'https://x/stable.dmg', created_at: '2026-01-01 00:00:00' }),
+      release({ os: 'macos', app_version: 'Octo 2.0.0-beta', download_url: 'https://x/beta.dmg', created_at: '2026-01-10 00:00:00' }),
+    ])
+
+    expect(offers.map((build) => build.download_url)).toEqual(['https://x/stable.dmg'])
+  })
+
+  it('returns the same installer whatever order the feed arrives in', () => {
+    // The API sorts by updated_at, so re-saving any row reshuffles the feed.
+    const rows = [
+      release({ os: 'windows', app_version: '1.0.0(5)', download_url: 'https://x/a.exe', created_at: '2026-01-10 00:00:00' }),
+      release({ os: 'windows', app_version: '1.0.0(9)', download_url: 'https://x/b.exe', created_at: '2026-01-02 00:00:00' }),
+      release({ os: 'windows', app_version: '1.0.0', download_url: 'https://x/c.exe', created_at: '2026-01-05 00:00:00' }),
+      release({ os: 'windows', app_version: 'latest', download_url: 'https://x/d.exe', created_at: '2026-01-03 00:00:00' }),
+    ]
+    const orders = [rows, [...rows].reverse(), [rows[2], rows[3], rows[0], rows[1]], [rows[3], rows[1], rows[2], rows[0]]]
+
+    for (const feed of orders) {
+      expect(latestDesktopDownloads(feed).map((build) => build.download_url)).toEqual(['https://x/b.exe'])
+    }
+  })
+
+  it('never lets a version it cannot read outrank one it can', () => {
+    const offers = latestDesktopDownloads([
+      release({ os: 'windows', app_version: 'latest', download_url: 'https://x/latest.exe', created_at: '2026-01-09 00:00:00' }),
+      release({ os: 'windows', app_version: '2.0.0', download_url: 'https://x/2.0.0.exe', created_at: '2026-01-01 00:00:00' }),
+    ])
+
+    expect(offers.map((build) => build.download_url)).toEqual(['https://x/2.0.0.exe'])
+  })
+
   it('does not let a prerelease displace the stable release it shares a number with', () => {
     const stable = release({ os: 'windows', app_version: '2.0.0', download_url: 'https://x/stable.exe', created_at: '2026-09-01 10:00:00' })
     const beta = release({ os: 'windows', app_version: '2.0.0-beta', download_url: 'https://x/beta.exe', created_at: '2026-09-10 10:00:00' })
@@ -436,5 +481,27 @@ describe('offeredAbove', () => {
   it('is false for a card that has no builds at all', () => {
     const [entry] = groupReleases([release({ os: 'ios', app_version: '3.4.1', download_url: 'https://apps.apple.com/x' })])
     expect(offeredAbove(entry, [offer('windows', 'https://x/a.exe')])).toBe(false)
+  })
+})
+
+describe('offerVersionLabel', () => {
+  const offer = (app_version: string) => ({ os: 'windows', app_version, download_url: 'https://x/a.exe', created_at: '2026-08-20 16:00:00', is_force: 0, update_desc: '' })
+
+  it('labels a shared stable version', () => {
+    expect(offerVersionLabel([offer('1.0.0'), { ...offer('1.0.0'), os: 'macos' }])).toBe('v1.0.0')
+  })
+
+  it('keeps a prerelease verbatim rather than badging it as the stable it is not', () => {
+    // formatVersion drops the suffix, so this would otherwise read "v1.0.0" above a
+    // button handing over the release candidate.
+    expect(offerVersionLabel([offer('1.0.0-rc1')])).toBe('1.0.0-rc1')
+  })
+
+  it('says nothing when the installers do not share a version', () => {
+    expect(offerVersionLabel([offer('1.1.0'), { ...offer('1.0.0'), os: 'macos' }])).toBeNull()
+  })
+
+  it('treats versions that format alike as shared', () => {
+    expect(offerVersionLabel([offer('1.0.0'), { ...offer('v1.0.0'), os: 'macos' }])).toBe('v1.0.0')
   })
 })
--- a/src/pages/Changelog/utils.test.ts
+++ b/src/pages/Changelog/utils.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { detectViewerOS, getVersionSeverity, groupReleases, noteBlocks, orderBuilds, parseContributors, parseUpdateDesc, safeDownloadUrl } from './utils'
+import { detectViewerOS, getVersionSeverity, groupReleases, latestDesktopDownloads, noteBlocks, offeredAbove, orderBuilds, parseContributors, parseUpdateDesc, safeDownloadUrl } from './utils'
 import type { AppVersion } from './utils'
 
 describe('parseContributors', () => {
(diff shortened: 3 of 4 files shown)


Jerry-Xin (dismissed):
## Verdict: APPROVE

Reviewed at head f573a896ea9f. Base is `claude/changelog-desktop-lane` @ 42edccc8 (= #140's current head), so this diff is exactly the strip and nothing else.

### What I verified independently (beyond reading the diff)

**URL safety — the main thing a reviewer must prove here.** This PR adds two new consumers of admin-entered `download_url`, and both go through `safeDownloadUrl()`: `latestDesktopDownloads()` filters rows before offering them, and the bar renders via `DesktopDownloads`, which re-filters every href (defense in depth). The `safeDownloadUrl` implementation itself is byte-identical to #140's 42edccc8 — this PR does not fork or weaken it. I ran the function verbatim in a Node harness against the full hostile corpus — 34/34 behaved as required:
- backslash/slash authority tricks `\/\/evil.com`, `\\evil.com`, `//evil.com`, `/\/evil.com`, `/\evil.com`, `\//evil.com`, `\/evil.com`, `\evil.com`, `///evil.com` — all rejected
- scheme tricks `javascript:` (incl. `Java<TAB>script:`, newline-split, NUL-injected, uppercase, space-prefixed), `data:`, `vbscript:`, `file://`, `ftp://`, and the scheme-without-authority forms `http:evil.com` / `https:evil.com` — all rejected
- C0-control prefixes (\u0000 / \u0001 / \t / \x7F before `//` or `javascript:`) — all rejected
- legit `http(s)` authority URLs and path-relative download paths — all allowed

**No duplication or collision with #140.** The full #140 diff (main...42edccc8) contains none of the strip symbols (`DesktopDownloadBar`, `latestDesktopDownloads`, `offeredAbove`, `DesktopDownload`, `supersedes`, `hideDownloads`) — the 42edccc8 refactor removed them cleanly and they exist only in this PR. Merge order #140 → #141 carries no same-package redeclaration risk.

**Build & tests.** `npx tsc --noEmit` clean. `vitest run`: `src/pages/Changelog/utils.test.ts` 41/41 including the 7 new tests. The 9 `src/store/feature.test.ts` failures (localStorage setup) exist identically on base 42edccc8 — pre-existing and unrelated to this PR.

**Wiring & theme.** `DesktopOutlined` / `platformConfig` / `formatVersion` are all in scope; `--brand-bg` and `--brand-text-on-bg` are defined for light and dark themes; the bar reads from `data` (not `filtered`), so switching platform tabs cannot take the installer away; the spotlight dedup (`offeredAbove`) is keyed by OS+URL as described, and `hideDownloads` only suppresses the repeated row.

### 💬 Non-blocking notes

1. **Prerelease tie-break in `supersedes()`.** Real, runtime-verified behavior: if a `2.0.0-beta` (or `1.0.0-rc1`) row is added *after* the same-numeric-version stable, the semver comparison ties (`parseSemVer` strips the suffix by design) and the timestamp tie-break promotes the prerelease — and `formatVersion` strips the suffix too, so the bar would label it `v2.0.0`. I'm not blocking on this because (a) the input is admin-curated and every URL is still `safeDownloadUrl`-gated, (b) the timeline spotlight already features the newest-by-time row on the same data, so the bar stays consistent with existing page behavior, and (c) the case this selection logic exists for — an older hotfix row displacing a newer build — works as intended and is tested. Suggested follow-up: make the comparator prerelease-aware (stable beats prerelease for equal numeric versions) and add the two cases above as tests.
2. The server still applies no scheme validation to `download_url` on write — per the PR body that is tracked in #140; keeping the client-side guard as the single choke point until then is the right call.

### ✅ Highlights

- Unsafe-URL rows are skipped at selection time, so a bad row leaves no dead button in the bar.
- Selection is deterministic: independent of feed order and of the active tab; version wins over upload time.
- The dedup predicate keys platform as well as URL, so one universal URL shared by two platforms cannot cover for the other.
- Focused tests for per-platform offer, partial and empty feeds, version-over-upload-time, unsafe-URL skipping, and the shared-URL dedup case.

an9xyz:
Thanks — took the prerelease note rather than leaving it as a follow-up, since it hands out a wrong binary rather than merely mislabelling one. Reproduced first on `f573a89`:

```
feed: windows 2.0.0-beta (uploaded 09-10), windows 2.0.0 (uploaded 09-01)
offered: 2.0.0-beta → https://x/beta.exe        // and the strip badges it "v2.0.0"
```

`962d8af`: at an equal numeric version the stable release wins regardless of upload order. A prerelease is still offered when it is the only build there is — something installable beats nothing. Both directions of the feed are pinned by tests, plus the only-a-prerelease case. 158 passing, `tsc --noEmit` clean, `npm run build` succeeds.

Sorry for the dismissed approval; this PR needs one more push regardless — it has had no `build` check at all, because `ci.yml` only triggers on PRs targeting `main` and retargeting fires an `edited` event those workflows don't listen for. Once #140 merges I'll bring `main` into this branch so the `synchronize` event actually runs CI, and re-request review then.

Your `https://` / `https:///` note on #140 is noted and not forgotten — a host check belongs in `safeDownloadUrl()`, which lives in #140's scope, and I'd rather not dismiss a five-round approval there for something you called unexploitable. I'll send it as a small follow-up PR once #140 lands, unless you'd rather have it folded in here.

---
_Generated by [Claude Code](https://claude.ai/code)_

an9xyz:
#140 merged, so this is standalone against `main` now — but it needed a rebuild rather than just a retarget, and the branch was force-pushed: `962d8af` → `8f9e2fc`.

The squash merge collapsed the lane's nine commits into one new commit on `main`, so this branch's parent history stopped being an ancestor of `main`. GitHub retargeted the base automatically, and the result was a conflicted PR showing 1082 additions across 10 files — the lane work plus the strip — because the merge base had fallen back to `e0c2bc3`. Rebuilt by cherry-picking the two strip commits onto the squashed `main`; both applied clean, no conflicts and no content change.

The diff is now what it should be:

```
src/pages/Changelog/index.tsx     | 99 +++++++++++++++++++++++++++-----
src/pages/Changelog/utils.test.ts | 91 ++++++++++++++++++++++++++++++-
src/pages/Changelog/utils.ts      | 82 ++++++++++++++++++++++++++++++
3 files changed, 260 insertions(+), 12 deletions(-)
```

This is also the first time this PR has had real CI: `ci.yml` and `check-sprint` only trigger on pull requests targeting `main`, and a base retarget fires an `edited` event that neither listens for — so the push is what actually starts them. Both are running now.

@Jerry-Xin — re-requesting your look. Versus the head you approved (`f573a89`), the only change is `8f9e2fc`, the prerelease fix you raised; everything else is the same content on a new base.

One thing I can't do from here: this PR needs its Sprint field set on the Octo Board, or `check-sprint` will block the merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

Jerry-Xin (dismissed):
Re-review (r2) at `8f9e2fc`.

## What changed since the previous review (`f573a89`)

Two content steps, byte-verified:

1. `962d8af` "fix(changelog): keep a prerelease from displacing the stable it shares a number with" — `src/pages/Changelog/utils.ts` (+15) and `src/pages/Changelog/utils.test.ts` (+14). This implements the semver-precedence note from the last round client-side: when the numeric triples tie, a stable build now beats a prerelease regardless of upload time, in either feed order.
2. A rebase onto `main` after #140 merged (the PR was auto-retargeted from `claude/changelog-desktop-lane` to `main`). The rebase is content-neutral: the full-tree diff between `962d8af` and `8f9e2fc` is 0 bytes, and the blobs of all three Changelog files hash identically across the two heads.

With #140 in `main`, the diff against `main` is exactly the strip — `src/pages/Changelog/index.tsx`, `utils.ts`, `utils.test.ts` — confirming the split is clean and nothing overlaps with what #140 already shipped.

## Verified on this head

- Logic trace of `supersedes()`: `2.0.0` vs `2.0.0-beta` ties on the triple, the new gate returns "stable wins" in both directions; both-stable and both-prerelease pairs still fall through to build number, then timestamp; unparseable versions keep the timestamp fallback. The `isPrerelease()` normalization strips parenthesized groups the same way `parseSemVer()` cleans its input, so the two cannot disagree on the triple they compare.
- `safeDownloadUrl()` is byte-identical to the previously reviewed head, where the hostile-corpus harness passed 34/34; the delta neither adds nor removes a render site feeding it.
- `npx tsc --noEmit`: clean.
- `vitest run src/pages/Changelog/utils.test.ts`: 43/43 passing (41 prior + 2 new). The new tests pin both feed orders for the tie and the prerelease-only feed.
- Full suite: 149/158 passing; all 9 failures are `src/store/feature.test.ts` (`localStorage` undefined in this environment) and they reproduce byte-identically at the previous head `f573a89` — pre-existing and unrelated to this PR, which touches nothing under `src/store/`.

## 💬 Non-blocking notes

1. **Prerelease label fidelity** (second-pass tool flagged this; verified as real, not blocking): the strip label renders `v${formatVersion(sharedVersion)}` (src/pages/Changelog/index.tsx:789), and `formatVersion()` (src/pages/Changelog/utils.ts:572) drops the `-suffix`, so a prerelease-only feed (e.g. the new test's `1.0.0-rc1`) badges as `v1.0.0`, and a higher-numbered prerelease superseding an older stable badges by its numeric triple. This follows the timeline card's pre-existing badge convention (src/pages/Changelog/index.tsx:1014 and :1113, both predating this PR) and is label-only — the href still points at exactly the file the admin filed, and this delta strictly reduces misbadging by making stable win ties. If you want tighter fidelity later, either preserve the suffix in this label or drop the shared-version label when it is a prerelease; it could ride along with the server-side version validation follow-up noted on #140.
2. **Prerelease-vs-prerelease ordering** (second-pass tool flagged this too; verified as real, not blocking): when two rows for one OS share a numeric triple and are both prereleases (`2.0.0-beta` vs `2.0.0-rc1`), `supersedes()` falls through to upload time, so a later-uploaded older prerelease can displace the newer one. This is strictly milder than the case this commit fixes — no stable build is ever displaced, and no stable is mislabelled — and it only triggers on an admin-curated feed holding two same-number prereleases for one OS at once. If full semver identifier precedence is wanted, it fits the same server-side validation follow-up as note 1.

## ✅ Highlights

- Minimal, targeted fix: 15 lines placed inside the existing tie path, with a comment explaining why the strip's numeric labelling makes the tie dangerous.
- Tests pin both feed orders plus the lone-prerelease case.
- The rebase onto post-#140 `main` is byte-neutral, keeping the review history trustworthy.

Verdict: APPROVE.

an9xyz:
Thanks. Both notes are real and both go into the follow-up rather than a third push here, since neither is reachable on the current feed — no desktop row carries a prerelease version today:

1. **Label fidelity** — `formatVersion()` drops the `-suffix`, so a prerelease-only offer badges `v1.0.0` while the href points at the rc. Label-only, and the timeline cards have badged this way since before either PR, so fixing it in one place and not the other would be worse than leaving both. I'd rather change `formatVersion()` once, with the cards and the strip moving together.
2. **Prerelease vs prerelease** — same numeric triple, both prereleases, falls through to upload time. Agreed it's strictly milder: no stable is ever displaced. Full identifier precedence belongs with the same change as (1).

Queued behind the P2s already tracked on #140 (blank-description re-upload, the 必须升级 badge label, the `1.0.0-rc1` assertion, `/i` on the version predicate, the `new URL()` host check, and a `### Security` line in `CHANGELOG.md`).

CI is green on `8f9e2fc` — 14 checks including `build (18)` / `build (20)` and `check-sprint`. GitHub still reports the PR as blocked, which I read as a second approval being required; #140 carried two when it merged.

---
_Generated by [Claude Code](https://claude.ai/code)_

yujiawei (changes requested):
# Code Review — PR #141 (octo-admin)

Reviewed at head `8f9e2fcb82c62e66a1cf6b24e3cecade311563d3` against `origin/main`, which already contains #140 — so the diff really is only the strip (3 files, +260 / −12).

The shape of the feature is right, and pulling the selection rules out into pure, testable helpers is the right call: `latestDesktopDownloads()` and `offeredAbove()` are exactly the two decisions worth pinning in tests, and the "read from `data`, not `filtered`" comment at `index.tsx:1219` earns its keep. Everything below lands in one place — the version comparison in `supersedes()`.

## What I verified locally

- `npx tsc --noEmit` — clean.
- `npx vitest run` — 158/158 pass (43 in `Changelog/utils.test.ts`).
- `npm run build` — succeeds.
- Untrusted `download_url` reaching an `href`: guarded twice, once when the offer is selected (`utils.ts:286`) and again at render (`index.tsx:546`). I could not find a scheme, control-character, or protocol-relative bypass in `safeDownloadUrl()`, and every `target="_blank"` link on this page carries `rel="noopener noreferrer"`. The claim in the PR description holds.
- Empty and partial feeds render nothing rather than an empty shell (`index.tsx:746`); the `offeredAbove()` dedup predicate holds for the shared-URL case, and it correctly compares pre-sanitisation strings on both sides, so a URL with surrounding whitespace still dedups (I checked this specifically — both `latestDesktopDownloads()` at `utils.ts:293` and `groupReleases()` at `utils.ts:445` copy `item.download_url` verbatim from the same row).
- `large` is a new optional prop, so the two existing `DesktopDownloads` call sites (`index.tsx:1048`, `index.tsx:1161`) are unaffected.

---

## P1 — a prerelease outranks the stable release, and is then relabelled as stable

`supersedes()` compares the numeric triple first (`utils.ts:249-252`) and only consults the prerelease flag once the triples tie (`utils.ts:257-259`). But a prerelease is normally numbered *ahead* of the last stable — you tag `2.0.1-beta` before `2.0.1` exists. So the guard closes the window that closes by itself (once the same-numbered stable ships) and leaves open the long one:

```js
latestDesktopDownloads([
  { os: 'windows', app_version: '2.0.0',      download_url: '…/stable.exe', created_at: '2026-01-01 00:00:00' },
  { os: 'windows', app_version: '2.0.1-beta', download_url: '…/beta.exe',   created_at: '2026-01-02 00:00:00' },
])
// → offers 2.0.1-beta
```

The strip then labels it through `formatVersion()` (`index.tsx:789`), which drops the `-beta` (`utils.ts:572-578`):

```jsx
{sharedVersion && `v${formatVersion(sharedVersion)} · `}支持 {platforms}
```

so the band reads **v2.0.1** above a beta installer. The same holds for a lone prerelease with no stable anywhere in the feed: `2.0.0-beta` is offered — correctly, it is all there is — and rendered as **v2.0.0**. The test at `utils.test.ts:404` pins that the lone prerelease *is offered*, but not what it gets *called*, which is the half that matters here.

This is the outcome the comment at `utils.ts:253-256` already declares unacceptable — "offering the prerelease would give everyone a beta badged as v2.0.0". I'm raising it as blocking rather than as a nit because the fix is a few lines, and because the change is precisely what turns this from a cosmetic slip into a real one: on a release card a mislabelled beta was one row among many, whereas here it becomes a permanent, page-top call to action on a public unauthenticated page, offered to every visitor for as long as that row is the highest-numbered one.

**Suggested fix.** Move the prerelease test ahead of the triple comparison: prefer a stable over any prerelease regardless of number, compare numerically only within the same class, and keep the existing "offer it when it is the only build there is" fallback. Then make the label honest for the case that survives — carry the raw `app_version` (or drop the version line) when the offered build is a prerelease.

---

## P2 — `supersedes()` is not a transitive comparator, so the offer depends on feed order

`latestDesktopDownloads()` folds over the feed with `supersedes()` (`utils.ts:287-296`), but that predicate mixes three incomparable orderings — numeric triple, build number, timestamp — and falls through to `created_at` whenever the higher-priority test is inconclusive (`utils.ts:265`). A fold over a non-transitive comparator gives an order-dependent result. Two reproductions, both run against this head:

```js
// (a) build number present on some rows, absent on others
const A = { os:'windows', app_version:'1.0.0(5)', download_url:'…/a.exe', created_at:'2026-01-10 00:00:00' }
const B = { os:'windows', app_version:'1.0.0(9)', download_url:'…/b.exe', created_at:'2026-01-02 00:00:00' }
const C = { os:'windows', app_version:'1.0.0',    download_url:'…/c.exe', created_at:'2026-01-05 00:00:00' }
latestDesktopDownloads([A, B, C])  // → c.exe
latestDesktopDownloads([C, A, B])  // → b.exe

// (b) one row whose version does not parse
const A = { os:'windows', app_version:'1.0.0',  download_url:'…/a.exe', created_at:'2026-01-05 00:00:00' }
const B = { os:'windows', app_version:'2.0.0',  download_url:'…/b.exe', created_at:'2026-01-01 00:00:00' }
const C = { os:'windows', app_version:'latest', download_url:'…/c.exe', created_at:'2026-01-03 00:00:00' }
latestDesktopDownloads([A, B, C])  // → c.exe   ("latest" wins)
latestDesktopDownloads([C, A, B])  // → b.exe   (2.0.0 wins)
```

Same rows, same OS, different installer — decided by arrival order. That contradicts the PR description's "picked by version, not upload time", and more to the point it breaks the invariant the neighbouring function documents for itself at `utils.ts:430-431`: *"neither depends on the order `raw` arrives in."* It is reachable without malformed data, too — `groupReleases()`'s own comment at `utils.ts:507-510` notes that the API sorts by `updated_at`, so an admin re-saving an old row to fix a typo reorders the feed.

Worth making `supersedes()` a genuine total order: rank into comparable classes (parsed-stable, parsed-prerelease, unparsed), compare within a class, and use a deterministic final tie-break rather than letting `created_at` arbitrate across classes.

## P2 — parse leniency is asymmetric between `parseSemVer()` and `isPrerelease()`

`parseSemVer()` matches `\d+\.\d+` anywhere in the string (`utils.ts:161`), while `PRERELEASE_SUFFIX` requires the string to *start* with the numeric part (`utils.ts:234`). Any version with a prefix therefore parses as a version but is invisible to the prerelease guard:

```js
latestDesktopDownloads([
  { os:'macos', app_version:'Octo 2.0.0',      download_url:'…/stable.dmg', created_at:'2026-01-01 00:00:00' },
  { os:'macos', app_version:'Octo 2.0.0-beta', download_url:'…/beta.dmg',   created_at:'2026-01-10 00:00:00' },
])  // → beta.dmg
```

and a wholly malformed row can outrank a real release on the strength of digits that happen to appear in it — `garbage-999.0.0-beta` beats `2.0.0`. `app_version` is free text entered through the admin, so anchoring `parseSemVer()` for this path (or validating on write) is the durable fix.

## P2 — test gaps around the rules the PR claims

The seven new cases are well chosen for the happy path, but the ones that would have caught the above are missing: a higher-numbered prerelease against a stable; the *label* a lone prerelease renders with; build-number-present vs build-number-absent; and a feed-order-independence property (same rows, shuffled, same offer). There is also no component-level test for `DesktopDownloadBar` even though the repo already does component tests under jsdom (`src/pages/Dashboard/index.test.tsx`) — one render assertion on the version line is what turns the P1 above into a caught regression.

---

## Nits (non-blocking)

- `index.tsx:757` — the padding template literal passes `space[6]` twice; a two-value shorthand with both values equal is just the one value.
- `index.tsx:748-749` — `sharedVersion` de-duplicates raw `app_version` strings, so two platforms on `1.0.0` and `v1.0.0` suppress the version line even though both format to `1.0.0`. Comparing `formatVersion()` output would match the stated rule ("shown when every offered installer shares one") more faithfully.
- `index.tsx:1315` — the band's focusable links precede the page's `<h1>` and sit in no labelled section. Inside `<main>` this is legal, but a `<section aria-label="…">` (or moving it below the title) would give keyboard and screen-reader users an anchor for what is now the first interactive thing on the page.

## Out of scope, noted only

- `index.tsx:1201` — `catch { setData([]) }` turns a fetch failure into an empty successful feed, so a backend outage renders as "no releases" rather than an error. This line is unchanged by this PR and predates it; flagging it so it isn't lost, not as a change request here.

## Coverage — what this review did not establish

- **Whether `-beta` / `-rc` version strings actually occur in this feed.** That single fact decides whether the P1 is live or latent; I had no access to the production changelog payload. If desktop versions are always plain `X.Y.Z`, the P1 is a guard against a future row rather than a present bug — but the code already carries the guard for the tie case, so completing it is cheap either way.
- **Browser behaviour.** I did not exercise the page: light/dark contrast of the new `quiet` large button (`colors.text.secondary` on `colors.surface.card`), wrapping of the 48px tile + text block + two `nowrap` buttons at narrow viewports, and `color-mix()` on older browsers are all unverified here. The PR says these were checked manually.
- **Feed size.** Whether `/v1/common/changelog` caps the rows it returns is server-side and outside this repo. If it does, an old desktop release can drop out of `data` and the band silently disappears — which is the exact failure this PR exists to prevent. Worth a glance before shipping.

---

**Verdict: CHANGES_REQUESTED** — on the P1 alone. The other three findings all live in `supersedes()` and are worth folding into the same fix; the nits are yours to take or leave.

an9xyz:
You're right, and the framing is the useful part: my previous fix guarded the window that closes by itself and left the long one open. Fixed in `d308b86` — all four findings reproduced on `8f9e2fc` first:

```
P1  2.0.0 stable + 2.0.1-beta      -> beta.exe
P2a 1.0.0(5), 1.0.0(9), 1.0.0      -> c.exe   |  reordered -> b.exe
P2a 1.0.0, 2.0.0, "latest"         -> latest  |  reordered -> 2.0.0
P2b "Octo 2.0.0" + "Octo 2.0.0-beta" -> beta.dmg
```

**Tier first.** `supersedes()` ranks stable → prerelease → unreadable, and compares number, build and upload time only within a tier. That fixes the P1 by construction rather than by special case, and makes the comparator a total order, so the fold no longer depends on arrival order — both P2(a) reproductions are now stable across every permutation, pinned by a shuffle test. A prerelease is still offered when it is the only build there is.

**`isPrerelease()` is now as lenient as `parseSemVer()`.** That asymmetry was the whole of P2(b): recognising `Octo 2.0.0` as a version while missing `Octo 2.0.0-beta` as a prerelease compared the pair as stable-versus-nothing.

**The label no longer lies.** `offerVersionLabel()` keeps a prerelease's version string verbatim, so the band reads `1.0.1-rc1 · 支持 macOS` above the button that hands over the rc. I left `formatVersion()` alone deliberately — the timeline cards badge the same way, and moving them apart is worse than moving them together, which is queued with the other version-handling follow-ups.

Tests for the gaps you named: higher-numbered prerelease in both feed orders, the prefixed-version pair, a four-row shuffle asserting one answer, an unreadable version losing to a readable one, and four cases on the label itself including the lone prerelease. That last one is the unit-level answer to your component-test point — I pulled the label decision into `offerVersionLabel()` so it is testable without rendering. A real `DesktopDownloadBar` render test is still missing and I'd rather add it with the other component tests this page needs than as a one-off here.

Nits: padding shorthand collapsed; the version line compares formatted versions so `1.0.0` and `v1.0.0` no longer read as disagreeing; the band is a labelled `<section>`.

**Not taken, with reasons:** a malformed row whose digits happen to be high (`garbage-999.0.0`) still outranks a real release — anchoring the parse would fix it but would also demote `Octo 2.0.0` to unreadable, which is worse; the durable fix is validating `app_version` on write. And the `catch { setData([]) }` you flagged as out of scope stays out of scope here.

166 tests passing, `tsc --noEmit` clean, `npm run build` succeeds, and I re-checked the strip in a browser — including the prerelease label.

---
_Generated by [Claude Code](https://claude.ai/code)_

yujiawei (approved):
# Code Review — PR #141 (octo-admin)

Reviewed at head `d308b86`, against merge-base `9af8f10` (`main`). Diff is exactly the strip: 3 files, +369 / −12.

Verified locally on this head, in a clean install:

- `npx tsc --noEmit` — clean
- `npm test` — 166 passed / 14 files (the 158 in the description plus tests that landed on `main` since)
- `npm run build` — clean
- an ad-hoc probe suite against the real `utils.ts` to confirm or refute every claim below

---

## 1. Scope

The diff implements what the description says it does, and nothing else — no new flags, fields, endpoints or config:

| Claim | Where | Verdict |
|---|---|---|
| One button per platform that has a build; nothing at all when none do | `index.tsx:746` (`if (downloads.length === 0) return null`) | ✅ |
| Viewer's own platform promoted, by the same `detectViewerOS()` rule as the cards | `index.tsx:545-549`, reuses `orderBuilds` | ✅ |
| Read from the whole feed, not the filtered view | `index.tsx:1221` — `latestDesktopDownloads(data)`, not `filtered` | ✅ |
| Picked by version, not upload time | `utils.ts:283-292` (`supersedes`) | ✅ |
| A prerelease never displaces the stable release it shares a number with | `utils.ts:259-266` (tier ahead of triple) | ✅ |
| Unusable URLs skipped rather than rendered as dead buttons | `utils.ts:328`, plus the existing guard at `index.tsx:546` | ✅ |
| Spotlight drops its own download row when already offered above, keyed by platform *and* URL | `utils.ts:350-355` | ✅ |
| Version shown only when the offered installers share one | `utils.ts:308-313` | ⚠️ see P2-1 |

**Security surface.** `download_url` is admin-entered and lands in an `href` on a public unauthenticated page, so this is the part worth being careful about. On this page there are exactly two sites that put it in an `href` — `index.tsx:546` and `index.tsx:1063` — and both go through `safeDownloadUrl()`. The new code adds **no** unguarded render site: `latestDesktopDownloads()` pre-filters at `utils.ts:328` and `DesktopDownloads` re-filters at the render. Deferring server-side scheme validation to a follow-up is reasonable given that, though it should not be deferred indefinitely — a guard that only exists on one of several clients is one client away from being bypassed.

**Consistency check on the dedup path.** I specifically went after the case where the strip and the spotlight could pick *different rows* for the same OS and same version, which would make `offeredAbove()` silently false and duplicate the buttons. It cannot happen: within one `app_version` string, tier/triple/build are identical, so `supersedes()` reduces to `created_at`, which is the same rule `groupReleases()` uses at `utils.ts:494`. Both sides land on the same row and therefore compare the same raw string. Confirmed by execution, including with whitespace-padded URLs.

---

## 2. Findings

Nothing here is a correctness, security, data-loss or build regression, and nothing here is worse than `main`. All P2 / non-blocking.

### P2-1 — `offerVersionLabel()` can badge a prerelease as the stable release (`utils.ts:308-313`)

The shared-version test runs through `formatVersion()`, which *drops* the `-suffix`, but the prerelease decision only reads `offers[0]`. So a mixed-tier set collapses to one version and is then labelled from whichever platform happens to be first in `desktopOrder`.

```ts
const versions = Array.from(new Set(offers.map((offer) => formatVersion(offer.app_version))))
if (versions.length !== 1) return null
const offered = offers[0].app_version.trim()
return isPrerelease(offered) ? offered : `v${versions[0]}`
```

Executed against the real implementation:

| Feed | Rendered label | Problem |
|---|---|---|
| win `2.0.0`, mac `2.0.0-rc1` | `v2.0.0` | the macOS button hands over an RC badged as stable |
| win `2.0.0-rc1`, mac `2.0.0` | `2.0.0-rc1` | the macOS button hands over the stable release badged as an RC |
| win `1.0.0-rc1`, mac `1.0.0-rc2` | `1.0.0-rc1` | rc2 labelled as rc1 |

This is exactly the failure the function's own docstring says it prevents, so it reads as an oversight rather than a decision. It is a **label**, not a wrong download — the installer offered is still the newest usable one per OS — and the timeline cards already badge the same way today (`index.tsx:1112`, `v${formatVersion(...)}`), which the author calls out deliberately. So this is not a regression, but it is a one-line fix worth taking:

```ts
const raw = Array.from(new Set(offers.map((offer) => offer.app_version.trim())))
if (raw.length === 1) return isPrerelease(raw[0]) ? raw[0] : `v${formatVersion(raw[0])}`
if (offers.some((offer) => isPrerelease(offer.app_version))) return null
// …existing formatVersion-equality path for the all-stable case
```

Mixed stable/prerelease across OSes is not exotic — Windows shipping `2.0.0` while macOS is still on `2.0.0-rc1` waiting on notarization is an ordinary Tuesday.

### P2-2 — `PRERELEASE_SUFFIX` treats every hyphenated suffix as a prerelease (`utils.ts:241-245`)

```ts
const PRERELEASE_SUFFIX = /\d+\.\d+(?:\.\d+)?-[0-9A-Za-z.-]+/
```

`[0-9A-Za-z.-]+` matches anything after the hyphen, so an arch or date suffix is ranked as a prerelease and loses the tier comparison to any stable row — including a much older one. Confirmed by execution:

```
isPrerelease('1.2.3-x64')        → true
isPrerelease('1.2.3-20260115')   → true

latestDesktopDownloads([
  { os: 'windows', app_version: '1.2.3-x64', download_url: '…/new.exe', created_at: '2026-09-01' },
  { os: 'windows', app_version: '1.2.2',     download_url: '…/old.exe', created_at: '2026-01-01' },
])
→ [['1.2.2', '…/old.exe']]     // the strip offers the older installer
```

Note this is strictly worse than the pre-tier behaviour for that input: the ranking that exists to protect users from betas would hand them a stale build. Nothing in the repo or the tests establishes that `app_version` never carries a non-prerelease hyphen suffix, so this is a real-if-conditional risk rather than a theoretical one. Consider narrowing to recognised prerelease tokens (`alpha|beta|rc|pre|dev|snapshot|canary|nightly`) with a documented fallback, or at minimum a test pinning whichever convention you intend.

### P2-3 — greedy paren strip can erase the version before the prerelease test (`utils.ts:244`)

`version.replace(/\(.*\)/, '')` is greedy, so it deletes from the first `(` to the last `)`:

```
isPrerelease('(a) 1.0.0-rc1 (b)') → false     // should be true
```

To be fair to the author, this is *deliberately* the same leniency as `parseSemVer` (`utils.ts:160`) and the docstring says so — and that consistency is the right call, since diverging here would create a new class of disagreement between the two. Worth noting only because this PR is the first thing that lets `parseSemVer`/`parseBuildNumber` decide **which binary a public CTA hands out**, where before they only drove a severity badge and a display string. Same functions, materially larger blast radius. Hardening them (anchor the match, non-greedy the paren strip) is now more valuable than it was, but belongs in its own change, not this one.

### P2-4 — the ranking is not quite the total order the docstring claims (`utils.ts:283-292`)

The comment argues the comparator must be total so the fold over the feed is order-independent. It isn't total: two rows identical in OS, version, build *and* `created_at` but differing in URL fall through to `return candidate.created_at > current.created_at` → `false`, so the fold keeps whichever arrived first.

```
latestDesktopDownloads([a.exe, b.exe]) → a.exe
latestDesktopDownloads([b.exe, a.exe]) → b.exe
```

The permutation test at `utils.test.ts:409` ("returns the same installer whatever order the feed arrives in") cannot catch this, because every version in that fixture has a unique maximum. Second-precision `created_at` collisions on the same OS are unlikely, so this is a docstring-vs-code gap more than a live bug — but the docstring is load-bearing here and should either be softened or the comparator given a final deterministic tiebreak (e.g. `download_url` lexicographic).

### P2-5 — the spotlight footer can render as an empty rule (`index.tsx:1131-1160`)

The footer `<div>` carrying `borderTop` + `paddingTop: space[3]` is unconditional, but its only contents are the three `stats.* > 0` spans, a flex spacer, and the download block. With `hideDownloads` true and no parsed changes, it renders as a bare horizontal rule with padding. Consider gating the whole footer on `hasStats || !hideDownloads`.

### P2-6 — accessibility and polish on the new strip

- **The visible title is a `<div>`, not a heading** (`index.tsx:781-787`). The `aria-label` on `<section>` (`index.tsx:753`) makes it reachable by landmark navigation, which is good and clearly intentional — but a screen-reader user navigating by heading skips straight past the page's primary CTA. An `<h2>` styled to match would cost nothing. (It sits above the `<h1>`, which is unusual but defensible for a CTA band; a heading there makes the outline read correctly rather than worse.)
- **No `:hover` or `:focus-visible` state.** These are inline styles and `theme.css` defines no focus rules, so the largest button on the page gives a keyboard user no focus affordance. That's a pre-existing page-wide gap, but it matters more now that this is the primary action.
- **Label order can disagree with button order.** `platforms` at `index.tsx:749` is built from `downloads`, which is in `desktopOrder`, while the buttons are reordered by `orderBuilds(…, viewerOS)`. A macOS visitor reads "支持 Windows、macOS" above a macOS-first button row. Cosmetic.
- **Layout shift.** `{!loading && <DesktopDownloadBar …>}` at `index.tsx:1315` inserts a ~100px band above the `<h1>` once the fetch resolves, pushing the whole page down. Reserving the space, or rendering it inside the existing skeleton, would avoid the jump.

---

## 3. What I could not verify

- No real browser render: no visual check, no responsive pass at narrow widths, no screen-reader run. The wrap behaviour of the `flex: 1; min-width: 16px` spacer at `index.tsx:790` is reasoned about, not observed. The description says both themes and all three feed shapes were checked by hand; I confirmed separately that every token and CSS custom property the new component uses (`--brand`, `--brand-bg`, `--brand-text-on-bg`, `--brand-solid`, `--brand-contrast`, `radius.xl/lg/md`, `space[5,6,8]`, `font.size.lg/base`) is defined in both light and dark (`theme.css:18-23`, `:138-143`, `:266-271`) and that `colors.*` are CSS-var backed, so the strip is genuinely theme-reactive rather than hardcoded light.
- No production feed sample, so the actual `app_version` naming convention in the wild is unknown — which is precisely what decides whether P2-2 ever fires.
- Server-side write validation of `download_url` is out of this diff by design.

---

## 4. Verdict

**Approve.** The change does what it says, adds nothing it wasn't asked for, and the reasoning in the new code is unusually well documented — the tier-before-triple ordering, the platform-keyed dedup, and the "read from `data`, not `filtered`" choice are all correct and all non-obvious. The tests are real: the feed-permutation and prerelease-ordering cases pin behaviour that would otherwise rot silently. Type-check, tests and build are all green on this head, and the URL guard is applied consistently at every new site.

Everything above is P2 — display-level, edge-case, or inherited from helpers this PR did not touch, and nothing is a regression against `main`. P2-1 and P2-2 are the two worth a follow-up commit; the rest can ride along whenever the file is next open.

Jerry-Xin (changes requested):
Re-review r3 at head `d308b86b1725` — delta since `8f9e2fcb82c6` is exactly one commit: `fix(changelog): rank versions by a total order, and label a prerelease honestly` (utils.ts, utils.test.ts, index.tsx). This round also adjudicates the peer REQUEST_CHANGES review (4982458611) filed at the old head; that verification is included below.

### 🔴 Blocking on this head

- 🔴 Mixed stable/prerelease offers can receive a stable label. `offerVersionLabel()` (src/pages/Changelog/utils.ts:308-313) dedupes by `formatVersion()` output — which strips the `-suffix` — and then consults only `offers[0]` for the prerelease check. A feed whose offers are Windows `1.0.0` + macOS `1.0.0-rc1` formats both to `1.0.0`, so the set has size 1 and the strip renders `v1.0.0 · 支持 Windows、macOS` while the macOS button hands over the release candidate. Reproduced against this head in both feed orders (the offer list is deterministically ordered windows→macos, so the label reads stable in either case). Two aggravating facts: (1) the buttons carry no version text of their own (`下载 Windows` / `下载 macOS`, src/pages/Changelog/index.tsx:588), so the shared line is the only version information anywhere near the download; (2) at the previous head this exact feed rendered no version line at all (raw-string dedup saw two distinct strings), so this narrow case is a label regression introduced by this commit — and it contradicts the PR's own stated contract, "the version is shown when every offered installer shares one, and left off when they differ": `1.0.0` and `1.0.0-rc1` do not share one. The reach is not hypothetical: a staggered rollout that ships a macOS RC while Windows is already stable at the same triple keeps this state up for the whole window, and this codebase's own comments document staggered platform rollouts. The fix is a few lines: require identical prerelease status (ideally the full version string modulo `v`-prefix) across every offer before labelling, else return null; add tests for the mixed case in both platform orders.

### Adjudication of the peer review at the old head (8f9e2fcb82c6)

I ran the old `supersedes()`/label path end-to-end against concrete feeds; every blocking claim reproduces on the actual old code:

- P1 (prerelease outranks stable, then gets badged as stable): valid, blocking. `2.0.0` stable + `2.0.1-beta` offers `2.0.1-beta` badged "v2.0.1" in both feed orders; prefixed `Octo 2.0.0` + `Octo 2.0.0-beta` offers the beta badged "v2.0.0" (the anchored `PRERELEASE_SUFFIX` never matched the prefixed string, so the tie-gate saw stable-vs-stable and fell through to upload time); a lone `1.0.0-beta` offers and badges as "v1.0.0". The old gate only fired on exact triple ties — the window that closes by itself once the stable ships — and left open the higher-numbered and prefixed paths.
- P2 (non-transitive comparator → order-dependent offer): valid. With `1.0.0(5)` / `1.0.0(9)` / `1.0.0` rows the old fold returns different winners for different arrival orders, because the build comparison is skipped whenever one side lacks `(n)` and falls through to `created_at`.
- P2 (parse-leniency asymmetry between `parseSemVer` and `isPrerelease`): valid — the prefixed case above, and `garbage-999.0.0-beta` outranking `2.0.0`.
- Test-gap note: valid at the old head.

For the record: my r2 APPROVE at 8f9e2fcb82c6 under-called this. I flagged the suffix-stripping label and the same-triple tie-break as non-blocking notes but did not trace the higher-numbered-prerelease or prefixed-string paths, where a beta actively displaces a stable and is badged as that stable. On the evidence above, the peer's blocking severity was right.

### What the new head fixes (verified, not assumed)

- `supersedes()` is now a genuine total order — tier (stable > prerelease > unreadable), then triple, then build number, then upload time. No prerelease outranks any stable regardless of numbering, an unreadable string can never outrank a parseable one, and the result no longer depends on feed arrival order (the API sorts by `updated_at`, so re-saving a row reshuffles the feed). Verified by running the new logic against shuffled feeds.
- `PRERELEASE_SUFFIX` is now deliberately as lenient as `parseSemVer`, so prefixed strings classify consistently; `garbage-999.0.0-beta` lands in the prerelease tier and loses to any stable.
- Prerelease-only offers keep their version string verbatim in the strip (`1.0.0-rc1`, not `v1.0.0`), so the surviving offer case is labelled honestly.
- 8 new tests pin all of the above. I read the test bodies rather than the green status: they assert the offered URL/version for higher-numbered prereleases in both feed orders, prefixed pairs, four shuffled arrival orders, the unreadable tier, and the verbatim prerelease label.
- `safeDownloadUrl()` is byte-identical to the r1/r2 version (function-level diff verified), so the hostile-corpus harness results carry over unchanged.
- `npx tsc --noEmit` clean; Changelog utils suite 51/51; full suite 157/166 with all 9 failures confined to the pre-existing `src/store/feature.test.ts` environment issue, identical at the r1/r2 heads and unrelated to this PR.
- All three peer nits are addressed: `padding: space[6]` (render-equivalent to the old two-value form), formatted-version dedup for the shared-version check, and `<section aria-label>` wrapping what are now the first focusable elements on the page.

### 💬 Non-blocking notes

- Timeline cards still badge prereleases by the stripped triple (`formatVersion` unchanged) — that is the pre-existing page-wide convention, the docstring's "move them together later" reasoning is fair, and server-side URL validation on write remains queued with the #140 follow-up.
- A component-level render assertion on the `DesktopDownloadBar` version line (the peer's suggestion) would have caught the 🔴 above at the component boundary; worth adding alongside the fix.
- `isPrerelease` is now exported but has no consumer outside utils.ts yet.

### Summary

The delta is real, focused and verified: it fixes every valid blocker the peer raised at the old head — tier-first total order, symmetric parse leniency, honest verbatim label for a prerelease-only offer — with tests that genuinely pin the behavior. One residual labelling bug remains inside the fix's own scope: a cross-platform stable/prerelease mix that formats alike gets a stable label above an RC button, in a case that showed no label before. REQUEST_CHANGES on that alone; everything else is verified clean.

yujiawei (changes requested):
# Code Review — PR #141 (octo-admin) — corrected verdict at `d308b86`

**This supersedes my Approve on this same head, submitted a minute earlier. That approval was wrong and I am withdrawing it.** I am reversing to Request Changes on P2-1 from that review, re-graded to P0 for the reason below.

## What I got wrong

I found the `offerVersionLabel()` mixed-tier bug and graded it P2 — non-blocking — on the reasoning that *"the timeline cards already badge the same way today (`index.tsx:1112`), so this is not a regression."*

That compared against the wrong baseline. For a re-review of a fix commit, the baseline that decides severity is **the previous head of this PR**, not `main`. I checked it after submitting, and it inverts the conclusion:

```
$ git show 8f9e2fc:src/pages/Changelog/index.tsx | sed -n '748,749p'
  const versions = Array.from(new Set(downloads.map((build) => build.app_version)))
  const sharedVersion = versions.length === 1 ? versions[0] : null
```

At `8f9e2fc` the dedup ran on **raw** `app_version` strings. A Windows `1.0.0` + macOS `1.0.0-rc1` feed yielded a set of size 2 → `sharedVersion = null` → **no version line rendered at all**. Correct behaviour, by omission.

`d308b86` moved that dedup to `formatVersion()` output (`utils.ts:309`), which strips the `-suffix` and collapses the pair to one entry, and then reads the prerelease flag from `offers[0]` only:

```ts
const versions = Array.from(new Set(offers.map((offer) => formatVersion(offer.app_version))))
if (versions.length !== 1) return null
const offered = offers[0].app_version.trim()
return isPrerelease(offered) ? offered : `v${versions[0]}`
```

So the strip now renders `v1.0.0 · 支持 Windows、macOS` above a macOS button that hands over the release candidate — **in a case that displayed no label one commit ago.** That is a regression introduced by this commit, inside the scope of the very fix it claims to make ("label a prerelease honestly"), and it breaks the contract stated in this PR's own description: *"the version is shown when every offered installer shares one, and left off when they differ."* `1.0.0` and `1.0.0-rc1` do not share one.

Two things make it worse rather than better:

- the buttons carry no version text of their own (`下载 Windows` / `下载 macOS`, `index.tsx:588`), so this line is the only version information anywhere near the download;
- it is on a public, unauthenticated, page-top call to action, and a staggered rollout — Windows stable at `2.0.0` while macOS is still `2.0.0-rc1` waiting on notarization — holds the bad state for the entire rollout window, not an instant.

Reproduced against this head in both feed orders (`latestDesktopDownloads` returns windows→macos deterministically, so it reads stable either way):

| Feed | Label at `8f9e2fc` | Label at `d308b86` |
|---|---|---|
| win `2.0.0`, mac `2.0.0-rc1` | *(none)* | `v2.0.0` — RC badged as stable |
| win `2.0.0-rc1`, mac `2.0.0` | *(none)* | `2.0.0-rc1` — stable badged as RC |
| win `1.0.0-rc1`, mac `1.0.0-rc2` | *(none)* | `1.0.0-rc1` — rc2 badged as rc1 |

## P0 — required change

Require identical prerelease status — ideally the identical version string modulo a `v` prefix — across every offer before labelling; otherwise return `null` and fall back to the previous head's behaviour of showing no version line.

```ts
export function offerVersionLabel(offers: DesktopDownload[]): string | null {
  const raw = Array.from(new Set(offers.map((offer) => offer.app_version.trim().replace(/^v/i, ''))))
  if (raw.length === 1) return isPrerelease(raw[0]) ? raw[0] : `v${formatVersion(raw[0])}`
  if (offers.some((offer) => isPrerelease(offer.app_version))) return null
  const formatted = Array.from(new Set(offers.map((offer) => formatVersion(offer.app_version))))
  return formatted.length === 1 ? `v${formatted[0]}` : null
}
```

Please add tests for the mixed case in **both** platform orders — the existing `offerVersionLabel` suite covers shared-stable, prerelease-only, differing-versions and format-alike, but has no mixed-tier case, which is exactly the hole. A component-level render assertion on the `DesktopDownloadBar` version line would have caught this at the boundary and is worth adding alongside.

## Everything else from my earlier review stands, unchanged and non-blocking

The rest of that review is accurate and I am not retracting any of it — P2-2 (`PRERELEASE_SUFFIX` treats any hyphen suffix as a prerelease, so `1.2.3-x64` loses to a stale `1.2.2`), P2-3 (greedy paren strip, deliberately consistent with `parseSemVer`), P2-4 (the comparator is not total on exact ties, and the permutation test cannot catch it), P2-5 (empty spotlight footer), P2-6 (heading semantics, no `:focus-visible`, label-vs-button ordering, layout shift). Verification on this head was and remains green: `npx tsc --noEmit` clean, `npm test` 166/166, `npm run build` clean, and the `download_url` → `href` guard applied at every render site with no new unguarded path.

The substance of this PR is good and the direction is right — the tier-first total order, the platform-keyed dedup and the "read from `data`, not `filtered`" choice are all correct and well argued. This is one regression inside an otherwise solid fix, and the change is a few lines.

an9xyz:
Both of you landed on the same thing and you're both right — and the regression framing is the correct one: at `8f9e2fc` that feed showed no version line, and my "compare formatted versions" nit-fix is what turned it into a wrong one. Fixed in `37da5f1`, reproduced first:

```
win 2.0.0    + mac 2.0.0-rc1  -> "v2.0.0"      RC badged as stable
win 2.0.0-rc1 + mac 2.0.0     -> "2.0.0-rc1"   stable badged as RC
win 1.0.0-rc1 + mac 1.0.0-rc2 -> "1.0.0-rc1"   rc2 badged as rc1
```

I took @yujiawei's implementation as written — it keeps the all-stable format-alike case labelled (`1.0` and `1.0.0`), which my own first attempt dropped. All three now return null, in both platform orders, with unit tests plus the component test.

**`1.2.3-x64` (P2-2) turned out to be the one I'd rank highest**, and it's now fixed too: `[0-9A-Za-z.-]+` matched an arch or date stamp, so `1.2.3-x64` was ranked a prerelease and lost to a stale `1.2.2` — the ranking that exists to keep betas away from users was handing out an older installer instead. Only `alpha|beta|rc|pre|preview|dev|snapshot|canary|nightly` count now; anything else after the hyphen is build metadata.

Also taken: the comparator settles exact ties on `download_url`, so it is a total order in fact and not only in the docstring (P2-4); the spotlight footer no longer renders as a bare rule (P2-5); and from P2-6 the title is an `<h2>`, the links have a `:focus-visible` outline, and the platform list follows button order.

**The component test exists now** — you both pointed at that boundary, and you were right that it's where this would have been caught. It asserts the version line for a shared version, mixed tiers in both platform orders, and a lone prerelease, plus one link per installer, nothing at all without one, and an unsafe URL dropped.

**Not taken, with reasons:** the greedy paren strip (P2-3) — @yujiawei's own argument that consistency with `parseSemVer` is worth more than the edge case convinced me, and it belongs with hardening both together. The layout shift (P2-6) — reserving space for a band that may legitimately not exist trades one shift for a permanent empty gap; better solved when the page gets a real skeleton.

175 tests passing (15 files), `tsc --noEmit` clean, `npm run build` succeeds.

@yujiawei — thanks for withdrawing the approval rather than letting it stand. That was the right call and it's what got the severity right.

---
_Generated by [Claude Code](https://claude.ai/code)_

Jerry-Xin (approved):
Re-review r4 at 37da5f1 (delta since d308b86: one commit, `fix(changelog): say nothing rather than badge a prerelease as the stable`). The remaining blocker from r3 is fixed and harness-verified; approving.

### ✅ Verified fixed on this head (independently re-run, not just green CI)

- 🔴→✅ Mixed stable/prerelease unified label (r3 blocker): `offerVersionLabel()` now compares raw versions (v-prefix-normalised), and if the offers disagree with any prerelease among them it returns `null` instead of collapsing `1.0.0` and `1.0.0-rc1` through `formatVersion()` and reading only `offers[0]`. I ran the real implementation against Windows `1.0.0` + macOS `1.0.0-rc1` in **both platform orders** — no label is emitted either way ([utils.ts](src/pages/Changelog/utils.ts)).
- ✅ No over-suppression: all-stable same-version feed still labels `v1.0.0`; a pure-beta feed still labels honestly as `1.0.0-beta` (never `v1.0.0`); `v1.0.0` vs `1.0.0` and `1.0` vs `1.0.0` still unify; different stable versions and different prereleases still return null. Verified by an independent 24-case harness against the built implementation — all pass.
- ✅ yujiwei's "mixed prerelease shown as stable" and "public-page mislabel / fallback" points: the fallback is now "say nothing" (subtitle renders `支持 Windows、macOS` with no version), asserted at both the unit level ([utils.test.ts](src/pages/Changelog/utils.test.ts)) and the rendered-component level ([DesktopDownloadBar.test.tsx](src/pages/Changelog/DesktopDownloadBar.test.tsx)).
- ✅ Mixed-version tests cover both platform orderings (unit and component), plus lone-prerelease-verbatim, XSS-URL drop, and empty-bar cases — test bodies read, assertions are real.
- ✅ Selection determinism strengthened: rows identical down to the second now settle on `download_url` instead of arrival order; shuffle-invariance re-verified on a mixed 4-build feed (6 permutations, one result).
- ✅ `isPrerelease()` now recognises only real prerelease tokens (alpha/beta/rc/pre/preview/dev/snapshot/canary/nightly): `1.2.3-x64` and `1.2.3-20260115` are no longer demoted below an older stable — the right call, with a regression test.
- ✅ `safeDownloadUrl()` is byte-identical to the r3-reviewed head (zero delta lines in this commit), so the previously verified hostile-URL filtering carries over unchanged.
- ✅ `tsc --noEmit` clean; Changelog suite 60/60 green (55 utils + 5 component) plus my 24-case harness. Repo-wide: the only failures are the 9 pre-existing `src/store/feature.test.ts` (probeAppBots TTL) cases, which fail identically on the merge-base — not introduced by this PR.

### 💬 Non-blocking

- The per-button version display suggested in peer review remains out: when platforms disagree, the buttons carry no version hint at all. "Say nothing" is honest and matches this PR's stated contract, and the timeline below still badges each release; a per-button version chip would be a nice follow-up but is not a correctness gap on this head.
- Corner case of the new suffix rule: a stacked suffix like `1.0.0-x86-rc1` captures only the first token (`x86`) and would rank as stable. No realistic feed does this; noting for completeness.
- An integration test for the spotlight `hideDownloads` conditional (new in this commit) would lock the empty-row guard.

### ✅ Highlights

- The fix is the minimal honest one — tighten the label contract rather than invent version strings — and ships with ordering-symmetric tests at both unit and DOM level.
- Accessibility touches (semantic `<h2>`, `:focus-visible` outline on the first focusable elements) and platform-list ordering now matching button ordering are well-considered details.

yujiawei (approved):
# Code Review — PR #141 (octo-admin)

Reviewed at head `37da5f1ce237456bb7ef0f836c9865639b23e9be`, merge-base `9af8f10` (`main`). Diff is 4 files, +519 / −13 — the three Changelog files plus the new `DesktopDownloadBar.test.tsx`.

This is the fourth review round on this PR. **The blocker from the previous round is fixed**, and I confirmed it by execution rather than by reading. Nothing I found on this head rises to blocking. See *On the round count* at the end for the one thing I would not spend a fifth round on.

## Verified locally on this head

- `npx tsc --noEmit` — clean
- `npx vitest run` — **175 passed / 15 files**, including the 5 new `DesktopDownloadBar` tests and 55 in `Changelog/utils.test.ts`. No pre-existing failures on this checkout.
- Ad-hoc probe suites executed against the real `utils.ts`, plus a real jsdom render of `DesktopDownloadBar`, so each claim below is a run result and not a trace.

**The previous round's blocker is closed.** `offerVersionLabel([{win 1.0.0}, {mac 1.0.0-rc1}])` returns `null` in both platform orders (executed). Rendered, the strip reads `Octo 桌面端 / 支持 Windows、macOS` with no version line, instead of badging the release candidate as `v1.0.0`. The three-branch rewrite at `utils.ts:319-332` is the right shape: exact-string match labels, a recognised prerelease among differing strings says nothing, and only all-stable strings fall through to a formatted comparison.

**`supersedes()` is now arrival-order independent.** I folded `latestDesktopDownloads()` over all 24 permutations of a 4-row feed mixing `1.0.0`, `1.0.0(5)`, `1.0.0(9)` and `1.0.1-rc1`; all 24 returned the same single offer. The lexicographic tuple (tier → triple → build → timestamp → URL) is a strict total order, so the fold is well defined. One consequence of the final tie-break is worth a look — P2-2.

**URL safety.** `safeDownloadUrl()` is untouched by this PR, and the PR adds no unguarded href site: `latestDesktopDownloads()` pre-filters at `utils.ts:348` and `DesktopDownloads` re-filters every href at render (`index.tsx:551`). I ran a 26-case hostile corpus against the real function — control-character-split and NUL-injected `javascript:`, `data:` / `vbscript:` / `file:` / `ftp:`, scheme-without-authority `http:host`, and every slash/backslash authority form (`//host`, `\\host`, `/\host`, `\/host`, `///host`) — **26/26 rejected**, and 6/6 legitimate http(s) and path-relative values allowed. Every `target="_blank"` link added here carries `rel="noopener noreferrer"`.

---

## 1. Spec compliance

| Stated behaviour | Where | Verdict |
|---|---|---|
| One button per platform with a build; nothing at all when none do | `index.tsx:751` (`if (downloads.length === 0) return null`) | ✅ (rendered `''` for an empty feed) |
| Viewer's platform promoted by the same `detectViewerOS()` rule as the cards | `index.tsx:550-554`, reusing `orderBuilds` | ✅ |
| Read from the whole feed, not the filtered view | `index.tsx:1231` — `latestDesktopDownloads(data)`, not `filtered` | ✅ |
| Picked by version, not upload time | `utils.ts:290-302` | ✅ |
| A prerelease never displaces the stable it shares a number with | `utils.ts:293` (tier ranked ahead of the triple) | ✅ |
| Unusable URLs skipped rather than rendered dead | `utils.ts:348` + `index.tsx:551` | ✅ |
| Spotlight drops its own row when already offered above, keyed by platform *and* URL | `utils.ts:370-374` | ✅ (with the caveat in P2-2) |
| Version shown only when every offered installer shares one | `utils.ts:319-332` | ⚠️ see P2-1 |

Nothing outside the stated feature is added — no new flag, field, endpoint or config, and the two pre-existing `DesktopDownloads` call sites are unaffected because `large` is optional.

**One scope observation, non-blocking.** The head commit is titled *"say nothing rather than badge a prerelease as the stable"*, but also carries four changes neither its message nor the PR description mentions: the `.download-link:focus-visible` ring (`index.tsx:82`, `591`), the title element becoming an `<h2>` (`index.tsx:786`), the `platforms` line re-ordered through `orderBuilds` (`index.tsx:756`), and a new conditional wrapper around the spotlight footer (`index.tsx:1141`). All four are confined to this PR's own surface, so I do not read them as scope creep — but the `<h2>` has a consequence, P2-3, and the last one is worth a sentence, P2-4.

**Spec: ✅**

---

## 2. Code quality

No P0 or P1. Everything below is P2 — worth addressing, none of it a reason to hold the merge.

### P2-1 — the prerelease token allow-list still lets an unrecognised suffix through, as both the offer and the label

`utils.ts:246-251` narrowed `isPrerelease()` from "any `-suffix`" to an allow-list:

```ts
const VERSION_SUFFIX = /\d+\.\d+(?:\.\d+)?-([0-9A-Za-z]+)/
const PRERELEASE_TOKEN = /^(alpha|beta|rc|pre|preview|dev|snapshot|canary|nightly)/i
```

The narrowing is right in itself — `1.2.3-x64` and `1.2.3-20260115` are an arch and a date stamp, and ranking those below a stable release would offer an *older* installer than the one they name, which is a worse failure than the one being fixed. But the residual is the same defect class the last two commits were about. Executed against this head:

```
isPrerelease: -alpha ✓  -beta ✓  -rc1 ✓  -preview2 ✓  -RC.2 ✓  Octo 2.0.0-beta ✓
              -b2 ✗  -a1 ✗  -m1 ✗  -eap ✗  -insider ✗  -next ✗  -unstable ✗  -experimental ✗
```

so, run against the real `latestDesktopDownloads`:

```js
latestDesktopDownloads([
  { os: 'windows', app_version: '2.0.0',     download_url: '…/stable.exe', created_at: '2026-01-01 00:00:00' },
  { os: 'windows', app_version: '2.0.1-b2',  download_url: '…/beta.exe',   created_at: '2026-01-02 00:00:00' },
])
// → offers 2.0.1-b2, and offerVersionLabel() returns "v2.0.1"
```

and rendered, the mixed-platform case that round 3 blocked on comes back verbatim for an unrecognised token:

```
DesktopDownloadBar([win 1.0.0, mac 1.0.0-b2])
→ "Octo 桌面端 v1.0.0 · 支持 Windows、macOS  下载 Windows  下载 macOS"
```

I am grading this P2, not P1, and the reasoning matters more than the number:

- The admin form is free text with a `required` rule and nothing else (`src/pages/Download/index.tsx:344`), and its placeholder is a bare `1.0.0`. Any suffix is already off-convention; a suffix outside the nine recognised words is off-convention twice over.
- No heuristic over a free-text field closes this. Widening the list to `-b2`/`-m1` re-opens `-x64`; the only real fix is upstream, not here.
- Against `main` this is not a regression — the page has no strip at all today — and against the previous head the ranking is strictly better.

**A cheap partial mitigation, if you want one in this PR:** delete the last branch of `offerVersionLabel()` (`utils.ts:330-332`) and return `null` whenever the raw strings differ at all. That closes the *label* half of this for every unrecognised token, at the cost of no longer collapsing `1.0` + `1.0.0` or `Octo 1.0.0` + `1.0.0` into one label — a much smaller loss than a version line that says `v1.0.0` above a `1.0.0-b2` installer. It does not touch the ranking half, which I think is fine as-is: an off-convention row is still the newest build the admin filed.

### P2-2 — the new URL tie-break makes the strip and the spotlight card pick different rows, so the dedup silently fails

`utils.ts:302` added a final tie-break so the fold is a total order:

```ts
if (candidate.created_at !== current.created_at) return candidate.created_at > current.created_at
return candidate.download_url > current.download_url
```

But `groupReleases()` resolves the same tie the other way — first-arrived wins (`utils.ts:524`):

```ts
} else if (item.created_at > sameOS.created_at) {
  Object.assign(sameOS, build)
}
```

Two rows for one OS, one version, the same second, different URLs, executed against this head:

| Feed order | strip offers | spotlight card links | `offeredAbove()` |
|---|---|---|---|
| `a.exe`, `b.exe` | `b.exe` | `a.exe` | **false** |
| `b.exe`, `a.exe` | `b.exe` | `b.exe` | true |

So in one arrival order the strip and the card hand out *different binaries for the same version*, and `offeredAbove()` returns false — the spotlight repeats the download row, which is precisely the duplication this PR set out to remove. One commit ago both sides kept the first-arrived row and agreed; this commit made the strip order-independent and, in doing so, desynchronised it from the card. The order-independence is worth keeping — the fix is to make `groupReleases()` resolve the tie the same way (compare `download_url` when `created_at` matches), or to have both sides share one comparator.

Narrow: it needs two rows entered in the same second for one OS and version, and the damage is cosmetic. But `latestDesktopDownloads`'s own new test only pins that the *strip* is order-independent (`utils.test.ts`, "settles rows alike down to the second on something other than arrival order") — it does not pin agreement with `groupReleases()`, which is the property `offeredAbove()` actually depends on. That is the test worth adding alongside the fix.

### P2-3 — the strip's `<h2>` lands before the page's `<h1>`

`index.tsx:786` promotes the title to an `<h2>`, and the strip renders at `index.tsx:1325`, above the page `<h1>` at `index.tsx:1328`. A jsdom render confirms one `<h2>` — "Octo 桌面端" — and the document outline therefore opens at level two before the page title. Heading navigation reads a subsection before it has a page. The `aria-label` on the `<section>` (`index.tsx:759`) already does the labelling work here, so a `<div>` with the same styling loses nothing; alternatively move the `<h1>` above the strip. Minor, but it is a change the commit made deliberately, so it is worth getting right.

### P2-4 — the empty-footer guard is half of a fix

`index.tsx:1141` now suppresses the spotlight footer when there are no stats *and* downloads are hidden. Good — that was a bordered empty bar. But the `!hideDownloads` disjunct means a card with no stats, no `builds` and an unusable `download_url` still renders the same empty bar. That case predates this PR, so it is not a regression, only an incomplete cleanup; keying the condition on whether the row actually renders anything would finish it.

### Nit — the strip's support line can name a platform it offers no button for

`index.tsx:756` builds `platforms` from all `downloads`, while `DesktopDownloads` filters each href through `safeDownloadUrl()` again at `index.tsx:551`. Rendered with an unsafe Windows URL, the strip says `支持 Windows、macOS` above a single macOS button. This is unreachable through the page, which only ever passes `latestDesktopDownloads()` output (already filtered) — but the new test at `DesktopDownloadBar.test.tsx` ("drops an installer whose URL should never reach an href") renders exactly that state and asserts only the href list, so it documents the mismatch without catching it. Deriving `platforms` from the same filtered list the buttons use would keep the component honest on its own terms.

### Nit — the greedy parenthesis strip can eat the suffix it is meant to preserve

`utils.ts:250` cleans the version with `version.replace(/\(.*\)/, '')`. That is greedy, so a string carrying a parenthesised group on *both* sides of the suffix loses the suffix entirely:

```
'1.0.0(5)-beta(6)'.replace(/\(.*\)/, '')  // → '1.0.0'   — isPrerelease() false
'1.0.0-beta(5)'                            // → '1.0.0-beta'  ✓
'1.0.0(5)-beta'                            // → '1.0.0-beta'  ✓
```

Only the two-group form breaks, and this is the pre-existing idiom shared with `parseSemVer()` (`utils.ts:160`) rather than something this PR introduced — the doc comment at `utils.ts:241-244` deliberately keeps the two in lockstep. So if you tighten it to `/\([^)]*\)/g`, tighten both together; changing only `isPrerelease()` would desynchronise them and re-open the prefixed-string hole the comment warns about. Fine to leave as is.

**Quality: Approved**

---

## 3. Verdict

**APPROVE.** Spec ✅ and no P0/P1. The feature does what it says, the selection rules are pure and pinned by tests that assert the right properties, the untrusted-URL path is guarded at every site and survives a hostile corpus, and the round-3 blocker is genuinely closed rather than moved. P2-1 and P2-2 are worth a follow-up; neither should hold this.

## 4. On the round count

This is round four, and rounds two, three and four have all been the same conversation: what a version string means when it is free text. Round two fixed the tie case, round three fixed the higher-numbered case and the label, round four narrowed the suffix rule and fixed the mixed-platform label — and P2-1 above is the same shape again.

I would not spend a fifth round widening the token list. The root cause is not in this file: `app_version` is accepted with `rules={[{ required: true }]}` and nothing else (`src/pages/Download/index.tsx:344`), so the page is reverse-engineering release semantics from a string no one validated. The durable fix is a format contract at the write side — validate or normalise `app_version` on submit and on the server — which would also retire most of the heuristic here. That belongs in the follow-up already queued for the server-side `download_url` scheme check, not in another iteration of this PR.

an9xyz:
Thanks — and agreed on the round count. Rounds 2–4 were all the same conversation, and the root cause is exactly where you put it: `app_version` is free text with a `required` rule and nothing else, so this page is inferring release semantics from a string nobody validated. No heuristic here closes that; I'm not spending a fifth round widening the token list either. Body updated to say so.

Both P2s are queued rather than pushed, and I want to be clear which one I think matters:

- **P2-2 is mine and it's the one I'd fix first.** The URL tie-break I added to make the fold order-independent desynchronised the strip from `groupReleases()`, which still keeps the first-arrived row — so in one arrival order the two hand out different binaries for the same version and `offeredAbove()` goes false, duplicating the row this PR exists to remove. The fix is to share one comparator between them, with a test pinning agreement rather than just each side's own determinism. Narrow trigger, cosmetic damage, but it's a self-inflicted inconsistency.
- **P2-1's label half** — dropping the last branch of `offerVersionLabel()` so any difference in raw strings means no label — I'll take in the same follow-up. You're right that `v1.0.0` above a `1.0.0-b2` installer is worse than losing the `1.0` + `1.0.0` collapse; I'd rather change it with the write-side validation in view than reverse the same three lines a third time in isolation.

On **P2-3**, the `<h2>`: you and @Jerry-Xin disagree here — he asked for a heading so keyboard users navigating by heading reach the page's primary CTA, you're pointing out the outline then opens at level two. Both are right, and the only way to satisfy both is putting the strip below the `<h1>`, which is the one thing its placement is for. Leaving it as a heading and noting the disagreement rather than silently picking a side.

P2-4, the platform-line nit and the greedy paren strip go in the same follow-up, along with what's already queued from #140.

175 tests across 15 files, `tsc --noEmit` clean, `npm run build` green, CI green, and both approvals now on this head.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #134, opened 2026-08-04
App Bot management lacks the scope dimension: space-scoped bots become invisible and unmanageable


StefanoWoo:
## Summary

The admin console's App Bot page only shows platform-scoped bots, and the create dialog offers no scope choice (silently defaults to `platform`). Once a bot's scope is set to `space` (which the data model fully supports: `app_bot.scope` + `app_bot.space_id`), it keeps running fine but disappears from every admin surface: the App Bot page shows "no data", and the Space detail drawer (Members / Invites / Join requests) has no App Bot tab in our deployed version.

## Steps to reproduce

1. Create an app bot in admin — note the create dialog only asks for Bot ID / display name / description / welcome message. The row is written with `scope='platform'`.
2. Set the bot to `scope='space'` + a valid `space_id` (we did this via DB; the scope values match the admin bundle's own i18n keys `detail.scope.platform` / `detail.scope.space`).
3. The bot stays connected and works in the space, but the App Bot page now shows "暂无数据" and there is no other page that lists it.

## Why it matters

Space scoping is the natural tenant boundary once a deployment hosts more than one space. Today the only way to run space-scoped bots is to manage them blind (DB/API only), so in practice everyone leaves bots at platform scope — which is exactly what you don't want on a multi-tenant deployment.

## Proposed design (happy to discuss)

Keep one unified list instead of splitting per-scope pages:

1. **App Bot page**: add an "Ownership" (归属) column showing `Platform` or the space name; add an ownership filter dropdown (All / Platform / each space) next to the existing status filter. Rename the page title from "平台应用 Bot" to "应用 Bot".
2. **Create dialog**: add a scope field — `Platform | Space` radio, with a space selector when `Space` is chosen; POST carries `scope` / `space_id`.
3. Server side: the admin list endpoint accepts an optional scope/space filter and returns space-scoped rows (with space name joined) to platform admins.

## Environment

- Self-hosted deployment from `octo-deployment` (merged upstream `3a869fd`), octo-admin bundle `index-JtgrN3tW.js`
- Single space today; preparing to open the deployment to external spaces

an9xyz:
The current implementation separates platform and Space App Bot management.

- `/admin/app-bots` manages platform-scoped bots only.
- Space-scoped bots are managed by a Space owner or admin in the separate Space Admin console, under the **App Bots** tab: `/admin/space/{space_id}/app-bots`.
[...]

PR #143, opened 2026-08-21
feat(space-console): give the space console its own sign-in

StefanoWoo:
## The problem

The space console has no way in of its own.

`SpaceEntry` can only adopt an IM session token that happens to be sitting in the same tab's `sessionStorage`, and the auth store holds a single token whose `super` / `space` scopes are mutually exclusive. Two consequences:

- Opening `/admin/space` in a fresh tab is a dead end. There is no token in that tab and no way to obtain one, so the entry point renders "no token" and the only offered action is the super admin login.
- A signed-in super admin who follows a link there is redirected to `/dashboard` with no explanation. From the outside it looks like the link is broken.

So in practice everything behind the space console — space-scoped app bots, members, invites, join requests — is unreachable, even though the server exposes the complete `/v1/space/:space_id/app_bot` surface and `checkSpaceAdmin` only asks for `space_member.role >= 1`.

We hit this while creating a space-scoped app bot on a self-hosted deployment. The bot was created on the platform App Bot page (which is the only place a create dialog exists), then moved to `scope='space'`. It immediately vanished from that page — correctly, since the platform list is scope-filtered server-side — and the space console was unreachable, so there was no page left anywhere that could publish it. The bot sat at `status=0`, `/v1/bot/register` kept answering `auth_failed`, and the only way forward was an `UPDATE` against the database. That is the part we would like to make unnecessary.

## What this changes

Frontend only. No server changes are needed — the space-scoped API already exists and is complete.

1. **The space console gets a real sign-in.** When the tab carries no session token, `SpaceEntry` renders a sign-in form and authenticates through `POST /v1/user/login`, the same endpoint the IM web client uses. It then confirms via `/v1/space/my` that the account administers at least one space and continues into the console exactly as before. The existing IM-tab flow is untouched: a session token still signs you straight in with no form.

2. **A super admin can switch identity instead of being bounced.** Arriving at `/space` while signed in as a super admin now shows the same form with a notice that continuing signs you in as a space administrator. The two identities remain separate; the difference is that the switch is explicit and explained rather than a silent redirect.

3. **The platform App Bot list says it is platform-scoped**, and links to the space console. A bot moved into a space used to disappear from that page with nothing on screen to say where it went.

4. **The super admin sidebar gains a space console entry**, so reaching it no longer requires typing the URL by hand.

## Notes

- Copy is added to both `zh-CN` and `en-US`. The three now-unused `entry.noToken.*` keys are removed.
- `src/pages/SpaceAdmin/SpaceEntry.test.tsx` covers the new behaviour; `npm test` passes 200 tests.
- Running on our deployment since 2026-08-21.

This overlaps in subject with #134 (ownership column, scope filter, scope selection in the create dialog) but is independent of it: that issue is about seeing and choosing scope, this is about being able to reach the space console at all. #134 also contains a claim of ours that we now know to be wrong — it says the deployed space drawer has no App Bot tab. The tab was there the whole time; we simply could not get into the console, and mistook unreachable for absent. We will correct that on the issue.

Happy to adjust the approach — in particular, if you would rather the space console keep out of `/v1/user/login` and obtain its token some other way, say so and we will rework it.

### DIFF
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -15,6 +15,21 @@ interface LoginResponse {
 export const login = (params: LoginParams) =>
   api.post<LoginResponse>('/v1/manager/login', params).then((res) => res.data)
 
+// 空间控制台自有登录（普通用户身份）。此前空间控制台没有登录入口，只能寄生于同一
+// 标签页 sessionStorage 里恰好存在的 IM 会话 token —— 换标签页就进不去，且与超管会话互斥。
+// 服务端 /v1/user/login 是 IM Web 用的同一个端点，空间管理员用自己的账号登录即可。
+export interface UserLoginResponse {
+  token: string
+  uid: string
+  name: string
+  username?: string
+}
+
+export const userLogin = (params: LoginParams) =>
+  api
+    .post<UserLoginResponse>('/v1/user/login', { ...params, flag: 1 })
+    .then((res) => res.data)
+
 export const getManagerMe = () =>
   api.get<ManagerMe>('/v1/manager/me').then((res) => ({
     ...res.data,
--- a/src/i18n/locales/en-US/appBots.json
+++ b/src/i18n/locales/en-US/appBots.json
@@ -93,5 +93,8 @@
   "form.welcome.extra": "Message automatically sent on a user's first connection; leave empty to use the default prompt",
   "form.welcome.placeholder": "Hi! I'm the xx assistant. How can I help you?",
   "edit.title": "Edit {{name}}",
-  "edit.toast.updated": "Updated"
+  "edit.toast.updated": "Updated",
+  "list.platformOnly.title": "This list shows platform-level app bots only",
+  "list.platformOnly.desc": "Space-level app bots, including any moved from the platform into a space, are managed in that space’s own console.",
+  "list.platformOnly.link": "Open the space console"
 }
--- a/src/i18n/locales/en-US/nav.json
+++ b/src/i18n/locales/en-US/nav.json
@@ -13,5 +13,6 @@
   "expertMarket": "Expert Market",
   "backup": "Backups",
   "download": "Downloads",
-  "systemSkill": "Skill Market"
+  "systemSkill": "Skill Market",
+  "spaceConsole": "Space console"
 }
--- a/src/i18n/locales/en-US/spaceAdmin.json
+++ b/src/i18n/locales/en-US/spaceAdmin.json
@@ -38,12 +38,19 @@
   "validate.logoTooLong": "Logo cannot exceed {{max}} characters",
   "validate.descriptionTooLong": "Space description cannot exceed {{max}} characters",
   "entry.loading": "Entering space admin…",
-  "entry.noToken.title": "No login session detected",
-  "entry.noToken.subtitle": "Please sign in within the app before entering space admin, or sign in with a super admin account.",
-  "entry.noToken.action": "Go to sign in",
   "entry.noSpace.title": "No manageable spaces",
   "entry.noSpace.subtitle": "You are not an admin or owner of any space yet.",
   "entry.noSpace.action": "Back to sign in",
   "entry.error.title": "Failed to load space",
-  "entry.error.action": "Retry"
+  "entry.error.action": "Retry",
+  "entry.login.title": "Space console",
+  "entry.login.subtitle": "Sign in with your account to manage the spaces you administer.",
+  "entry.login.superNotice": "You are signed in as a super administrator. The space console is a separate identity; signing in here switches you to a space administrator.",
+  "entry.login.username": "Username",
+  "entry.login.password": "Password",
+  "entry.login.usernameRequired": "Enter your username",
+  "entry.login.passwordRequired": "Enter your password",
+  "entry.login.submit": "Sign in",
+  "entry.login.toSuper": "Sign in as super administrator instead",
+  "entry.login.noToken": "Login returned no token"
 }
--- a/src/i18n/locales/zh-CN/appBots.json
+++ b/src/i18n/locales/zh-CN/appBots.json
@@ -93,5 +93,8 @@
   "form.welcome.extra": "用户首次连接时自动发送的消息，留空则使用默认提示",
   "form.welcome.placeholder": "你好！我是 xx 助手，有什么可以帮你的？",
   "edit.title": "编辑 {{name}}",
-  "edit.toast.updated": "已更新"
+  "edit.toast.updated": "已更新",
+  "list.platformOnly.title": "这里只列平台级应用 Bot",
+  "list.platformOnly.desc": "空间级应用 Bot（含从平台移入空间的）在各自空间的控制台管理。",
+  "list.platformOnly.link": "前往空间控制台"
 }
--- a/src/i18n/locales/zh-CN/nav.json
+++ b/src/i18n/locales/zh-CN/nav.json
@@ -13,5 +13,6 @@
   "expertMarket": "专家市场",
   "backup": "备份管理",
   "download": "下载配置",
-  "systemSkill": "Skill 市场"
+  "systemSkill": "Skill 市场",
+  "spaceConsole": "空间控制台"
 }
--- a/src/i18n/locales/zh-CN/spaceAdmin.json
+++ b/src/i18n/locales/zh-CN/spaceAdmin.json
@@ -38,12 +38,19 @@
   "validate.logoTooLong": "Logo 不能超过 {{max}} 个字符",
   "validate.descriptionTooLong": "空间描述不能超过 {{max}} 个字符",
   "entry.loading": "正在进入空间管理…",
-  "entry.noToken.title": "未检测到登录态",
-  "entry.noToken.subtitle": "请先在应用内登录后再进入空间管理，或使用超级管理员账号登录。",
-  "entry.noToken.action": "去登录",
   "entry.noSpace.title": "暂无可管理的空间",
   "entry.noSpace.subtitle": "你还不是任何空间的管理员或拥有者。",
   "entry.noSpace.action": "返回登录",
   "entry.error.title": "加载空间失败",
-  "entry.error.action": "重试"
+  "entry.error.action": "重试",
+  "entry.login.title": "空间控制台",
+  "entry.login.subtitle": "用你的账号登录，管理你是管理员的空间。",
+  "entry.login.superNotice": "你当前是超级管理员身份。空间控制台是另一套身份，登录后将切换为空间管理员。",
+  "entry.login.username": "用户名",
+  "entry.login.password": "密码",
+  "entry.login.usernameRequired": "请输入用户名",
+  "entry.login.passwordRequired": "请输入密码",
+  "entry.login.submit": "登录",
+  "entry.login.toSuper": "改用超级管理员登录",
+  "entry.login.noToken": "登录未返回 token"
 }
--- a/src/layouts/MainLayout.tsx
+++ b/src/layouts/MainLayout.tsx
@@ -120,6 +120,11 @@ const MainLayout: React.FC = () => {
     if (appBotsAvailable === true) {
       list.push({ key: '/app-bots', icon: <RobotOutlined />, label: t('nav:appBots'), group: 'management' })
     }
+    if (appBotsAvailable === true) {
+      // 空间级 Bot / 成员 / 邀请在各空间自己的控制台管理，超管这边只能看到平台级。
+      // 没有这个入口时，空间控制台只能靠手敲 URL 进入。
+      list.push({ key: '/space', icon: <AppstoreOutlined />, label: t('nav:spaceConsole'), group: 'management' })
+    }
     if (hasManagerCapability(managerCapabilities, 'system_setting')) {
       list.push({ key: '/system-setting', icon: <SettingOutlined />, label: t('nav:systemSetting'), group: 'system' })
     }
--- a/src/pages/AppBots/index.tsx
+++ b/src/pages/AppBots/index.tsx
@@ -10,6 +10,7 @@ import {
   message,
   Typography,
   Avatar,
+  Alert,
 } from 'antd'
 import { PlusOutlined, SearchOutlined, RobotOutlined } from '@ant-design/icons'
 import type { ColumnsType } from 'antd/es/table'
@@ -247,6 +248,24 @@ export default function AppBotsPage({ spaceId }: Props) {
           </Button>
         </Space>
       </div>
+
+      {!spaceId && (
+        // 这张表只列平台级 Bot（服务端 /v1/admin/app_bot 按 scope 过滤，见 botInRouteScope
+        // 的跨租户防护）。空间级 Bot 在各自空间的控制台管理。没有这句提示时，
+        // 一个被移到空间的 Bot 会从这页凭空消失，用户无从知道它去哪了。
+        <Alert
+          type="info"
+          showIcon
+          style={{ marginBottom: 16 }}
+          message={t('list.platformOnly.title')}
+          description={
+            <span>
+              {t('list.platformOnly.desc')}{' '}
+              <Typography.Link href="/admin/space">{t('list.platformOnly.link')}</Typography.Link>
+            </span>
+          }
+        />
+      )}
 
       <Table
         rowKey="id"
--- a/src/pages/SpaceAdmin/SpaceEntry.test.tsx
+++ b/src/pages/SpaceAdmin/SpaceEntry.test.tsx
@@ -1,0 +1,76 @@
+import { describe, expect, it } from 'vitest'
+import enUS from '../../i18n/locales/en-US/spaceAdmin.json'
+import zhCN from '../../i18n/locales/zh-CN/spaceAdmin.json'
+import navEN from '../../i18n/locales/en-US/nav.json'
+import navZH from '../../i18n/locales/zh-CN/nav.json'
+import appBotsEN from '../../i18n/locales/en-US/appBots.json'
+import appBotsZH from '../../i18n/locales/zh-CN/appBots.json'
+// 用 Vite 的 ?raw 读源码（不引 node API，`tsc` 在 build 时也编译测试文件，
+// 而本仓库没有 @types/node）。断言的是「这段行为写在了源码里」，不是实现细节的复刻。
+import entrySource from './SpaceEntry.tsx?raw'
+import listSource from '../AppBots/index.tsx?raw'
+
+describe('space console sign-in', () => {
+  // The space console used to have no sign-in of its own: it could only borrow an IM
+  // session token that happened to sit in the same tab's sessionStorage. Open it in a
+  // fresh tab and it was a dead end; arrive as a super admin and it silently bounced
+  // you back to the super dashboard. Both paths now land on a real sign-in form.
+  it('signs in through the user login endpoint rather than borrowing a session token', () => {
+    expect(entrySource).toContain("import { userLogin } from '../../api/auth'")
+    expect(entrySource).toContain('async function handleLogin')
+    expect(entrySource).toContain('await userLogin(values)')
+  })
+
+  it('offers the form when the tab carries no session token', () => {
+    expect(entrySource).toMatch(/const token = readSessionToken\(\)\s*\n\s*if \(!token\) \{[\s\S]*?setStatus\('form'\)/)
+    expect(entrySource).not.toContain("setStatus('no-token')")
+  })
+
+  it('lets a super admin switch identity instead of bouncing them away', () => {
+    expect(entrySource).toMatch(/scope === 'super'[\s\S]*?setCameFromSuper\(true\)[\s\S]*?setStatus\('form'\)/)
+    expect(entrySource).not.toMatch(/scope === 'super'[\s\S]{0,120}navigate\('\/dashboard'/)
+    expect(entrySource).toContain("t('entry.login.superNotice')")
+  })
+
+  it('keeps sign-in copy in both locales', () => {
+    const keys = [
+      'entry.login.title',
+      'entry.login.subtitle',
+      'entry.login.superNotice',
+      'entry.login.username',
+      'entry.login.password',
+      'entry.login.usernameRequired',
+      'entry.login.passwordRequired',
+      'entry.login.submit',
+      'entry.login.toSuper',
+      'entry.login.noToken',
+    ] as const
+    for (const key of keys) {
+      expect(zhCN[key], `zh-CN ${key}`).toBeTruthy()
+      expect(enUS[key], `en-US ${key}`).toBeTruthy()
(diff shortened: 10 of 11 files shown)


Jerry-Xin (changes requested):
The feature is relevant and well-motivated, but the new sign-in flow is broken for authentication failures by the shared API interceptor.

## 🔴 Blocking

- 🔴 Critical — Invalid space-console credentials redirect users to the super-admin login instead of displaying the form error. `userLogin()` uses the shared Axios client ([src/api/auth.ts:28](src/api/auth.ts:28)), whose interceptor logs out and assigns `/admin/login` for every 401 ([src/api/index.ts:37](src/api/index.ts:37)). Consequently, the catch block intended to show `formError` ([src/pages/SpaceAdmin/SpaceEntry.tsx:150](src/pages/SpaceAdmin/SpaceEntry.tsx:150)) cannot preserve the page. It also destroys an existing super-admin session after a mistyped password. Exempt login requests from global 401 redirection, or use a dedicated unauthenticated client and handle the error locally.

## 💬 Non-blocking

- 🟡 Warning — The shared request interceptor attaches the current super-admin token to `/v1/user/login` ([src/api/index.ts:25](src/api/index.ts:25)). Authentication endpoints should normally omit stale or differently scoped credentials.

- 🟡 Warning — The tests inspect source text with regexes rather than render the component or mock API behavior. This misses the blocking interceptor interaction. Add behavioral tests covering invalid credentials, successful login, no managed spaces, and switching from a super-admin session.

- 🔵 Suggestion — The sidebar entry is conditional on App Bot availability ([src/layouts/MainLayout.tsx:123](src/layouts/MainLayout.tsx:123)), although the console also manages members, invitations, and join requests. Consider exposing it independently.

## ✅ Highlights

- Existing IM session-token entry remains supported.
- Space membership is checked before console access.
- Both supported locales include the new user-facing copy.

yujiawei (changes requested):
# Code Review — PR #143 (octo-admin)

Reviewed at head `ee6b01a`. I checked out the branch, read the diff against `main`, traced the auth paths through `src/api/index.ts` and `src/store/auth.ts`, and cross-checked the server side (`modules/user/api.go`, `pkg/httperr`, `pkg/errcode`) to confirm what `/v1/user/login` actually returns on the wire.

Verified the PR body's claims: `tsc --noEmit` is clean and `npm test` passes 200 tests. No server change is needed — `POST /v1/user/login` takes `{username, password, flag}` and `flag: 1` is the Web device flag, which is the right value.

The problem this solves is real. The space console genuinely had no way in, and the fix is the right shape. One blocking issue, all in the same root cause.

---

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the four changes the description lists — space-console sign-in, super-admin identity switch instead of a silent bounce, the platform-scope notice on the App Bot list, and the sidebar entry. Nothing extra is smuggled in: no new server calls beyond `/v1/user/login`, no new flags, no unrelated refactors. The three `entry.noToken.*` keys are removed in both locales as stated.

One wording note, not a defect: the description says "the two identities remain separate." At the account level that is true, but `src/store/auth.ts:41-116` has a single `token` slot, so signing in here *replaces* the super-admin session rather than running alongside it. The user-facing copy (`entry.login.superNotice`) says "switches you to a space administrator," which is accurate — it is only the PR description that reads as if both sessions survive. That single-slot design is what causes the P1 below.

---

## 2. Code quality

### P1 — A failed space sign-in destroys the caller's existing session

`src/pages/SpaceAdmin/SpaceEntry.tsx:143-157`

`handleLogin` calls `useAuthStore.getState().logout()` in its catch, and `enterWithToken` writes the new token into the store at line 119 *before* authorization is confirmed, then calls `logout()` at line 123 on the no-space path. There is no snapshot of what was there before, so every failure path clears the store.

```ts
// SpaceEntry.tsx:118-125
async function enterWithToken(token: string, knownUid: string) {
  useAuthStore.setState({ scope: 'space', token, isLoggedIn: true, name: '', uid: '' })
  const list = await getMySpaces()
  const managed: MySpace[] = (list || []).filter((s) => s.role >= 1)
  if (managed.length === 0) {
    useAuthStore.getState().logout()
    setStatus('no-space')
```

```ts
// SpaceEntry.tsx:150-154
} catch (error) {
  useAuthStore.getState().logout()
  setFormError(error instanceof Error ? error.message : String(error))
  setStatus('form')
```

This matters specifically because of the new `cameFromSuper` path added at lines 63-68. A signed-in super admin now lands on this form by design — from the new sidebar entry, no less. Concrete failure:

1. Super admin clicks **Space console** in the sidebar. `cameFromSuper` is set, the form renders, the super token is still in the store.
2. They mistype the password. `/v1/user/login` rejects.
3. The catch fires `logout()`. `zustand/persist` writes the cleared state to `localStorage['dm-admin-auth']`.
4. The form redisplays with the error — and the super-admin session is gone. Clicking "Sign in as super administrator instead" now means re-entering super credentials.

A typo costs the user their unrelated session. Three variants of the same bug: wrong password, a response with no `token` (line 148 throws), and a successful login by an account that administers no space (line 123). The last one is the worst, because the credentials were correct.

Suggested shape — snapshot and restore rather than `logout()`:

```ts
async function handleLogin(values: { username: string; password: string }) {
  const prev = useAuthStore.getState()
  const restore = () => useAuthStore.setState({
    scope: prev.scope, token: prev.token, name: prev.name,
    role: prev.role, uid: prev.uid, isLoggedIn: prev.isLoggedIn,
  })
  setSubmitting(true)
  setFormError('')
  try {
    const res = await userLogin(values)
    if (!res.token) throw new Error(t('entry.login.noToken'))
    await enterWithToken(res.token, res.uid || '')
  } catch (error) {
    restore()
    setFormError(error instanceof Error ? error.message : String(error))
    setStatus('form')
  } finally {
    setSubmitting(false)
  }
}
```

The `managed.length === 0` branch in `enterWithToken` needs the same treatment — it should hand the failure back to the caller rather than calling `logout()` itself, since the correct rollback differs depending on whether the caller was the session-token path or the form.

### P2 — The space console entry is gated on the App Bots feature probe

`src/layouts/MainLayout.tsx:120-127`

```tsx
if (appBotsAvailable === true) {
  list.push({ key: '/app-bots', ... })
}
if (appBotsAvailable === true) {
  list.push({ key: '/space', icon: <AppstoreOutlined />, label: t('nav:spaceConsole'), group: 'management' })
}
```

The space console manages members, invites, and join requests, none of which depend on app bots. Gating on `appBotsAvailable` means a deployment with app bots off — or one where the `/v1/app_bot/available` probe in `src/store/feature.ts` transiently fails — loses the only discoverable link to the space console. Given that reachability is the entire point of this PR, that is worth getting right. The duplicated condition reads like the block above it was copied.

`space.read` on `managerCapabilities` (already used for `/spaces` at line 118) is the closer match, or no gate at all.

Minor, same file: `/spaces` and `/space` now both render `AppstoreOutlined` with adjacent labels ("Spaces" / "Space console"). Different icons would help. Menu highlighting is fine — `selectedKeys={[location.pathname]}` at line 241 is an exact match, so `/spaces` does not shadow `/space`.

### P2 — Reloading `/admin/space` as a signed-in space admin asks for the password again

`src/pages/SpaceAdmin/SpaceEntry.tsx:70-95`, `src/store/auth.ts:118-127`

`partialize` persists `scope`, `token`, and `isLoggedIn` but deliberately not `mySpaces`. So after any page reload the store rehydrates with a valid space token and an empty `mySpaces`. The effect then:

- skips line 70 (`state.mySpaces.length > 0` is false),
- reaches line 93, finds no IM `sessionStorage` token in an admin-only tab,
- and renders the sign-in form.

The valid `state.token` sitting in the store is never tried. Before this PR the same path produced the "no login session" dead end, so it is not a regression — but this PR adds `enterWithToken`, which is exactly the missing piece, and promotes `/space` to a first-class entry point via the sidebar. Reaching for the persisted token before falling through to the form is a two-line change:

```ts
const token = readSessionToken() || (state.scope === 'space' ? state.token : '')
```

### P2 — The tests assert on source text, not behaviour

`src/pages/SpaceAdmin/SpaceEntry.test.tsx:10-11, 19-32, 73-74`

The file imports the components as raw strings and greps them:

```ts
import entrySource from './SpaceEntry.tsx?raw'
expect(entrySource).toContain('await userLogin(values)')
expect(entrySource).toMatch(/scope === 'super'[\s\S]*?setCameFromSuper\(true\)[\s\S]*?setStatus\('form'\)/)
```

Nothing is rendered, no request is mocked, no store transition is observed. These assertions pass whether or not the component works, and they break on any rename or reformat even when behaviour is unchanged. None of them can catch the P1 above — that is the practical cost, not a style objection.

The locale-key assertions (lines 34-58) and the nav-key assertions (lines 62-65) are genuinely useful and worth keeping; it is the six source-grep assertions that should become real tests. The repo already renders components under jsdom in `src/pages/Changelog/DesktopDownloadBar.test.tsx` and `src/pages/Dashboard/index.test.tsx`, so the pattern is established. A test that renders the form, mocks `userLogin` to reject, and asserts the super-admin token survives would pin the P1 fix.

The description says the file "covers the new behaviour." It covers the new *source*.

### P2 — Deployments with local login disabled get a form that cannot succeed

`src/pages/SpaceAdmin/SpaceEntry.tsx:93-97`

When the server has local login turned off, `/v1/user/login` refuses every request before looking at credentials. On an SSO-only deployment the new form is therefore a dead end — and the copy this PR removed (`entry.noToken.subtitle`, "please sign in within the app before entering space admin") was the correct instruction for exactly that case.

The IM-tab `sessionStorage` path still works there, and the "Sign in as super administrator instead" link is still present, so this degrades rather than breaks. Still worth either probing the setting or keeping a hint in the error state.

### Nit — the platform-scope link forces a full page reload

`src/pages/AppBots/index.tsx:264`

```tsx
<Typography.Link href="/admin/space">{t('list.platformOnly.link')}</Typography.Link>
```

A raw `href` to a hardcoded absolute path re-downloads the bundle and re-runs the store rehydration. The router basename is `/admin` (`src/App.tsx:112-115`), so a react-router `Link to="/space"` gives the same destination as a client-side navigation and stops the base path from being hardcoded in a second place.

### Checked and clear

- `flag: 1` is correct — the Web device flag, matching `config.DeviceFlag` ordering (0=APP, 1=Web, 2=PC).
- The token the request interceptor attaches to the `/v1/user/login` call (`src/api/index.ts:26-29` — during `cameFromSuper` this is the super-admin token) is harmless: the route is registered on an unauthenticated group, so the server never reads it. Worth knowing, not worth fixing here.
- I initially suspected the 401 branch in the response interceptor (`src/api/index.ts:37-40`, `logout()` + `window.location.href`) would fire on a wrong password and hard-redirect away from the form. It does not. `/v1/user/login` returns wire status **400** for invalid credentials — the semantic 401 rides in the `error.http_status` body field, not the transport status. The interceptor branches on `error.response.status`, so the inline error Alert does render. The session loss described in the P1 comes from the component's own `logout()` call, not from the interceptor.
- `enterWithToken` is a hoisted function declaration used above its definition — fine, and the `exhaustive-deps` disable is correctly placed.
- Both locale bundles are complete and symmetric; the removed keys are gone from both.

---

## 3. Verdict

**CHANGES_REQUESTED**

The direction is right and I would like to see this land. The blocker is the single P1: a failed or unauthorized space sign-in must not clear an unrelated session. That is a contained fix in `handleLogin` / `enterWithToken`. The `appBotsAvailable` gate on the sidebar entry is a one-line change I would fold into the same round since it undercuts the PR's own goal.

---

## 4. Suggested order of work

1. Snapshot-and-restore the prior auth state around `handleLogin`, and stop `enterWithToken` from calling `logout()` on the no-space path. **(blocking)**
2. Re-gate the `/space` sidebar entry on `space.read` (or remove the gate).
3. Add one rendering test: mock `userLogin` to reject, assert the pre-existing super-admin token is still in the store afterwards. That is the regression test for item 1.
4. Try the persisted `state.token` before falling through to the form.
5. Swap the `href="/admin/space"` link for a router `Link`.

---

## 5. For a human reviewer to confirm

This touches authentication, so a few things deserve eyes rather than reasoning:

- **A general user token now lives in the admin origin's `localStorage`.** Before this PR the space console could only borrow a token from `sessionStorage`; now it mints a full IM user token and `zustand/persist` writes it to `localStorage['dm-admin-auth']`, where it survives tab close. Confirm that matches the intended session lifetime for the admin console, and that the admin origin's XSS posture is acceptable for holding IM user tokens.
- **Does signing in here with `flag: 1` (Web) disturb the same user's existing IM web session?** Both now issue Web-flag sessions for the same account. Worth one manual check that opening the space console does not sign the admin out of the IM web client.
- **Accounts that require device phone verification.** The server answers that case with a bespoke `{status: 110, msg: "需要验证手机号码！"}` payload. The form surfaces the raw message but has no verification flow, so such an admin cannot get in. Confirm that is acceptable, or route them elsewhere.
- **Browser credential autofill across two sign-in forms on one origin.** `/admin/login` (super) and `/admin/space` (user) now both present username/password with the same `autoComplete` hints. A manager whose saved super credentials get autofilled here will submit them to `/v1/user/login`, which counts failures against the per-account lockout guard for any *user* account sharing that username. Low likelihood, easy to confirm in a browser.
- The rate limiting on this endpoint is server-side (10 req/min, burst 5, per IP) and applies to the new form as-is. No client-side throttle is needed, but confirm that limit is appropriate now that a second surface hits it.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #143 (octo-admin)

## Summary

Gives the space console its own sign-in: `SpaceEntry` now shows a username/password form (backed by the new `userLogin` → `POST /v1/user/login`, the same endpoint IM web uses, with `flag: 1`) instead of the old "no IM session in this tab" dead end, and super admins get an explicit identity switch instead of a silent bounce to the dashboard. Also adds a super-console nav entry for the space console, a platform-only notice on the App-Bots list, and the matching i18n keys. The direction is right and the backend contract is real, but the new form's failure path destroys active sessions it should not touch — that is the blocker below.

## Verification

Static analysis only at head `ee6b01a`; build and tests not executed in this environment.

- ✅ **Backend contract exists** — cross-checked read-only against octo-server default branch: `POST /v1/user/login` is registered on the public `/v1` group (`modules/user/api.go:380`), accepts `{username, password, flag}` (`loginReq`, `:4409`), and returns `token/uid/name/username` (`loginUserDetailResp`, `:4447-4456`) — matching `UserLoginResponse`. `GET /v1/space/my` (`modules/space/api.go:82`) and `GET /v1/users/:uid` (`modules/user/api.go:270`) sit on the session-auth tree that reads the `token` header, exactly what the shared axios client sends.
- ✅ **`flag: 1` (PC) does not evict existing sessions** — for the PC device flag `execLogin` reuses the existing token instead of revoking it (`modules/user/api.go:1872-1876`, PC 支持多登), so signing into the space console shares the user's IM web session token rather than killing it. Device-lock verification applies only to the APP flag.
- ✅ **Token flows from login to the space APIs** — `enterWithToken` writes the token into the auth store *before* `getMySpaces()`; the request interceptor reads `useAuthStore.getState().token` (`src/api/index.ts:26-29`), so the fresh token actually reaches `/v1/space/my`.
- ✅ **i18n parity** — every new key exists in both locales; the removed `entry.noToken.*` keys have no remaining consumers.
- ✅ **First review round for this PR** — no prior blockers to re-check.

## Findings

One P1 blocker below; three P2 items.

### P1 — Failed space-console login destroys the active session (`src/pages/SpaceAdmin/SpaceEntry.tsx:151`)

The new form's failure path wipes whatever session the browser holds, although a failed login must not touch it. Two mechanisms compound: (1) wrong credentials come back as HTTP 401 (server `pkg/errcode/user.go:56`), which hits the unchanged global interceptor at `src/api/index.ts:37-39` — it calls `logout()` and hard-redirects to `/admin/login` before the form's error Alert can ever render; (2) every other failure (timeout, 5xx, missing token, or a successful login with zero manageable spaces via `:123`) is caught by `handleLogin`, whose catch at `:151` calls `logout()` as well.

Concrete regression: a super admin arrives via this PR's own new nav item (`src/layouts/MainLayout.tsx:123`) holding a valid persisted super session (localStorage `dm-admin-auth`); one mistyped password wipes the super session entirely and dumps the user on the super login page. Before this PR, super admins at `/space` were redirected to `/dashboard` with their session intact (verified against `pr-base`). A mistyped password is a common, production-reachable trigger — this makes a previously working path (super session persistence) unavailable, and the user never sees an inline error because the page is already navigating away.

```ts
// current: catch unconditionally wipes the store
} catch (error) {
  useAuthStore.getState().logout()
  ...
}
```

Fix direction: snapshot the auth state before entering the form and restore it on failure instead of `logout()` (same for the no-space branch at `:123`), and exempt login requests from the 401 interceptor's logout+redirect so credential failures render inline in `formError`.

### P2 — Entry page ignores a valid persisted space session (`src/pages/SpaceAdmin/SpaceEntry.tsx:93`)

`mySpaces` is deliberately not persisted (`src/store/auth.ts` partialize), so after a fresh tab or full reload the second effect branch (`mySpaces.length > 0`) can never match; the page falls through to `readSessionToken()`, and sessionStorage is per-tab. A user who already signed in through the new form — token persisted in the store — is shown the login form again. Deep pages under `/space/:spaceId` recover from the persisted token via `SpaceAdminLayout`, so behavior is inconsistent, and the PR's stated goal ("换标签页就进不去") is only half met: you can re-login in a new tab, but the existing session is not honored. Fix: before showing the form, if `isLoggedIn && scope === 'space' && token`, call `enterWithToken(state.token, state.uid)`.

### P2 — Space-console nav entry gated on the app-bots probe (`src/layouts/MainLayout.tsx:123`)

The new `/space` menu item is gated on `appBotsAvailable === true`, i.e. on the `GET /v1/app_bot/available` probe (`src/store/feature.ts`: 404 → cached false for 10 min; 5xx/network → false for 30 s). The space console's core (members/invites/join-applies) does not depend on the app-bots feature, so on deployments without app-bots the entry this PR delivers disappears stably for 10 minutes at a time. Show the space-console entry unconditionally (or on its own signal) and keep the probe gate on `/app-bots` only.

### P2 — New tests assert source text, not runtime behavior (`src/pages/SpaceAdmin/SpaceEntry.test.tsx:19`)

Every assertion string-matches the `?raw`-imported source; nothing executes the login orchestration (credential-failure handling, session preservation, token propagation to `/v1/space/my`). That is exactly the territory the P1 lives in — these tests cannot catch it. Alongside the P1 fix, add one behavioral test (mock at the api/axios level): an active super session survives a wrong password, `formError` renders, no redirect happens.

## Human-verify

1. With the PC flag the space console receives the *same* token the user's IM web session already holds (server reuses it — verified in `execLogin`). Please confirm operationally that token sharing between IM web and the space console is acceptable in your deployments. **Not a merge blocker for this PR.**
2. If a deployment sets `LocalLoginOff` (server system setting), `POST /v1/user/login` is rejected globally and the space-console form can never succeed. Confirm local login is enabled wherever the space console is offered. **Not a merge blocker for this PR.**

- Paired paths (login↔logout): logout semantics intact; the problem is failure-path scope, covered by the P1. The successful super→space switch correctly overwrites the store and is warned by `entry.login.superNotice`.
- Reused control flow: `enterWithToken` serves both entry paths with identical ordering (store write → `getMySpaces`); no double effect, `onceRef` guards StrictMode double-run.
- Authorization boundary: space roles are enforced server-side (`/v1/space/my` + per-space checks); the client-side `role >= 1` filter is display-only. The login route is public by design, and the stale `token` header on the login POST is ignored server-side.
- Runtime vs build: backend contract verified against server source rather than trusted from build/tests; the new tests' limitation is called out above.
- `href="/admin/space"` in the App-Bots notice matches the production basename (`vite.config.ts` base `/admin/`, `detectBasename` → `/admin`).

## Verdict: CHANGES_REQUESTED

The sign-in flow is the right shape and the backend contract is real, but a wrong password must not destroy an unrelated active session. Fix the P1 (restore-instead-of-logout plus a login exemption from the 401 interceptor's side effects) and this is in good shape to land; the three P2s are recommended, not blocking.

[Octo-Q] verdict: REQUEST_CHANGES — P1: 空间控制台登录失败（输错密码/任意失败）会经 401 拦截器 + catch 的双重 logout() 清掉无关的活跃会话（超管会话被误删、用户被甩到超管登录页、行内错误无法渲染），相对 pr-base 是新引入回归。

yujiawei:
Follow-up to my review — I dug into the server side on two points, and one of them resolves an open question I'd left for manual checking.

**`flag: 1` is Web, not PC — but the no-eviction conclusion holds anyway.**

`octo-lib/config/msg.go:31-38` defines `APP DeviceFlag = iota`, then `Web`, then `PC`, so `flag: 1` is Web. The reason it doesn't evict anything is that `execLogin` branches on `flag == config.APP` only (`modules/user/api.go:1852`); Web and PC share the `else`:

```go
} else { // PC暂时不执行删除操作，因为PC可以同时登陆
    if strings.TrimSpace(oldToken) != "" { // 如果是web或pc类设备 因为支持多登所以这里依然使用老token
        token = oldToken
        reuseExistingToken = true
    }
}
```

So my earlier "please confirm signing in here doesn't sign the admin out of the IM web client" is answered: it doesn't. You can drop that one.

**It does mean something sharper, though.** Because `oldToken` is *reused* rather than reissued, an admin who already has an IM web session gets back **the exact same token string** that session is using. This PR then persists it to the admin origin's `localStorage` under `dm-admin-auth`, where it outlives the tab. Two consequences worth a deliberate decision rather than an accident:

- Revoking that token anywhere (IM logout, session revocation, password reset) silently signs the user out of the space console too, and vice versa. They are not independent sessions; they are one session with two homes.
- The blast radius of an XSS on the admin origin is now a live IM user session token, not just admin-console state.

Neither blocks this PR — the token was already reachable via the `sessionStorage` path this replaces. But that path was tab-scoped and read-only; this one mints and persists. Worth confirming that matches the intended session lifetime.

**Separately — one thing not to fix.** If anyone suggests exempting the login call from the 401 handler in `src/api/index.ts:37-40`: that isn't needed, because the interceptor never fires here. `/v1/user/login` returns wire status **400** for bad credentials, not 401. `respondUserError` routes through `httperr.ResponseErrorL` → `respondL(..., useSemanticStatus=false)` (`pkg/httperr/respond.go:58-70`), which pins the transport status at 400 during the D14 compatibility window; the semantic 401 from `pkg/errcode/user.go:56` appears only in the `error.http_status` body field. There's a regression test pinning exactly this divergence (`pkg/i18n/renderer_test.go:86`). The interceptor branches on `error.response.status`, so it sees 400 and does nothing — your inline `formError` Alert renders as intended.

It's an easy inference to get wrong, since the errcode registry does say `http.StatusUnauthorized`. The session loss in my P1 comes entirely from the component's own `logout()` calls at `SpaceEntry.tsx:123` and `:151`. Fixing those two is sufficient; no interceptor change is required.

PR #139, opened 2026-08-20
feat(auth): support email MFA login flow

lhk0504:
## Related PR

- Depends on [Mininglamp-OSS/octo-server#813](https://github.com/Mininglamp-OSS/octo-server/pull/813)
- This PR should be merged after `octo-server#813` has been merged and deployed.

## Merge order

Please merge the changes in the following order:

1. Merge and deploy `octo-server#813` first.
2. Merge this `octo-admin#139` PR afterwards.

Server PR #813 introduces the dedicated manager MFA verification-lockout error code:

```text
err.server.user.manager_mfa_verify_locked
```

This PR handles that server contract. The frontend also keeps a compatibility branch for the legacy `err.server.user.manager_mfa_rate_limited` response, but the canonical client-server contract is the one introduced by `octo-server#813`. Therefore, this PR should not be merged or released before the server PR is available.

## Summary

Improve the management console email MFA integration and handle verification lockouts after repeated incorrect codes.

This change is based on the latest `main` branch. It preserves the existing session-restore flow and integrates the MFA state machine into the credentials form.

### 1. Complete the manager MFA login flow

Integrate the following endpoints:

- `/v1/manager/login`
- `/v1/manager/login/send`
- `/v1/manager/login/resend`
- `/v1/manager/login/verify`

The login flow is now:

1. The administrator submits their username and password.
2. The server creates an MFA challenge.
3. The administrator requests an email verification code.
4. The administrator enters the code and verifies the challenge.
5. A management-console session is established after successful verification.

The client also tolerates the actual response shape returned by the server. `mfa_required`, `code_sent`, and `resend_after` are treated as optional fields so that responses from server versions that omit them remain valid.

### 2. Handle code sending, resending, and cooldowns

- Support the initial verification-code request.
- Support resending a code for the current challenge.
- Read `retry_after` from server error details when available.
- Apply the client-side minimum send cooldown.
- Show a countdown when sending is rate-limited.
- Return to the credential form when the challenge is no longer valid.
- Preserve the challenge sending state during the MFA flow.

### 3. Handle verification lockouts

After repeated incorrect verification codes, the server locks verification for the email address.

This PR uses the dedicated error code introduced by `octo-server#813` as the canonical contract:

```text
err.server.user.manager_mfa_verify_locked
```

It also supports the legacy error code used by older server versions:

```text
err.server.user.manager_mfa_rate_limited
```

When a verification lockout is returned:

- A matching HTTP `429` response is recognized as a verification lockout.
[...]

PR #144, opened 2026-08-21
feat(app-bots): give the platform list an ownership dimension

StefanoWoo:
> Stacked on #143 — it modifies the platform-only notice that PR introduces. The diff shown here therefore contains that commit too; review or merge #143 first and this becomes a single commit.

## The problem

The platform App Bot list is scope-filtered server-side, so a bot that moves into a space disappears from it. Nothing on the page says the list is scoped, and there is no way to look wider, so the bot appears to have been deleted. On our deployment the bot was running the whole time.

This is the second half of #134. The first half — being able to reach the space console at all — is #143.

## What this changes

- **An ownership filter** (Platform / Spaces / All owners) next to the status filter, defaulting to Platform so the page opens on exactly what it showed before.
- **An ownership column** naming the owning space, falling back to the space id when the name is unavailable. Platform console only: inside a space console every row has the same owner, so the column would be noise.
- The platform-only notice from #143 yields once the filter widens, since at that point it would be describing something that is no longer true.

## Server support and degradation

Uses the `scope` parameter proposed in [Mininglamp-OSS/octo-server#799](https://github.com/Mininglamp-OSS/octo-server/pull/799), where rows also gain `space_id` / `space_name`.

Against a server without that change the parameter is ignored and every option returns platform bots — which is what the default option shows — so an older server degrades to the previous behaviour instead of erroring. The column renders `Platform` for those rows.

## Still open from #134

Scope selection in the create dialog. I left it out on purpose: creating a bot directly into a space means a platform admin writing into a space they may not administer, which crosses the boundary `botInRouteScope` draws on the server. With #143 the space console is reachable and has its own create dialog, so a space bot can be created where the authority for it lives. If you would rather have the platform dialog able to create into a space, say how you want the authority checked and I will send that too.

### DIFF
--- a/src/api/app-bot.ts
+++ b/src/api/app-bot.ts
@@ -24,6 +24,8 @@ export interface AppBot {
   welcome_msg: string
   scope: AppBotScope
   space_id: string | null
+  /** Only present on a mixed (scope=all / scope=space) platform listing. */
+  space_name?: string
   status: AppBotStatus
   token?: string
   connect?: BotConnectInfo
@@ -68,13 +70,25 @@ export interface ListParams {
   status?: number
 }
 
+/**
+ * Ownership filter for the platform listing. A server that predates the scope parameter
+ * ignores it and answers with platform bots only, which is exactly what the 'platform'
+ * option shows — so an older server degrades to the previous behaviour rather than erroring.
+ */
+export type AppBotScopeFilter = 'platform' | 'space' | 'all'
+
+export interface PlatformListParams extends ListParams {
+  scope?: AppBotScopeFilter
+  space_id?: string
+}
+
 // --- Helpers ---
 
 const unwrap = <T>(p: Promise<{ data: T }>): Promise<T> => p.then((r) => r.data)
 
 // --- Platform App Bot (Super Admin) ---
 
-export const listAppBots = (params: ListParams = {}): Promise<AppBotListResp> =>
+export const listAppBots = (params: PlatformListParams = {}): Promise<AppBotListResp> =>
   unwrap(api.get<AppBotListResp>('/v1/admin/app_bot', { params }))
 
 export const getAppBot = (id: string): Promise<AppBot> =>
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -15,6 +15,21 @@ interface LoginResponse {
 export const login = (params: LoginParams) =>
   api.post<LoginResponse>('/v1/manager/login', params).then((res) => res.data)
 
+// 空间控制台自有登录（普通用户身份）。此前空间控制台没有登录入口，只能寄生于同一
+// 标签页 sessionStorage 里恰好存在的 IM 会话 token —— 换标签页就进不去，且与超管会话互斥。
+// 服务端 /v1/user/login 是 IM Web 用的同一个端点，空间管理员用自己的账号登录即可。
+export interface UserLoginResponse {
+  token: string
+  uid: string
+  name: string
+  username?: string
+}
+
+export const userLogin = (params: LoginParams) =>
+  api
+    .post<UserLoginResponse>('/v1/user/login', { ...params, flag: 1 })
+    .then((res) => res.data)
+
 export const getManagerMe = () =>
   api.get<ManagerMe>('/v1/manager/me').then((res) => ({
     ...res.data,
--- a/src/i18n/locales/en-US/appBots.json
+++ b/src/i18n/locales/en-US/appBots.json
@@ -96,5 +96,11 @@
   "edit.toast.updated": "Updated",
   "list.platformOnly.title": "This list shows platform-level app bots only",
   "list.platformOnly.desc": "Space-level app bots, including any moved from the platform into a space, are managed in that space’s own console.",
-  "list.platformOnly.link": "Open the space console"
+  "list.platformOnly.link": "Open the space console",
+  "column.ownership": "Ownership",
+  "list.scopeFilter.platform": "Platform",
+  "list.scopeFilter.space": "Spaces",
+  "list.scopeFilter.all": "All owners",
+  "scope.platform": "Platform",
+  "scope.space": "Space"
 }
--- a/src/i18n/locales/en-US/appBots.json
+++ b/src/i18n/locales/en-US/appBots.json
@@ -93,5 +93,8 @@
   "form.welcome.extra": "Message automatically sent on a user's first connection; leave empty to use the default prompt",
   "form.welcome.placeholder": "Hi! I'm the xx assistant. How can I help you?",
   "edit.title": "Edit {{name}}",
-  "edit.toast.updated": "Updated"
+  "edit.toast.updated": "Updated",
+  "list.platformOnly.title": "This list shows platform-level app bots only",
+  "list.platformOnly.desc": "Space-level app bots, including any moved from the platform into a space, are managed in that space’s own console.",
+  "list.platformOnly.link": "Open the space console"
 }
--- a/src/i18n/locales/en-US/nav.json
+++ b/src/i18n/locales/en-US/nav.json
@@ -13,5 +13,6 @@
   "expertMarket": "Expert Market",
   "backup": "Backups",
   "download": "Downloads",
-  "systemSkill": "Skill Market"
+  "systemSkill": "Skill Market",
+  "spaceConsole": "Space console"
 }
--- a/src/i18n/locales/en-US/spaceAdmin.json
+++ b/src/i18n/locales/en-US/spaceAdmin.json
@@ -38,12 +38,19 @@
   "validate.logoTooLong": "Logo cannot exceed {{max}} characters",
   "validate.descriptionTooLong": "Space description cannot exceed {{max}} characters",
   "entry.loading": "Entering space admin…",
-  "entry.noToken.title": "No login session detected",
-  "entry.noToken.subtitle": "Please sign in within the app before entering space admin, or sign in with a super admin account.",
-  "entry.noToken.action": "Go to sign in",
   "entry.noSpace.title": "No manageable spaces",
   "entry.noSpace.subtitle": "You are not an admin or owner of any space yet.",
   "entry.noSpace.action": "Back to sign in",
   "entry.error.title": "Failed to load space",
-  "entry.error.action": "Retry"
+  "entry.error.action": "Retry",
+  "entry.login.title": "Space console",
+  "entry.login.subtitle": "Sign in with your account to manage the spaces you administer.",
+  "entry.login.superNotice": "You are signed in as a super administrator. The space console is a separate identity; signing in here switches you to a space administrator.",
+  "entry.login.username": "Username",
+  "entry.login.password": "Password",
+  "entry.login.usernameRequired": "Enter your username",
+  "entry.login.passwordRequired": "Enter your password",
+  "entry.login.submit": "Sign in",
+  "entry.login.toSuper": "Sign in as super administrator instead",
+  "entry.login.noToken": "Login returned no token"
 }
--- a/src/i18n/locales/zh-CN/appBots.json
+++ b/src/i18n/locales/zh-CN/appBots.json
@@ -96,5 +96,11 @@
   "edit.toast.updated": "已更新",
   "list.platformOnly.title": "这里只列平台级应用 Bot",
   "list.platformOnly.desc": "空间级应用 Bot（含从平台移入空间的）在各自空间的控制台管理。",
-  "list.platformOnly.link": "前往空间控制台"
+  "list.platformOnly.link": "前往空间控制台",
+  "column.ownership": "归属",
+  "list.scopeFilter.platform": "平台级",
+  "list.scopeFilter.space": "空间级",
+  "list.scopeFilter.all": "全部归属",
+  "scope.platform": "平台级",
+  "scope.space": "空间级"
 }
--- a/src/i18n/locales/zh-CN/appBots.json
+++ b/src/i18n/locales/zh-CN/appBots.json
@@ -93,5 +93,8 @@
   "form.welcome.extra": "用户首次连接时自动发送的消息，留空则使用默认提示",
   "form.welcome.placeholder": "你好！我是 xx 助手，有什么可以帮你的？",
   "edit.title": "编辑 {{name}}",
-  "edit.toast.updated": "已更新"
+  "edit.toast.updated": "已更新",
+  "list.platformOnly.title": "这里只列平台级应用 Bot",
+  "list.platformOnly.desc": "空间级应用 Bot（含从平台移入空间的）在各自空间的控制台管理。",
+  "list.platformOnly.link": "前往空间控制台"
 }
--- a/src/i18n/locales/zh-CN/nav.json
+++ b/src/i18n/locales/zh-CN/nav.json
@@ -13,5 +13,6 @@
   "expertMarket": "专家市场",
   "backup": "备份管理",
   "download": "下载配置",
-  "systemSkill": "Skill 市场"
+  "systemSkill": "Skill 市场",
+  "spaceConsole": "空间控制台"
 }
--- a/src/i18n/locales/zh-CN/spaceAdmin.json
+++ b/src/i18n/locales/zh-CN/spaceAdmin.json
@@ -38,12 +38,19 @@
   "validate.logoTooLong": "Logo 不能超过 {{max}} 个字符",
   "validate.descriptionTooLong": "空间描述不能超过 {{max}} 个字符",
   "entry.loading": "正在进入空间管理…",
-  "entry.noToken.title": "未检测到登录态",
-  "entry.noToken.subtitle": "请先在应用内登录后再进入空间管理，或使用超级管理员账号登录。",
-  "entry.noToken.action": "去登录",
   "entry.noSpace.title": "暂无可管理的空间",
   "entry.noSpace.subtitle": "你还不是任何空间的管理员或拥有者。",
   "entry.noSpace.action": "返回登录",
   "entry.error.title": "加载空间失败",
-  "entry.error.action": "重试"
+  "entry.error.action": "重试",
+  "entry.login.title": "空间控制台",
+  "entry.login.subtitle": "用你的账号登录，管理你是管理员的空间。",
+  "entry.login.superNotice": "你当前是超级管理员身份。空间控制台是另一套身份，登录后将切换为空间管理员。",
+  "entry.login.username": "用户名",
+  "entry.login.password": "密码",
+  "entry.login.usernameRequired": "请输入用户名",
+  "entry.login.passwordRequired": "请输入密码",
+  "entry.login.submit": "登录",
+  "entry.login.toSuper": "改用超级管理员登录",
+  "entry.login.noToken": "登录未返回 token"
 }
--- a/src/layouts/MainLayout.tsx
+++ b/src/layouts/MainLayout.tsx
@@ -120,6 +120,11 @@ const MainLayout: React.FC = () => {
     if (appBotsAvailable === true) {
       list.push({ key: '/app-bots', icon: <RobotOutlined />, label: t('nav:appBots'), group: 'management' })
     }
+    if (appBotsAvailable === true) {
+      // 空间级 Bot / 成员 / 邀请在各空间自己的控制台管理，超管这边只能看到平台级。
+      // 没有这个入口时，空间控制台只能靠手敲 URL 进入。
+      list.push({ key: '/space', icon: <AppstoreOutlined />, label: t('nav:spaceConsole'), group: 'management' })
+    }
     if (hasManagerCapability(managerCapabilities, 'system_setting')) {
       list.push({ key: '/system-setting', icon: <SettingOutlined />, label: t('nav:systemSetting'), group: 'system' })
     }
--- a/src/pages/AppBots/index.tsx
+++ b/src/pages/AppBots/index.tsx
@@ -10,6 +10,7 @@ import {
   message,
   Typography,
   Avatar,
+  Alert,
 } from 'antd'
 import { PlusOutlined, SearchOutlined, RobotOutlined } from '@ant-design/icons'
 import type { ColumnsType } from 'antd/es/table'
@@ -247,6 +248,24 @@ export default function AppBotsPage({ spaceId }: Props) {
           </Button>
         </Space>
       </div>
+
+      {!spaceId && (
+        // 这张表只列平台级 Bot（服务端 /v1/admin/app_bot 按 scope 过滤，见 botInRouteScope
+        // 的跨租户防护）。空间级 Bot 在各自空间的控制台管理。没有这句提示时，
+        // 一个被移到空间的 Bot 会从这页凭空消失，用户无从知道它去哪了。
+        <Alert
+          type="info"
+          showIcon
+          style={{ marginBottom: 16 }}
+          message={t('list.platformOnly.title')}
+          description={
+            <span>
+              {t('list.platformOnly.desc')}{' '}
+              <Typography.Link href="/admin/space">{t('list.platformOnly.link')}</Typography.Link>
+            </span>
+          }
+        />
+      )}
 
       <Table
         rowKey="id"
--- a/src/pages/AppBots/index.tsx
+++ b/src/pages/AppBots/index.tsx
@@ -19,6 +19,7 @@ import type { TFunction } from 'i18next'
 import {
   listAppBots,
   listSpaceAppBots,
+  type AppBotScopeFilter,
   deleteAppBot,
   deleteSpaceAppBot,
   publishAppBot,
@@ -55,6 +56,15 @@ const statusOptions = (t: TFunction) => [
   { value: '2', label: t('status.unpublished') },
 ]
 
+// Ownership filter, platform console only. 'platform' is the default so the page opens on
+// exactly what it has always shown; a server without scope support answers the same way
+// for every option, which degrades to that same platform-only list instead of erroring.
+const scopeOptions = (t: TFunction) => [
+  { value: 'platform', label: t('list.scopeFilter.platform') },
+  { value: 'space', label: t('list.scopeFilter.space') },
+  { value: 'all', label: t('list.scopeFilter.all') },
+]
+
 interface Props {
   spaceId?: string
 }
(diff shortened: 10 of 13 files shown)


Jerry-Xin (changes requested):
The ownership view is relevant and useful, but space-owned rows incorrectly retain platform-only management actions.

## 🔴 Blocking

- 🔴 **Critical — Space-owned bots are presented as editable from the platform console.** In [src/pages/AppBots/index.tsx:217](src/pages/AppBots/index.tsx:217), every row receives edit, token, publish, and delete actions. These ultimately use platform endpoints because `spaceId` is undefined. The PR itself states that platform administrators must not write into spaces they may not administer, and the existing API separates platform and space mutations. Consequently, space-owned rows returned by `scope=space/all` expose actions that will fail authorization or violate the intended ownership boundary if the server permits them. The clickable bot name at line 166 also opens the platform detail/token endpoint. Render space-owned rows as read-only in this view, or provide an explicitly authorized navigation/management flow.

## 💬 Non-blocking

- 🟡 **Warning — Creating while filtered to “Spaces” produces a platform bot that immediately disappears.** The create button at [src/pages/AppBots/index.tsx:278](src/pages/AppBots/index.tsx:278) remains enabled for every ownership filter, while creation always uses the platform endpoint. Consider labeling it “Create platform bot,” disabling it in the Spaces view, or switching the filter to Platform after creation.

- 🟡 **Warning — The new tests inspect source strings rather than behavior.** [src/pages/AppBots/ownership.test.ts:1](src/pages/AppBots/ownership.test.ts:1) would pass even if the controls or actions were functionally broken. Add component/API tests covering scope requests, stale responses, and read-only space rows.

## ✅ Highlights

- The default preserves the historical platform-only behavior.
- Ownership names gracefully fall back to space IDs.
- The ownership column and platform-only notice are correctly limited by console/filter context.
- Focused tests and the build could not be executed because project dependencies are not installed (`vitest: command not found`).

yujiawei (changes requested):
# Code Review — PR #144 (octo-admin)

Reviewed at head `90c2e51`. I fetched the branch, read `git diff ee6b01a..90c2e51` (this PR's own commit) separately from the inherited `#143` commit, and traced every row-level action in the App Bot list through to the handler it reaches in `octo-server` (`modules/app_bot/app_bot.go`, `pkg/errcode/app_bot.go`, `pkg/i18n/locales/`).

Verified locally: `tsc --noEmit` is clean and `vitest run` passes 205 tests across 19 files.

The problem is real and the diagnosis in the description is right — a bot that moves into a space vanishes from the platform list with nothing to explain it, and an ownership dimension is the correct answer. The read-only widening is also the right security shape, and the server-side IDOR guard it relies on is intact. Two blocking issues, both in the half of the feature the PR is named after.

---

## 1. Spec compliance

**Spec: ✅**

The commit implements exactly the three things its description lists, and no more:

| Claimed | Where | Verdict |
|---|---|---|
| Ownership filter (Platform / Spaces / All owners), defaulting to Platform | `src/pages/AppBots/index.tsx:62-66`, `:80`, `:255-262` | ✅ default is `'platform'`, so the page opens on the historical view |
| Ownership column, platform console only | `src/pages/AppBots/index.tsx:183-199` | ✅ gated on `!spaceId`, falls back `space_name → space_id → "Space"` |
| Platform-only notice yields once the filter widens | `src/pages/AppBots/index.tsx:288` | ✅ `!spaceId && scopeFilter === 'platform'` |
| Uses the `scope` parameter from octo-server#799 | `src/api/app-bot.ts:78-92` | ✅ |

No new endpoints, no new flags, no unrelated refactors, both locales complete and symmetric.

One minor over-build worth naming, though not enough to fail this gate: `PlatformListParams.space_id` (`src/api/app-bot.ts:82`) is declared but never set by any caller — `grep -rn "PlatformListParams" src/` returns only the declaration and `listAppBots`. It mirrors a server capability the console has no UI to reach. Either wire up a space picker for `scope=space` or drop the field.

**On the stacking.** The description discloses that the diff carries `ee6b01a` from #143, so nothing is smuggled in — but that commit is byte-identical to #143's current head, and #143 currently sits at `CHANGES_REQUESTED` from two reviewers over an unfixed session-loss bug in `handleLogin` / `enterWithToken`. Merging #144 into `main` on its own would land that bug. This PR should not merge ahead of, or independently of, #143's fix. See §3.

---

## 2. Code quality

### P1 — Once the filter widens, every action on a space row answers "Bot not found."

`src/pages/AppBots/index.tsx:126-149`, `:166`, `:217-245`; `src/pages/AppBots/EditModal.tsx:58`

The list is now allowed to contain space-scoped rows, but every per-row action still selects its route from the page-level `spaceId` **prop**, which is `undefined` on the platform console. Nothing looks at `record.scope`:

```tsx
// index.tsx:126-134
const handleDelete = async (id: string) => {
  try {
    spaceId ? await deleteSpaceAppBot(spaceId, id) : await deleteAppBot(id)
```

```tsx
// index.tsx:136-143
if (bot.status === 1) {
  spaceId ? await unpublishSpaceAppBot(spaceId, bot.id) : await unpublishAppBot(bot.id)
} else {
  spaceId ? await publishSpaceAppBot(spaceId, bot.id) : await publishAppBot(bot.id)
```

So a space row's **Edit**, **Token**, **Publish/Unpublish** and **Delete** all address `/v1/admin/app_bot/{id}`, and clicking the bot's *name* (`index.tsx:166` → `DetailDrawer` with `spaceId` undefined → `getAppBot`) does too. Edit fails twice — `EditModal.tsx:58` fetches the bot on open, then the submit at `:86` fails again.

The server refuses all of it. `botInRouteScope` returns `false` for a space bot reached through the platform route:

```go
// octo-server modules/app_bot/app_bot.go:178-183
func botInRouteScope(bot *appBotModel, spaceID string) bool {
	if spaceID == "" {
		return bot.Scope == "platform"
	}
	return bot.Scope == "space" && bot.SpaceID == spaceID
}
```

and all seven guarded handlers — get-detail `:519`, update `:584`, delete `:664`, rotate `:727`, reveal `:820`, publish `:861`, unpublish `:919` — respond identically:

```go
if !botInRouteScope(bot, spaceID) {
	respondAppBotNotFound(c)
	return
}
```

That responder resolves to HTTP 404 / `err.server.app_bot.not_found`:

```go
// octo-server pkg/errcode/app_bot.go:44-49
ErrAppBotNotFound = register(codes.Code{
	ID:             "err.server.app_bot.not_found",
	HTTPStatus:     http.StatusNotFound,
	DefaultMessage: "Bot not found.",
})
```

whose zh-CN string is `机器人不存在。` (`pkg/i18n/locales/active.zh-CN.toml:1015`). The list's error path surfaces it verbatim as a red toast (`index.tsx:116`, `:132`, `:147`).

Concrete sequence:

1. Operator opens App Bots, reads the notice, switches the filter to **All owners**.
2. The bot they were looking for appears, tagged with its space. The notice disappears, because the list is no longer platform-only.
3. They click its name to confirm it is alive → **"Bot not found."**
4. Edit → *Bot not found.* Token → *Bot not found.* Publish → *Bot not found.* Delete → *Bot not found.*

This blocks because of what the PR set out to fix. The description's own framing is *"the bot appears to have been deleted… on our deployment the bot was running the whole time."* The widened view now shows the bot exists, then asserts five different ways — in the server's own words — that it does not. The old behaviour failed by omission, which is merely ambiguous; this fails by confident contradiction, on a page that has just told the operator it is showing them everything.

Note this is a UI defect, not a hole: the server is doing the right thing, and `TestScopeWidening_DoesNotOpenMutationsAcrossScopes` in octo-server#799 pins it. The fix belongs entirely on this side — mirror the server's read-only widening in the UI instead of leaving write affordances that cannot resolve.

Suggested shape — branch the row on the record, not the prop:

```tsx
const foreign = !spaceId && record.scope === 'space'   // visible here, manageable elsewhere
```

- For a foreign row, replace the four action links with one link to where the authority lives — `/space/${record.space_id}/app-bots` is already a route (`src/App.tsx:150-157`).
- Make the name in the first column non-clickable for a foreign row, or point it at that same destination, so the detail drawer is never opened against a 404.

A guard at each call site would also work, but pointing the operator at the space console reads better and matches what the notice at `:288` already tells them.

### P1 — Against a server without `scope` support, widening the filter produces a *stronger* false-deletion signal than today

`src/pages/AppBots/index.tsx:110-112`, `:288`

The description says an older server "degrades to the previous behaviour instead of erroring." The list degrades. The notice does not, and together they invert the PR's purpose.

The currently deployed server never reads the parameter:

```go
// octo-server modules/app_bot/app_bot.go:448-458 (main)
pageIndex, pageSize := c.GetPage()
keyword := c.Query("keyword")
statusStr := c.Query("status")
...
bots, total, err := ab.db.queryBotsByScope("platform", "", pageIndex, pageSize, keyword, statusFilter)
```

Unknown query parameters are ignored, so:

- Old server, filter = **Platform** → platform bots, notice shown. Correct, unchanged.
- Old server, filter = **All owners** → still platform bots — but `scopeFilter !== 'platform'`, so `index.tsx:288` suppresses the notice.

Now walk the PR's own user story on that deployment. An operator is hunting a bot that moved into a space. They select **All owners** — the control that promises to show everything. The bot is not there, and the sentence that used to explain why is gone. They conclude the bot really was deleted. Before this PR they at least had the notice; after it, in the one view they would naturally reach for, they have a confident empty answer.

octo-server#799 is still **OPEN and unmerged**, so this is the day-one state for anyone who deploys the console first, not a hypothetical. Options: gate merge/deploy on #799 landing first, or key the notice off the response (no space-scoped rows came back) rather than off the selected filter, or have the server advertise support and disable the control when it is absent.

I initially scored this P2 on the grounds that the ownership column tags every row `Platform`, so nothing on screen is literally false. On re-reading, that is too generous: the column explains the rows, but nothing explains the *absence*, and absence is the entire failure mode this PR exists to address.

### P2 — Concurrent list requests can land out of order, showing one scope's rows under another's label

`src/pages/AppBots/index.tsx:101-124`

```tsx
const fetchList = useCallback(async () => {
  setLoading(true)
  ...
  const resp = spaceId ? await listSpaceAppBots(spaceId, params) : await listAppBots({ ...params, scope: scopeFilter })
  setData(resp.list || [])
  setTotal(resp.count || 0)
}, [page, debouncedKeyword, statusFilter, scopeFilter, spaceId])

useEffect(() => { fetchList() }, [fetchList])
```

There is no `AbortController` and no generation counter, so if a slower earlier request settles after a faster later one it overwrites both `data` and `total`. Switch from **Platform** to **Spaces** on a loaded connection and the table can end up showing platform rows while the Select reads "Spaces", with pagination computed from the wrong count.

To be fair to the author: this race predates the PR — `git show ee6b01a:src/pages/AppBots/index.tsx` has the same shape for `statusFilter` and `debouncedKeyword`. But a stale *status* filter is visibly wrong at a glance, whereas a stale *ownership* filter is exactly the "which bots am I actually looking at" question this feature exists to answer. Worth a generation ref while the file is open.

### P2 — The tests assert on source text, so none of them can see either P1

`src/pages/AppBots/ownership.test.ts:11-33`

The file imports the component and the API module as raw strings and greps them:

```ts
import listSource from './index.tsx?raw'
expect(listSource).toContain('scope: scopeFilter')
expect(listSource).toMatch(/\}, \[page, debouncedKeyword, statusFilter, scopeFilter, spaceId\]\)/)
expect(listSource).toContain("{!spaceId && scopeFilter === 'platform' && (")
```

Nothing renders, no request is mocked, no row is clicked. These pass whether or not the feature works, and the dependency-array regex at line 21 fails on a reformat that changes nothing. The practical cost is precisely the findings above: a test that rendered the table with one `scope: 'space'` row and asserted which endpoint Delete calls would have caught the first P1 in round one. No amount of source-grepping can.

The locale assertions at lines 33-42 are genuinely useful — keep those. It is the four source-grep cases that should become rendering tests. The repo already renders under jsdom in `src/pages/Dashboard/index.test.tsx` and `src/pages/Changelog/DesktopDownloadBar.test.tsx`, so the pattern exists.

This is the same pattern flagged on #143 — worth fixing once, here, rather than accumulating a third instance.

### P2 — zh-CN terminology drift between the new tag and the existing detail drawer

`src/i18n/locales/zh-CN/appBots.json:45` vs `:104`

```json
"detail.scope.platform": "平台",
"scope.platform": "平台级",
```

The same bot is `平台` in the detail drawer and `平台级` in the new list column. en-US is consistent ("Platform" in both); only zh drifts. The test comment at `ownership.test.ts:38` correctly explains why `detail.scope.space` could not be reused (it interpolates `{{spaceId}}`), but that reasoning does not extend to `platform`, which takes no argument.

### Nits

- `src/pages/AppBots/index.tsx:300` — `<Typography.Link href="/admin/space">` hardcodes the router basename a second time and forces a full page reload. A react-router `Link to="/space"` reaches the same place client-side. Inherited from #143, still worth fixing.
- `src/pages/AppBots/index.tsx:198` — `as ColumnsType<AppBot>[number]` papers over inference on the spread branch. Hoisting the column to a typed `const` avoids the assertion.
- `src/pages/AppBots/index.tsx:278-284` — with the filter on **Spaces**, **Create** still creates a *platform* bot, which then does not appear in the list the operator is looking at. Not worth blocking, but a scope hint on the button would close the loop.
- `src/pages/AppBots/index.tsx:255-262` — the new Select has no visible or accessible label, so a screen reader announces two adjacent unlabelled comboboxes. Matches the existing status filter, so this is a pre-existing pattern rather than a regression.

### Checked and clear

- **The security posture is sound**, and this is the thing I would most want a human to confirm given the `needs-human-review` label. The widening is read-only. Every mutation *and* the per-bot read stay behind `botInRouteScope`; the platform route cannot publish, unpublish, delete, rotate or reveal a space's bot by global id. That IDOR guard is unmodified by this PR and verified above at seven call sites.
- `space_name` is rendered as a React child inside a `<Tag>`, so a hostile space name is escaped — no injection through the new column.
- `setPage(1)` fires on scope change (`index.tsx:258`), so the filter cannot strand the operator on an out-of-range page.
- `scopeFilter` is correctly present in the `fetchList` dependency array (`index.tsx:120`) — no stale-closure refetch bug.
- The column is correctly suppressed inside a space console (`index.tsx:183`), where every row would carry the same value.
- `scope` is sent even for the default `'platform'`; harmless against an old server, and octo-server#799 treats absent and `platform` identically.
- Sibling PR #139 also edits `src/api/auth.ts` and `src/api/index.ts`, but only adds a `details` field to the error envelope and does not touch the 401 branch — no semantic conflict with this branch beyond a textual one in `auth.ts`.

---

## 3. Verdict

**CHANGES_REQUESTED**

The feature is worth having and the shape is right. Three things stand between it and merge:

1. A widened list must not offer write actions that resolve to *"Bot not found."* Contained fix in `src/pages/AppBots/index.tsx`.
2. The widened list must not silently answer "nothing here" on a server that does not yet support `scope`, with the explanatory notice hidden.
3. Merge order. This branch carries #143's commit unchanged, and #143 has an open blocking review. #144 must not reach `main` before that is resolved.

---

## 4. Suggested order of work

1. Branch the row actions and the name link on `record.scope`, not the `spaceId` prop; send foreign rows to `/space/{space_id}/app-bots`. **(blocking)**
2. Decide the behaviour against a server without `scope` support — keep the notice, disable the control, or gate the deploy on octo-server#799. **(blocking)**
3. Add one rendering test: a table containing a `scope: 'space'` row, asserting no platform-route mutation is reachable from it. That is the regression test for step 1.
4. Add a generation guard to `fetchList` so a late response cannot repaint the table under a newer filter.
5. Rebase once #143's blocking fix lands.

---

## 5. Also noticed

- `src/layouts/MainLayout.tsx:120-127` (inherited from #143) gates the **Space console** sidebar entry on `appBotsAvailable === true`. The space console also manages members, invites and join requests, none of which depend on app bots, so a deployment with app bots off loses the only discoverable link to it. Since #144's notice at `index.tsx:300` now points the operator at exactly that console, this PR raises the cost of getting that gate wrong. Already raised on #143; not blocking here.
- Reachability from the platform console into a space console still requires an identity switch — `/space/:spaceId` sits behind `SpaceOnlyRoute` (`src/App.tsx:66-70`), which bounces a super-admin session back to `/space`. Whatever link step 1 adds will land on the sign-in form from #143 rather than on the bot itself. That is acceptable and arguably correct, but the copy should set that expectation rather than let it read as another dead end.

### Not covered by this review

For whoever does the manual pass: I did not exercise the UI in a browser, did not run octo-server#799's branch to confirm the exact `space_name` payload shape, and did not check what the ownership column renders for a bot whose owning space has since been deleted (a dangling `space_id` would fall through to the raw id, which is probably fine but is untested). Migrations, deployment/proxy configuration, and avatar-upload authorization end to end were also out of scope.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #144 (octo-admin)
**Head:** `90c2e51592a03eccdbae4a74de09f0c2189b3f00` · **merge-base:** `0393ca1c` · **base:** main · round 1

## Summary

Stacked on #143 (whose commit `ee6b01a` is included in this diff), this PR gives the platform App Bot list an ownership dimension: a Platform/Spaces/All-owners filter defaulting to the historical platform-only view, an Ownership column naming the owning space, and a platform-only notice that yields once the filter widens. The #143 half replaces the space console's dead-end "borrow an IM sessionStorage token" path with its own sign-in form (`/v1/user/login`) and adds a Space-console entry to the super-admin nav. The listing half is clean with an honest degradation story; the sign-in half carries one P1 in its error handling that can destroy an unrelated live session.

## Verification

Static analysis only at head `90c2e515`; build and tests not executed in this environment.

- ✅ **Listing contract matches the server** — verified read-only against octo-server PR #799 (head `ffd34650`): `scope` absent/`platform` = historical default, `space` (+optional `space_id`), `all` drops the predicate, unknown value → 400; rows carry `space_id` plus best-effort `space_name`. Client values and the documented old-server degradation hold.
- ✅ **`/v1/user/login` contract** — verified against octo-server `modules/user/api.go`: `flag: 1` is the Web device class (APP=0/Web=1/PC=2, octo-lib `config/msg.go`), which is multi-login-safe: an existing Web token is reused and no other device session is revoked; the response carries `token`/`uid`/`name`/`username` exactly as `UserLoginResponse` expects.
- ✅ **Ownership column fallback** — `space_name || space_id || t('scope.space')` matches server behavior (`space_name` is dropped on lookup failure, `space_id` always present on space rows); rows from an old server carry no `scope` and render the Platform tag, matching the documented degradation.
- ✅ **Fetch wiring** — `scope` is only sent on the platform path (`src/pages/AppBots/index.tsx:112`), `setPage(1)` on filter change, `scopeFilter` in the `useCallback` deps; all new i18n keys exist in both locales and the removed `entry.noToken.*` keys are gone from both.

## Findings

One P1 blocker below; two P2 items.

### P1 — Failed space-console login wipes the live super-admin session (`src/pages/SpaceAdmin/SpaceEntry.tsx:151`)

The auth store is a single-slot zustand store persisted to localStorage (`src/store/auth.ts`); `logout()` clears all of it. When a super admin opens the new Space-console nav entry, the mount effect shows the sign-in form while the super session is still in the store (`:63-68`). In `handleLogin`, any failure of `userLogin(values)` — wrong password, server-side loginGuard lockout, a transient network error on the POST, or a deployment with local login disabled — reaches the catch at `:151`, which unconditionally calls `useAuthStore.getState().logout()` and destroys the untouched super session. Pre-PR, a super admin visiting `/space` was bounced to `/dashboard` with the session intact; now one typo (or a network blip, no typo needed) forces full super re-authentication. Fix direction: only clean up the store when it was actually mutated — treat a `userLogin` failure as form-error-only, and reserve `logout()` for failures after `enterWithToken` has set the space token (or snapshot the prior session and restore it on failure).

### P2 — Persisted space session ignored when re-opening `/space` (`src/pages/SpaceAdmin/SpaceEntry.tsx:93`)

Form-login sessions persist `scope`/`token`/`isLoggedIn` in the `dm-admin-auth` store, but `mySpaces` is intentionally not persisted. On a reload or bookmark of `/space` — or opening it in a new tab — the re-entry condition at `:70` (`mySpaces.length > 0`) fails and the effect falls back to `readSessionToken()` (per-tab sessionStorage), never trying the persisted store token; the user is asked to sign in again despite a valid session. Fall back to the persisted token (e.g. `enterWithToken(state.token, state.uid)`) before showing the form.

### P2 — Space-console nav entry gated on the app-bots feature probe (`src/layouts/MainLayout.tsx:123`)

The new entry renders only when `appBotsAvailable === true` (GET `/v1/app_bot/available`). On deployments where app-bots is disabled (404, cached for 10 min), or during the 30s transient-failure window, the only visible way into the space console disappears even though members/invites/join-applies management does not depend on app-bots. Consider showing the entry unconditionally in the super console or probing it independently.

## Human-verify

1. The diff contains #143's commit `ee6b01a` (#143 still open). If #143 lands first this PR becomes a single commit; this review covers both. **Not** a merge blocker for this PR.
2. Server support (octo-server #799) is still open; against an older server all three filter options degrade to the platform-only list as documented. Worth confirming #799 lands with or before this PR so the new options are meaningful. **Not** a merge blocker for this PR.

- Authorization boundary (C3): the widening is read-only; every mutation still routes through `botInRouteScope` server-side, pinned by `TestScopeWidening_DoesNotOpenMutationsAcrossScopes` (publish/unpublish/delete/token/reveal all 404 for a space bot via the platform route).
- Ownership column renders only in the platform console; the platform-only notice hides once `scopeFilter !== 'platform'`.
- `href="/admin/space"` matches the app's hardcoded `/admin` basename convention (same as the 401 interceptor in `src/api/index.ts`).
- `flag: 1` cannot revoke a mobile APP session (revocation only happens for `flag == config.APP`); the Web class reuses an existing token.
- Tests follow the repo's established `?raw` source-assertion pattern; no test locks in the P1 wipe as intended behavior.

## Verdict: CHANGES_REQUESTED

The listing/ownership half is solid and the degradation story holds end to end, but the sign-in error path can destroy a live super-admin session on a plain failed login attempt — strictly worse than the pre-PR bounce. Splitting the `userLogin` failure from the `enterWithToken` cleanup (the P1) is sufficient to clear this; the two P2s are optional.





---

## Octo automated review补充（给 Opus 终审腿）

### 数据流回溯（被消费数据 → 上游来源 → 是否真流到消费点）
1. `scopeFilter`(state) → `fetchList` params → `listAppBots` → `GET /v1/admin/app_bot?scope=` → server switch(#799) → `queryBotsByScope` → rows。✅ 真流到（含 setPage(1) 与 deps）。
2. `record.space_name`/`space_id` ← `toBotListRespWithSpaceNames`（best-effort）/ `toBotListResp` ← `bot.SpaceID`。space 行 space_id 必非空；name 查失败降级为 id。✅ 与前端 `||` 兜底链匹配。
3. `res.token`/`res.uid` ← `loginUserDetailResp{token,uid,name,username}` ← `execLogin`。✅ 字段形状核实（`modules/user/api.go:4446`）。
4. `flag:1` ← `loginReq.Flag` → DeviceFlag Web(=1) → 多登/复用既有 Web token、不吊销其他设备会话。✅（`config/msg.go` iota: APP=0/Web=1/PC=2）。
5. 持久化 store token ← zustand persist partialize（含 token）→ **但 SpaceEntry 重入时不消费它**（只读 sessionStorage）→ 即 P2-1。⚠️
6. `appBotsAvailable` ← `probeAppBots` GET `/v1/app_bot/available` → 门控新导航项 → 即 P2-2。⚠️
7. 超管会话 ← `loginSuper` → 单槽 store → `logout()` 全清 → 即 P1。⚠️

### diff-scope 三问（每个 finding）
- **P1（:151）**：new —— handleLogin 及其 catch 全为本 PR 新增；PR 前 /space 对超管是 navigate('/dashboard')，无任何可失败登录路径。非既有行为。
- **P2-1（:93）**：new —— 表单登录会话为本 PR 新产物；PR 前所有 space 会话都由 per-tab sessionStorage token 支撑，fallback 与会话来源天然一致。
- **P2-2（MainLayout:123）**：new —— 导航项本身新增，非既有行为降级。

### 盲点 checklist（security_sensitive，全项）
- **C1 双路径 parity**：CLEAR —— platform↔space 成对路径逐项核对：`spaceId ? listSpaceAppBots : listAppBots`、ownership 列 `...(spaceId ? [] : [...])`、platformOnly 提示 `!spaceId && scopeFilter==='platform'`，两端条件一致。
- **C2 control-flow ordering / 复用**：CLEAR —— `enterWithToken` 两个 call-site（mount effect / handleLogin）顺序均为先 setState 再 getMySpaces；`onceRef` 防重入；无双重作用。
- **C3 授权边界**：CLEAR（重点核验）—— 列表放宽为只读；mutation 仍全部走 `botInRouteScope`，#799 带回归测试钉死平台路由对 space bot 的 publish/unpublish/delete/token/reveal 全部 404 且不泄露 token。
- **C4 授权生命周期/级联**：N/A —— 本 PR 无鉴权生命周期或容器-成员状态逻辑改动。
- **C5 build 过 ≠ 运行期正确**：HIT 已做 —— 全部结论来自运行期推演（mount 顺序、store 持久化 partialize、sessionStorage per-tab 语义、401 拦截器副作用），未以静态通过缓解；未执行 build/tests（环境限制，已如实声明）。
- **C6 治理文档自洽**：N/A —— 无治理/策略/安全文档改动。

### 跨轮 blocker 复检（R6）
N/A —— 本 PR 首轮审查，无上轮未解决 blocker。

### 建议（具体）
1. P1 修法：把 `handleLogin` 拆两段 —— `userLogin` 失败只 `setFormError`（不动 store）；仅 `enterWithToken` 阶段包 try/catch 走 `logout()`。或进入前快照 store、失败恢复。
2. P2-1：entry effect 增加持久化 token 兜底分支（`isLoggedIn && scope==='space' && token` → `enterWithToken(state.token, state.uid)`），先于表单。
3. P2-2：导航项与 app-bots 探测解耦（超管控制台恒显示，或独立探测）。

### 额外发现（非阻塞）
- 部署若关闭本地登录（`LocalLoginOff`），该表单对此类部署恒失败并回显服务端错误 —— 运维需知晓。
- 登录失败计数受服务端 `loginGuard` 按用户名锁定保护，与既有登录面一致。

[Octo-Q] verdict: REQUEST_CHANGES —— 存在 1 个 P1（登录失败无差别清空在用的超管会话，生产可达且严格劣于 PR 前行为），按 R4 不得 APPROVE；listing 半区与降级路径核验干净，P1 修复即可放行。

PR #152, opened 2026-08-25
fix(login): restore a live session instead of forcing a new sign-in

an9xyz:
## Why

The login page never asked whether this browser was already signed in. There is no route guard on `/login`, the component had no `useEffect`, and the single call site of `getManagerMe()` lives in `MainLayout` — which only mounts once you are already inside the console. So an operator returning to `/admin/login` holding a valid manager token had exactly one thing to do: submit the form again.

That is not free. `/v1/manager/login` mints a fresh UUID on every successful authentication and never reuses the token the request already carries, so each redundant sign-in strands another server-side session. Sessions are capped per uid, and an admin who signs in a few times a day reaches that cap — at which point signing in fails outright, with an error that says nothing about the cause.

Reproduced against the test backend before the change, by stashing this PR and re-running the same browser scenario:

| | before | after |
|---|---|---|
| landing URL | `/admin/login` | `/admin/dashboard` |
| credentials form | **shown** — holding a valid token | not shown |
| `/v1/manager/me` calls | **0** | 1 |
| outcome | only option is to re-submit the form → server mints a new session | token unchanged, no session created |

This is not a regression — the structure dates to the initial commit (44e49b7).

## What it does

On mount the page probes `/v1/manager/me` with the persisted credential. Four outcomes, each treated as a different thing:

- **valid** — hydrate the capability map and navigate to the path those capabilities imply. `/v1/manager/login` is never called, so no session is created. `MainLayout`'s own fetch is guarded on `managerProfileStatus`, so this costs one request, not two.
- **401** — the credential itself was rejected. Drop it, show the form.
- **403** — the credential is fine; this account no longer has console access. Its own copy and a sign-out exit, rather than outage text and an infinite Retry.
- **anything else** — the server did not rule. A timeout, a 502, a 5xx leaves the credential in place and offers a retry. Clearing it would turn one backend hiccup into every admin being signed out, and each re-submitted form is another leaked session.

The probe owns its 401 rather than tripping the global interceptor, which clears credentials and hard-redirects on **every** 401 — the exact conflation this PR exists to undo. A `skipAuthRedirect` request option carries that, defaulting to off, with the login probe named in its doc comment as the only sanctioned caller. `src/api/index.test.ts` pins both halves, so the default path cannot quietly stop revoking credentials.

The probe uses its own 8s timeout rather than the shared client's 30s, and `checking` carries the same "sign in with a password instead" affordance as the failure states — restoring a session only saves one form submission, so it has no business being slower than the thing it replaces. Abandoning cancels for real (generation counter + `AbortController`), so an operator who gives up and starts typing is not yanked away when the probe later succeeds.

### Two status axes, two questions

`ApiError` carries both `status` (envelope-preferred, what existing call sites branch on for 404/409/400) and `transportStatus` (the wire status). They disagree by design: octo-server renders a set of endpoints through `httperr.ResponseErrorL`, which pins the wire status to 400 for legacy clients and puts the real status only in `error.http_status`.

- **"Is the credential still valid?"** → the transport status. octo-lib's `AuthMiddleware` answers an expired token with a genuine wire 401 and a bare `{"msg": …}` body. This must agree with the global interceptor.
- **"What did authorization decide?"** → the envelope. `/v1/manager/me`'s 403 goes through `respondManagerForbidden` → `ResponseErrorL`, arriving as wire 400 with `http_status: 403` — octo-server pins this in `api_manager_me_test.go`. The verdict exists nowhere else.

Matching the error code as well as the status means this keeps working if that endpoint ever moves to `ResponseErrorLWithStatus`, and it already works against an older server build that returned a semantic 403.

### Load-bearing details

- **`isLoggedIn` is in the probe precondition.** Restoring goes through `setManagerMe()`, which does not set that flag — a stored credential that outlived it would restore, navigate, get bounced by `SuperOnlyRoute`, remount the login page, and probe again. A redirect loop.
- **A `mounted` ref, not an unmount abort.** The cleanup flips a flag rather than aborting, so StrictMode's mount → unmount → remount does not turn one fetch into two, while a real unmount leaves the completion handler inert. This makes "nothing lands after the page is gone" structural instead of resting on every route off `/login` remembering to `logout()` first.
- **antd's `Alert` already carries `role="alert"`** and its `Spin` already carries `aria-live="polite"`. Wrapping either in another live region makes a screen reader announce twice. The spinner is decorative and is `aria-hidden`; the status text is what gets announced. The unit test's antd mock now states antd's real contract — a mock that lied about it is why the double live region survived the unit suite in the first place.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run` — **228 passing across 21 files** (+24 over main), changed files at 100% statements / lines / functions
- [x] `vite build` clean
- [x] End-to-end against the test backend, **15 assertions**, driven in a real browser:
  - [x] **valid token** — lands on `/dashboard`, token byte-identical, zero POSTs to `/v1/manager/login`
  - [x] **repeated visits** — four consecutive loads of `/login`, token never changes
  - [x] **StrictMode** — `/v1/manager/me` fetched exactly once despite double-invoked effects
  - [x] **expired token** — 401 clears it, form returns, `/v1/manager/login` never posted
  - [x] **403 at its real wire shape** (400 + envelope `http_status: 403`) — forbidden copy renders, credential survives, sign-out clears it
  - [x] **502 / connection-refused** — credential survives, error state shows, retry reaches the real backend
  - [x] **escape hatch** — returns to the form, and a probe released afterwards does not navigate away
  - [x] **fail fast** — a hung backend surfaces the error state in 8.5s, not 30s
- [ ] Reviewer sanity check in a browser against a deployment

The E2E scripts were run from a scratch directory and are not included — the repo has no Playwright setup, and adding one is its own change.

Note for anyone re-running locally: use Node 18/20/22 (CI runs 18 and 20). On Node 26 the suite fails in zustand-persist tests because Node's experimental `localStorage` accessor shadows jsdom's — an environment artifact, not a defect.

## Known follow-up

octo-server now has manager console email MFA (#791), which answers `/v1/manager/login` with a challenge and finishes at `/v1/manager/login/verify`. This page does not implement that branch yet. It is not enabled on the deployments this targets, and it wants its own change — tracked in #139.


Jerry-Xin (dismissed):
Reviewed at c2093413 against the checked-out head. The change is sound: it restores a live manager session via a single `/v1/manager/me` probe on mount, classifies probe failures correctly (401 → revoke credential + show form; network/5xx/403 → keep credential + retry with a password escape hatch), and the new `skipAuthRedirect` request option lets the probe own its 401 without tripping the global interceptor. The option defaults to off, and `src/api/index.test.ts` pins both halves of that contract.

Verified independently in this checkout:
- `tsc --noEmit` clean; `vite build` clean.
- `vitest run`: 216/216 passing on Node 22 (CI matrix runs Node 18/20).
- No restore/redirect loop: on a probe 401 the credential is cleared locally and the form is shown without a hard redirect; any subsequent mount sees an empty token and skips probing. The `isLoggedIn` precondition closes the stale-flag loop through `SuperOnlyRoute`.
- No double fetch after restore: `MainLayout`'s profile effect is guarded on `managerCapabilities !== null || managerProfileStatus !== 'idle'`, so the restored profile is reused, not refetched.
- StrictMode double-invocation and retry-while-probe-in-flight are guarded by the `probing` ref and covered by tests.
- Auth handling is clean: the token travels only in the existing `token` request header, nothing is put in URLs, and storage (localStorage via the existing zustand persist key) is unchanged by this PR. Restore reuses the existing token and never calls `/v1/manager/login`, which is exactly what stops the per-uid session cap from being exhausted by redundant sign-ins.

💬 Non-blocking

- If the component unmounts while a probe is in flight, the completion handler (`setManagerMe` + navigate) still runs. Harmless in practice given the short window, but an unmount guard or AbortController would make a late response strictly inert (src/pages/Login/useSessionRestore.ts).
- The password form still assumes a direct `LoginResponse`; once octo-server's manager MFA lands on these deployments, `/v1/manager/login` will answer with a challenge instead. The PR body already discloses this as a follow-up — noted here for continuity with #139.

✅ Highlights

- Correct three-way split of probe outcomes (valid / confirmed 401 / server gave no verdict), each with the right default behavior.
- The interceptor change is opt-in; the default revoke+redirect path is explicitly pinned by tests so it cannot silently regress.
- Strong coverage: success, expiry, transient failure, retry, dismiss, non-Error rejection, and single-live-region a11y.

APPROVE

yujiawei (dismissed):
# Code Review — PR #152 (octo-admin)

Reviewed at head `c209341`, merge-base `0393ca1`. Checked out and verified locally.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Each of the three probe outcomes described in the PR body is implemented and reachable: valid → `setManagerMe` + `navigate(firstManagerPath(...))` (`src/pages/Login/useSessionRestore.ts:55-57`), 401 → `logout()` + form (`:59-61`), anything else → keep credential + retry/escape hatch (`:62-65`). All five new i18n keys are consumed; none are dead.
- **Out of scope additions:** none. The eleven changed files are login page, the single API option it needs, its locale keys, and tests. `skipAuthRedirect` defaults to off and has exactly one caller passing `true`.
- **Deviation:** none. The MFA branch is explicitly declared as follow-up rather than half-built, which is the right call.

Test-plan claims reproduced locally on this head:

| claim | result |
|---|---|
| `tsc --noEmit` clean | ✅ exit 0 |
| `vitest run` — 216 tests / 21 files | ✅ 216 passed, 21 files |
| `vite build` clean | ✅ built in 6.02s |
| browser E2E, 28 assertions | ⚠️ not reproducible — scripts are not in the repo (as the PR states) |

## 2. Code quality

**Quality: Approved** — no P0/P1. The findings below are all P2 and none of them should block merge.

### Security assessment

I looked specifically for a way this widens the trust boundary, and did not find one.

- **No new entry.** Restoring requires `isLoggedIn && scope === 'super' && token !== ''` (`useSessionRestore.ts:32-35`) **and** a 200 from `/v1/manager/me`. Anyone who can forge that localStorage record could already navigate straight to `/admin/dashboard` — `SuperOnlyRoute` (`src/App.tsx:30-35`) only checks `isLoggedIn`/`scope`, and `MainLayout` then does the same `/v1/manager/me` fetch. The probe is strictly *more* validation than the old path, not less.
- **Revocation is preserved.** `skipAuthRedirect` suppresses only the global handler; the probe does its own `logout()` on 401. `src/api/index.test.ts` pins both halves, so the default path cannot silently stop revoking. Confirmed the custom key survives axios config merging into `error.config` — the test exercises the real interceptor, not a stub.
- **No privilege inflation.** `normalizeManagerCapabilities` whitelists known keys and coerces `=== true`; the navigation target comes from `firstManagerPath`, a closed set of literal paths, so there is no open-redirect surface.
- **The `isLoggedIn` precondition is load-bearing.** `setManagerMe` does not set that flag, so a token outliving it would loop `/login → probe → /dashboard → SuperOnlyRoute → /login`. Good catch by the author, and it is pinned by a test.

I also traced every in-app route to `/login` — `MainLayout.tsx:167-168`, `SpaceAdminLayout.tsx:179-180`, `SpaceEntry.tsx:152` and `:167` — and each one either calls `logout()` first or is only reachable in a state where the token is already cleared. That closes the "probe in flight while the user is somewhere else" class of races; see P2-1.

### P2-1 — the probe has no cancellation, and `checking` has no escape hatch

`useSessionRestore.ts:48-69` has no `AbortController`, no mounted guard, and no token-generation check. State corruption from this is **not** reachable today, and I want to record why so a future change does not quietly reopen it:

- `dismiss()` is only rendered by `RestoreErrorBody` (`src/pages/Login/index.tsx:20-22`), i.e. only after the probe has already settled. Clicking retry flips status back to `checking`, which unmounts the dismiss button before a second click can land.
- There is no in-app navigation off `/login` while `checking`, and every code path that navigates there clears the token first, so `restorable` is false and no probe starts.
- Back/forward out of `/login` is a full document navigation in every scenario I could construct, which aborts the request.

What *is* reachable: with `timeout: 30000` (`src/api/index.ts:36`) and a slow-but-alive backend, the operator sits on a spinner for up to 30 seconds with no way to reach the form, where before the change the form was immediate. Suggestion: add a "sign in with a password instead" affordance to the `checking` state too, and/or drop the probe timeout to a few seconds — the probe is a latency optimization, so it should fail fast. An `AbortController` on unmount plus a token-generation check would also make the invariants above structural rather than incidental.

### P2-2 — the classifier reads a server-controlled status, the interceptor reads the transport status

`src/api/index.ts:51` decides whether to revoke from `error.response?.status` (real HTTP status), but `:57` builds `ApiError.status` as `errorEnvelope?.http_status ?? error.response?.status` — preferring the **response body**. `classifyRestoreFailure` (`src/pages/Login/sessionRestore.ts:18`) consumes the latter. So a genuine HTTP 401 whose envelope carries a different `http_status` is classified `retryable` and the rejected credential is kept; the inverse mismatch revokes on a non-401. Both directions are benign in practice — a dead token still 401s on the next call and hits the global handler — but the module now has two answers to "did the server reject this credential?". Worth pinning `classifyRestoreFailure` to the transport status, or at least adding an interceptor test with a non-empty error envelope (the existing one uses `data: {}`, so this path is untested).

### P2-3 — 403 keeps the operator in a locally-authenticated but server-denied state

`sessionRestore.ts:15-19` folds every non-401 into `retryable`, 403 included, and the comment argues the token is still valid. Fair for a genuinely under-privileged manager — but the result is that a demoted admin stays `isLoggedIn`, keeps satisfying `SuperOnlyRoute`, and only ever sees "the server is unreachable, please retry", which is not what happened. There is no terminal access-denied path. Consider splitting 403 into its own state with a message that matches reality rather than reusing the outage copy.

### P2-4 — raw backend error strings override the localized copy

`useSessionRestore.ts:63` stores `error.message`, and `index.tsx:60` prefers it over `t('restore.error.description')`. For the exact failures this feature targets, axios supplies English strings — `Network Error`, `timeout of 30000ms exceeded`, `Request failed with status code 502` — so a zh-CN operator gets untranslated technical text in place of the carefully worded fallback. Suggestion: keep the localized description as the primary copy and demote the raw message to a detail line, or only surface it when it came from the server envelope rather than from axios.

### P2-5 — the success callback runs inside the `try`

`useSessionRestore.ts:57`: `onRestoredRef.current(...)` sits in the try block, so an exception thrown by the navigation callback is caught at `:58` and classified as a probe failure. The user would see "Couldn't verify your session" for a session that verified fine, with capabilities already hydrated. Cosmetic today, but moving the callback after the try/catch costs nothing.

### P2-6 — `skipAuthRedirect` is augmented globally but honored by one instance

The `declare module 'axios'` block (`src/api/index.ts:5-17`) widens `AxiosRequestConfig` process-wide, so `marketplaceApi` (`src/api/marketplace.ts:29`) accepts the option too and type-checks cleanly — but its interceptor at `src/api/marketplace.ts:58` does not read it and will still revoke + hard-redirect. A silent no-op that looks like it works. Either honor the flag there for symmetry, or note in the doc comment that it is specific to the default instance.

### Nit — `src/api/index.test.ts:28`

`Object.defineProperty(window, 'location', ...)` is never restored in `afterEach` (the adapter is). Harmless under vitest's default per-file isolation, but it would bite if anything is ever added to this file below these tests.

### Test quality

Good — these assert behavior, not mock shape. `useSessionRestore.test.tsx` pins the things that actually matter: token byte-identity across a restore (`:106`), the in-flight retry guard with a manually released promise (`:147-170`), and the stale-token-outlives-`isLoggedIn` loop guard (`:87-95`). The antd mock in `index.test.tsx:69-70` reproducing `Alert`'s real `role="alert"` contract is the right instinct — a mock that lies about the contract is why the double live region survived the unit suite in the first place.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved.

This is a real bug with a real cost (leaked server sessions against a per-uid cap), the fix is proportionate, and the 401-vs-no-verdict distinction is the correct axis to split on. The P2s above are follow-ups, not blockers.

## 4. Suggested follow-ups

1. Give `checking` its own escape hatch and shorten the probe timeout (P2-1) — this is the one with user-visible impact.
2. Make `classifyRestoreFailure` read the transport status, and add an interceptor test with a populated error envelope (P2-2).
3. Localize the error copy (P2-4).
4. Move `onRestoredRef.current(...)` out of the `try` (P2-5).

## 5. Recommended human verification

This carries `needs-human-review`; these are the parts I could not settle from this repo alone:

- **Backend contract for `/v1/manager/me`.** Does octo-server return 401 (not 403) for an expired or revoked manager token? The whole "clear vs keep" split rests on that. If revocation ever surfaces as 403, the credential is retained indefinitely.
- **Does the error envelope ever set `http_status` to something other than the transport status?** That is the trigger for P2-2.
- **Browser confirmation on a deployment.** The E2E evidence is credible and detailed, but the scripts are not in the repo, so I verified everything except the browser behavior. The StrictMode single-fetch claim and the 502/connection-refused paths are the ones worth re-running by hand.
- **Confirm MFA (#791) is genuinely disabled on every deployment this ships to.** Not caused by this PR, but if it is enabled anywhere, the password fallback path is already broken there.
- **Cross-tab behavior** (pre-existing, unchanged): the auth store persists to `localStorage` with no `storage`-event sync, so two tabs share one credential record and last-writer-wins. Worth confirming that is acceptable for the admin console.

an9xyz:
Pushed `90982be` addressing all six P2s. Thanks — three of them turned out to have more behind them than the review could see from the diff alone, and two of the "recommended human verification" questions are now settled with evidence.

## Answers to the open questions

**Does octo-server return 401 (not 403) for an expired/revoked manager token?** — **401.** Measured against the test backend:

| request | wire | envelope `http_status` | code |
|---|---|---|---|
| no token | 401 | 401 | `err.shared.auth.token_missing` |
| unknown token (≡ revoked/expired — same cache-miss path) | 401 | 401 | `err.shared.auth.required` |

The clear-vs-keep split rests on solid ground.

**Does the error envelope ever set `http_status` to something other than the transport status?** — **Yes, by design.** octo-server has two renderers: `httperr.ResponseErrorLWithStatus` (wire == `HTTPStatus`) and `httperr.ResponseErrorL`, which pins the wire status to 400 for legacy clients and puts the real status only in the envelope. `pkg/errcode/messages_search.go` states it directly:

> renderer keeps wire status pinned to **400** for legacy compatibility while exposing the real status in `error.http_status`

So P2-2 moves from "worth pinning" to "should be fixed", and it is: `ApiError` now carries `transportStatus` alongside `status`, and `classifyRestoreFailure` reads it. `status` keeps its current meaning because Dashboard, SkillMarket and ExpertMarket all branch on it for 404/409/400. Two new interceptor tests cover both directions of disagreement with a populated envelope.

**Browser confirmation on a deployment** — still yours. Everything below was verified against the test backend from a dev server, not a deployed build.

## The rest

**P2-1** — probe now has its own 8s timeout (was inheriting the client's 30s), and `checking` carries the same "sign in with a password instead" affordance as the error state. Abandoning cancels rather than just switching views: a generation counter plus an `AbortController`, so an operator who gives up and starts typing is not yanked away when the probe later succeeds.

I deliberately did **not** abort on unmount. StrictMode's mount→unmount→remount would turn the verified single-fetch property into two requests, and as you established, every route off `/login` either clears the token first or is a full document navigation. User-driven abandonment is the reachable case, and that is the one that now cancels. Happy to reconsider if you would rather have the structural guarantee and lose single-fetch.

**P2-3** — 403 is its own outcome now, with its own copy and a sign-out exit. The credential is still kept, since it is valid; it just no longer opens this door.

**P2-4** — localized copy is the body, raw string is a dimmed detail line beneath it. Kept rather than dropped, because without it there is no way to tell a timeout from a 502.

**P2-5** — moved out of the `try`. Worth noting the consequence: such an exception now surfaces as an unhandled rejection instead of being swallowed. That is intended — a throwing `navigate()` is the caller's bug and should be loud, not disguised as a probe failure. I dropped the test I wrote for it rather than suppress the rejection to make it pass.

**P2-6** — good catch, and it was the worst of the six: type-checks, reads as working, silently does nothing. `marketplaceApi`'s interceptor now honours the flag. Nit fixed too (`window.location` restored in teardown).

## Verification

`tsc` clean, `vitest run` 224 passing across 21 files (+8), `vite build` clean, changed files at 100% statements/lines/functions.

17 browser assertions against the test backend, including the new paths: the escape hatch returns to the form and a probe released afterwards does not navigate away (no `POST /v1/manager/login`); a hung backend surfaces the error state in **8.5s** rather than 30s; 403 renders its own copy, keeps the credential, and its sign-out clears it. The first round's properties re-verified unchanged — valid token lands on `/dashboard` with the token byte-identical across four consecutive visits, `/v1/manager/me` still fetched exactly once under StrictMode.

Jerry-Xin (dismissed):
Re-review of the follow-up commit `90982be` (tightening after the first round). Approving.

## What the new commit changes (verified against the delta from the previously reviewed head)

- **AbortController + generation guard on the probe** (`src/pages/Login/useSessionRestore.ts`): every interactive exit from the restore flow (`dismiss`, `signOut`) bumps a generation counter and aborts the in-flight request, so a late-arriving response after the operator has moved to the password form is inert — no store mutation, no navigation. The generation check runs before failure classification in the catch path, so an abort can never be misread as a 401 revocation. The navigation callback moved outside the try/catch so a `navigate()` bug cannot masquerade as a probe failure.
- **Fast-fail probe timeout** (`PROBE_TIMEOUT_MS = 8000`, passed through the new `timeoutMs` option): restoring is a latency optimization and now gives up well before the shared 30s client default instead of holding the operator on a spinner.
- **403 gets its own terminal state**: `classifyRestoreFailure` now returns `expired | forbidden | retryable`. A demoted account sees dedicated copy (`restore.forbidden.*`, present in both en-US and zh-CN) and a sign-out exit instead of the "server unreachable" outage text, and the credential is kept until the operator chooses to drop it.
- **Transport-vs-envelope status split** (`src/api/index.ts`): `ApiError.transportStatus` carries the wire status. Both the global 401 interceptor and the restore classifier judge on the wire status, which is the right call given octo-server deliberately pins some endpoints' wire status to 400 while reporting the real status only in the error envelope — the two code paths now provably agree, and `src/api/index.test.ts` pins both halves (envelope-disagrees-in-each-direction).
- **`skipAuthRedirect` honored by the marketplace instance too** (`src/api/marketplace.ts`), closing the "type-checks but silently redirects" trap on the second axios instance.
- The escape hatch is now offered during the `checking` state as well, and the raw transport message is demoted to a detail line under the localized copy.

## Prior-round notes

- **Late probe response after unmount** — addressed. The interactive paths (the reachable hazard: user already typing in the form, then getting carried away) are covered by the generation guard + real abort, with a test pinning that a late success after dismiss neither navigates nor mutates the store. The deliberate decision to keep no unmount cleanup is documented in-code and I traced it independently: nothing on `/login` performs a client-side navigation away from the page (no `Link`s, only `mailto:` anchors); every programmatic `navigate()` fires after the probe has settled or been abandoned; SuperOnlyRoute only sends users to `/login` when `isLoggedIn` is false, which cannot start a probe; browser-level departures are full-page loads that kill the request. Adding abort-on-unmount cleanup would also break the verified StrictMode single-fetch property (cleanup between the double-invoked effects would re-fire the probe), so the trade is the right one.
- **Password path assuming `LoginResponse` / MFA** — unchanged; the MFA branch remains tracked in #139.

## Verification on this head

- `tsc --noEmit` clean; `vite build` clean.
- Full vitest suite: **224/224 passing across 21 files** (previous 216 + 8 new) on Node 22. Note for anyone re-running locally: on Node 26 the suite fails in zustand-persist tests (`localStorage` undefined) because Node 26's experimental `localStorage` accessor shadows jsdom's — that is an environment artifact, not a defect of this change; the baseline is Node 22.
- Re-verified unchanged properties: MainLayout double-fetch guard intact (`managerProfileStatus` gate), restore precondition still `isLoggedIn && scope === 'super' && token !== ''` (no restore/redirect loop with SuperOnlyRoute), token stays header-only, auth storage untouched, default revoke-and-redirect path still pinned by tests.

## Non-blocking observations

- `getManagerMe` forwards `signal: options.signal` unconditionally, so an undefined signal is passed through; harmless with axios, purely cosmetic.
- If `/v1/manager/me` were ever rendered through octo-server's wire-pinning error path (400 on the wire, real status in the envelope), an expired credential would now classify as retryable rather than expired. That matches the interceptor's own criterion and the current backend behavior (E2E in the PR description confirms a wire 401), so this is only worth keeping in mind if the server-side rendering of that endpoint changes.

yujiawei (changes requested):
# Code Review — PR #152 (octo-admin)

Reviewed at head `90982be`, merge-base `0393ca1`. Checked out and verified locally; also read the corresponding `octo-server` / `octo-lib` sources to settle the backend contract this change depends on.

This is the second round. Round one (at `c209341`) approved with six P2 follow-ups; this head is the commit that addresses them. All six are genuinely addressed, and the core of the change — a `/v1/manager/me` probe that restores a live session instead of minting a redundant server-side one — is correct and worth landing. One of the round-two additions, however, cannot execute against the real backend, and I'd like that resolved before merge.

## 1. Spec compliance

**Spec: ❌** — one deviation, no missing or out-of-scope work.

- **Missing:** none. All six round-one follow-ups are implemented: probe timeout cut to 8s (`useSessionRestore.ts:26`), `AbortController` + generation counter (`:54-55`, `:113-118`), escape hatch during `checking` (`index.tsx:19`, `:45-47`), a dedicated 403 state (`sessionRestore.ts:28`, `index.tsx:102-118`), `transportStatus` split (`api/index.ts:36-49`), `marketplaceApi` honoring `skipAuthRedirect` (`marketplace.ts:62`), localized copy promoted over the raw axios string (`index.tsx:68-87`), and the success callback moved out of the `try` (`useSessionRestore.ts:102-108`).
- **Out of scope:** none. Twelve files: the login page, the one API option it needs, its locale keys, and tests. `skipAuthRedirect` still defaults to off with exactly one caller passing `true`.
- **Deviation:** the 403 `forbidden` branch is wired to a status code this endpoint never puts on the wire, so the state it adds is unreachable in production. Detail in P1-1 below. The PR narrative and the test names both assert this path works; it does not.

Test-plan claims reproduced on this head:

| claim | result |
|---|---|
| `tsc --noEmit` clean | ✅ exit 0 |
| `vite build` clean | ✅ built in 6.00s |
| `vitest run` — "216 passing across 21 files" | ⚠️ actually **224** passing across 21 files — the PR body was not refreshed after this round (also still cites `timeout of 30000ms exceeded`, now 8000ms) |
| browser E2E, 28 assertions | ⚠️ not reproducible — scripts are not in the repo, as the PR states |

## 2. Code quality

**Quality: Changes-Requested** — one P1, the rest advisory.

### Security assessment

I looked hard for a way this widens the trust boundary and did not find one.

- **No new entry.** Restoring requires `isLoggedIn && scope === 'super' && token !== ''` (`useSessionRestore.ts:44-47`) *and* a 200 from `/v1/manager/me`. `setManagerMe` (`store/auth.ts:80-87`) writes only `name`/`role`/`uid`/`capabilities`/`status` — it cannot set `token`, `scope`, or `isLoggedIn`, so the restore path cannot elevate anything. Anyone able to forge that localStorage record could already navigate straight to `/dashboard`, where `SuperOnlyRoute` (`App.tsx:30-34`) checks the same two fields and `MainLayout` runs the same probe. This is strictly *more* validation than the old path.
- **Revocation is preserved and now properly pinned.** `skipAuthRedirect` suppresses only the global handler; the probe does its own `logout()` on 401. `api/index.test.ts:49-97` exercises the real interceptor across four cases including both envelope/transport mismatch directions.
- **The 401 path is correct against the real backend.** octo-lib's `AuthMiddleware` (`pkg/wkhttp/http.go:272-290`) answers a missing/invalid/expired token with a genuine wire `401` and a bare `{"msg": …}` body — no error envelope — so `transportStatus` is 401 and `classifyRestoreFailure` returns `expired`. Credential cleared, form shown. This is the load-bearing case and it works.
- **No privilege inflation or open redirect.** `normalizeManagerCapabilities` whitelists keys and coerces `=== true`; `firstManagerPath` returns one of a closed set of literal paths, with `/no-access` as the terminal fallback (`capabilities.ts:73`), and that route exists (`App.tsx:265`).
- **`isLoggedIn` in the precondition is still load-bearing** and still pinned by a test (`useSessionRestore.test.tsx:97-105`).

### P1-1 — the 403 branch cannot fire; a demoted admin gets outage copy and an infinite Retry

`classifyRestoreFailure` judges on the transport status:

```ts
// src/pages/Login/sessionRestore.ts:26-28
const status = error instanceof ApiError ? error.transportStatus : undefined
if (status === 401) return 'expired'
if (status === 403) return 'forbidden'
```

But `/v1/manager/me` never returns wire 403. Tracing the server side:

- `octo-server modules/user/api_manager.go:174-177` — `me()` rejects a non-manager role with `respondManagerForbidden(c)`.
- `octo-server modules/user/api_i18n.go:155-157` — that calls `httperr.ResponseErrorL(c, errSharedForbidden, nil, nil)`.
- `octo-server pkg/httperr/respond.go:68-71` — `ResponseErrorL` passes `useSemanticStatus=false`, so `transportStatus := http.StatusBadRequest`. The canonical 403 (`pkg/i18n/codes/shared.go:58-64`) is carried only in the body.
- `octo-server pkg/i18n/renderer.go:55-64` — the envelope is `c.JSON(400, {"error": {"code": "err.shared.auth.forbidden", "http_status": 403, …}, "status": 400})`.
- The backend pins this in its own test: `octo-server modules/user/api_manager_me_test.go:155` — `assert.Equal(t, http.StatusBadRequest, w.Code, "forbidden is pinned to wire 400 (D14)")`.

So the client actually receives `ApiError { status: 403, transportStatus: 400, code: 'err.shared.auth.forbidden' }` and classifies it **`retryable`**.

Failure scenario: a super admin whose console role is revoked server-side opens `/admin/login` holding a still-valid token. Probe → wire 400 → `retryable` → they are shown `restore.error.title` / "The server is unreachable right now. Your saved credentials were kept — please retry." plus a Retry button that will return the same 400 every time. The real reason does appear, but only as the small grey detail line (`index.tsx:74-85`), under a headline that contradicts it. `RestoreForbiddenBody` (`index.tsx:102-118`), `signOut` (`useSessionRestore.ts:139-144`), the `forbidden` status, and eight `restore.forbidden.*` locale entries are dead code.

This is the exact scenario the round-two change set out to fix — `sessionRestore.ts:10-12` argues that reusing the outage copy for 403 "是在撒谎，而且没有终态", which is precisely what still happens.

The tests do not catch it because they hand-build a shape the backend never emits: `new ApiError('forbidden', 403)` (`useSessionRestore.test.tsx:158`, `sessionRestore.test.ts:11`) leaves `transportStatus` defaulting to `status` via the constructor (`api/index.ts:49`). `useSessionRestore.test.tsx:157` is named *"gives a demoted admin a terminal state instead of outage copy"* — it asserts production behavior that does not occur.

There is a real design tension here, and I don't think it should be papered over: reading the envelope status would reintroduce the divergence round one flagged, and `api/index.test.ts:85-97` deliberately pins "does not revoke on a wire 400 whose envelope claims 401". The narrow fix that keeps both properties is to classify on the pair rather than one axis, since this endpoint has exactly two failure shapes:

```ts
if (transportStatus === 401) return 'expired'                        // octo-lib AuthMiddleware
if (transportStatus === 400 && status === 403) return 'forbidden'    // D14-pinned ResponseErrorL
return 'retryable'
```

Matching on `code === 'err.shared.auth.forbidden'` would be more precise still. Whichever you pick, please state in the comment which axis is authoritative for which question, and add a test built from the real wire shape (status 400 + `http_status` 403) rather than a synthesized `ApiError`. If you'd rather not carry the D14 special case in the client, the alternative is to move this endpoint to `ResponseErrorLWithStatus` server-side — but that is a separate change in another repo, and until it lands the branch here is dead.

### P2-1 — the escape hatch is what keeps `checking` from being a trap, and nothing pins that

On the success path `status` is never moved off `'checking'`; the page relies on the caller navigating away. That is fine today, and the deliberate choice to let a throwing `navigate()` surface as an unhandled rejection (`useSessionRestore.ts:102-107`) is defensible — but note the practical effect is a console log, not an Error Boundary, so the visible outcome of that bug would be a spinner. The only reason it isn't a trap is the dismiss button added this round (`index.tsx:19`). Worth a test asserting the escape hatch is present in `checking` specifically *because* the success path never resolves the status itself, so a future refactor can't quietly remove it.

### P2-2 — the no-unmount-abort invariant is incidental, not structural

`useSessionRestore.ts:120-127` deliberately skips cleanup so StrictMode stays at one fetch, and argues every path off `/login` either calls `logout()` first or is a document navigation. I traced all five and it holds today: `MainLayout.tsx:167-168` and `SpaceAdminLayout.tsx:180-182` both `logout()` first; `SpaceEntry.tsx:152` is the `no-token` branch; `SpaceEntry.tsx:167` is `no-space`, which `logout()`s at `:99`; and `App.tsx:32` only fires when `!isLoggedIn`, where `restorable` is false.

But that is four call sites' worth of coincidence, not an enforced invariant. Add one client-side `navigate('/login')` that doesn't `logout()` first and a late probe can `setManagerMe` + `navigate(…, {replace: true})` up to 8s after the user left, silently. A cleanup that calls `abandon()` only on real unmount, or a comment naming the invariant at each of those call sites, would make it structural.

### P2-3 — `skipAuthRedirect` is now a global type augmentation honored by both clients

Making `marketplaceApi` honor the flag (`marketplace.ts:62`) fixed round one's silent-no-op, but the flag now suppresses mandatory 401 revocation on both instances for anyone who passes it. It is opt-in with one caller, so this is fine as shipped — but the doc comment at `api/index.ts:7-14` should say plainly that the login-page restore probe is the only sanctioned caller, so the next person reaching for it has to justify it.

### P2-4 — refresh the PR body before merge

It still reports 216 tests and a 30s timeout; head is 224 tests and 8s. Minor, but the body is the record of what was verified.

### Test quality

Strong overall — these assert behavior, not mock shape. `useSessionRestore.test.tsx:137-155` releases a deferred promise after `dismiss()` to prove a late success cannot navigate the user away, and checks the signal actually aborted. `api/index.test.ts:68-97` drives the real interceptor through both envelope-mismatch directions. `index.test.tsx:69-70` reproducing antd's real `role="alert"` contract in the mock remains the right instinct. The one gap is the 403 tests described in P1-1, which pin an invented wire shape.

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ❌ (P1-1) and Quality Changes-Requested.

To be clear about proportion: everything else here is good work, the 401 path is correct against the real backend, and no user ends up worse off than on `main` — the demoted admin can still click through to the password form. I'm blocking on P1-1 because it ships a state machine branch that cannot be entered, plus locale keys and two tests that assert it can, and because the fix is small.

## 4. Suggested follow-ups

1. Fix the 403 classification against the real wire shape and add a test built from `status 400 + http_status 403` (P1-1). Refresh the `sessionRestore.ts` comment to say which status axis answers which question.
2. Pin the `checking` escape hatch with a test that names why it is required (P2-1).
3. Make the no-unmount-abort invariant explicit at the `navigate('/login')` call sites, or enforce it in the hook (P2-2).
4. Tighten the `skipAuthRedirect` doc comment to name its only sanctioned caller (P2-3).
5. Refresh the PR body's test counts and timeout figure (P2-4).

## 5. Recommended human verification

This carries `needs-human-review`. These are the parts I could not settle from source alone:

- **Which server build each target deployment runs.** The wire-400 contract above is read from `main`; if a deployment runs an older build whose `/v1/manager/me` returned real 403, the classification behaves differently there. Worth confirming before choosing between the client-side and server-side fix.
- **Browser confirmation on a deployment.** The E2E evidence is detailed and credible but the scripts are not in the repo, so I verified everything except actual browser behavior. The StrictMode single-fetch claim, the 502/connection-refused paths, and the new dismiss-during-probe abort are the ones worth re-running by hand.
- **Confirm manager console MFA is genuinely disabled on every deployment this ships to.** The PR body discloses this as a follow-up, and the feature is now merged server-side — if it is enabled anywhere, the password fallback path is already broken there, which matters more now that the fallback is the escape hatch from a failed probe.
- **Cross-tab behavior** (pre-existing, unchanged): the auth store persists to `localStorage` with no `storage`-event sync, so two tabs share one credential record, last writer wins. Worth confirming that is acceptable for the admin console.

an9xyz:
`d3c2606` — you were right, and thank you for tracing it to the server. I verified the chain independently before changing anything:

```
me() 非管理角色 → respondManagerForbidden
  → httperr.ResponseErrorL(errSharedForbidden)   // useSemanticStatus = false
    → transportStatus := http.StatusBadRequest   // 400 on the wire
       SemanticStatus  := 403                    // lives only in the envelope
```

plus `api_manager_me_test.go` pinning `assert.Equal(t, http.StatusBadRequest, w.Code, "forbidden is pinned to wire 400 (D14)")`. Confirmed: the branch could not fire, and my tests passed only because `new ApiError('forbidden', 403)` lets the constructor default `transportStatus` to `status` — a combination the backend never emits.

Worth naming the failure mode: the previous round *discovered* the wire-pinning pattern while fixing the transport-vs-envelope split, and then failed to apply it to the branch it was adding in the same commit. Finding the rule and not applying it one file over.

## P1-1 — fixed, and stated as two axes

Rather than special-casing the `(400, 403)` pair, the comment in `sessionRestore.ts` now says which axis answers which question, because that is the thing that was actually confused:

- **"Is the credential still valid?"** → transport status. octo-lib's `AuthMiddleware` answers with a genuine wire 401 and a bare `{"msg": …}` body. Must agree with the global interceptor — round one's point.
- **"What did authorization decide?"** → the envelope. For any `ResponseErrorL` endpoint the verdict exists nowhere else.

```ts
if (error.transportStatus === 401) return 'expired'
if (error.status === 403 || error.code === MANAGER_FORBIDDEN_CODE) return 'forbidden'
return 'retryable'
```

Matching the code as well as the status, as you suggested. That also **settles your first human-verification item**: a semantic 403 sets `status === 403` too, so an older server build whose `/v1/manager/me` returned a real 403 classifies identically. Both shapes have tests, so the deployment's server version no longer has to be answered before merging.

Tests rebuilt from the real wire shape. Two new ones in `src/api/index.test.ts` drive the **real interceptor** with octo-server's actual forbidden envelope and octo-lib's actual bare 401, then assert what `classifyRestoreFailure` makes of each — so the interceptor and the classifier can no longer be tested against two different imagined shapes. The browser E2E now serves wire 400 + envelope 403 as well.

## P2-2 — made structural, without losing single-fetch

You are right that four call sites of coincidence is not an invariant. A `mounted` ref closes it and keeps the StrictMode property: the cleanup **flips a flag rather than aborting**, so mount → unmount → remount sets it back to true and the second effect still short-circuits on `probing`, while a real unmount leaves it false and the completion handler returns before touching the store. Aborting in cleanup was the option I rejected — it would leave StrictMode with a cancelled first probe and a second that early-returns, i.e. no probe at all. Test added: unmount mid-flight, release the promise, assert no navigation and no store write.

## P2-1, P2-3, P2-4

- The `checking` escape-hatch test now names *why* it is load-bearing: the success path deliberately never resolves `status` off `checking`, so without that button a navigation that does not happen is a dead end. Pinned from both sides — the hook test asserts `status` stays `checking` after a successful probe.
- `skipAuthRedirect`'s doc comment now names the login probe as its only sanctioned caller and says plainly that it disables mandatory revocation on both instances.
- PR body refreshed: 228 tests, 8s timeout, 15 E2E assertions. Also added your Node 26 note for anyone re-running locally.

## Remaining human verification

Two of your four are now settled from source (the server build question above; the 401 contract from the previous round). The other two still need a person:

- **Browser confirmation on a deployment** — everything here was verified against the test backend from a dev server, not a deployed build.
- **MFA genuinely disabled everywhere this ships** — this matters more now that the password form is the escape hatch from a failed probe, as you note.

**Cross-tab last-writer-wins** is pre-existing and unchanged by this PR; worth its own issue rather than widening this one.

## Verification

`tsc` clean, `vitest run` **228 passing across 21 files**, `vite build` clean, changed files at 100% statements/lines/functions. 15 browser assertions against the test backend, all passing, including the corrected 403 path.

Jerry-Xin (approved):
Re-review at `d3c2606` (delta vs `90982be`: one commit, seven files — classifier, its tests, the hook's unmount guard, and doc comments). This head resolves the round-two changes-requested review; every point was independently re-verified against the backend sources, not taken on faith.

## Adjudication of the prior review

**P1-1 — the 403 branch was dead code on `90982be`; it is now wired to the shape the server actually sends. Confirmed fixed.**

The prior analysis of the backend contract is correct, byte-verified here against `octo-server` main (`86336f2`): `me()` rejects a non-manager-console role with `respondManagerForbidden` (`modules/user/api_manager.go`), which renders through `httperr.ResponseErrorL` → `respondL(..., useSemanticStatus=false)` → wire status pinned to **400** (D14), the canonical 403 (`err.shared.auth.forbidden`, `pkg/i18n/codes/shared.go`) carried only in `error.http_status`. The backend pins this itself: `api_manager_me_test.go` asserts `w.Code == http.StatusBadRequest` with the comment *"forbidden is pinned to wire 400 (D14)"*. So on the previous head, where the classifier judged on `transportStatus === 403`, the `forbidden` state was unreachable in production and a demoted admin landed on outage copy with an infinite Retry — and the old tests masked it by hand-building `new ApiError('forbidden', 403)`, which defaults `transportStatus` to 403, a combination the backend never emits.

The fix splits the two questions onto the two axes, which is the right resolution:

```ts
if (!(error instanceof ApiError)) return 'retryable'
if (error.transportStatus === 401) return 'expired'
if (error.status === 403 || error.code === MANAGER_FORBIDDEN_CODE) return 'forbidden'
```

- *"Is the credential still valid?"* stays on the transport axis (`transportStatus === 401`), identical to the 401 interceptor's criterion — verified against `octo-lib` `AuthMiddleware`, which answers missing/invalid/expired tokens with a genuine wire 401. The two mismatch directions remain pinned (`revokes on the transport status even when the envelope disagrees` / `does not revoke on a wire 400 whose envelope claims 401`), so round one's divergence property is preserved.
- *"What did authorization decide?"* now reads the envelope (`status`/`code`), the only place that verdict exists for this endpoint. Matching the code as well as the status keeps the branch working if the endpoint ever migrates to `ResponseErrorLWithStatus` (both axes 403 — pinned by a dedicated test), and a bare upstream wire 403 still classifies as before. Since this classifier is fed only by the `/v1/manager/me` probe, the broader match carries no misclassification risk.
- The synthesized-shape problem is fixed at the root: tests now construct errors from the real wire shapes (`wire 400 + envelope {code: 'err.shared.auth.forbidden', http_status: 403}`; bare-msg wire 401), and two new tests in `src/api/index.test.ts` drive the **real interceptor** and hand its output to the classifier, joining the two halves that were previously tested against imagined shapes.

**P2-1 — addressed.** The success path's deliberate non-convergence is now pinned (`expect(state?.status).toBe('checking')` after a successful restore), and `index.test.tsx` documents in-place why the escape hatch is load-bearing rather than cosmetic: without it, a missing navigation would leave `checking` as a dead end.

**P2-2 — addressed, structurally.** A `mounted` ref replaces the cross-four-files coincidence: unmount flips the flag instead of aborting, and both completion handlers bail on `!mounted.current`. The choice of flag-over-abort is the right one and the comment explains why — aborting in cleanup would let StrictMode's mount → unmount → remount kill the only probe (the second effect returns early on `probing.current`), while the flag is restored by the remount effect before any response can land. A new test settles a deferred probe after unmount and asserts nothing is written and nothing navigates.

**P2-3 — addressed.** The `skipAuthRedirect` doc now names the login-page probe as the only sanctioned caller and requires justification for any new one.

**P2-4 — addressed.** The PR body now matches reality: it claims **228 tests across 21 files**, and a fresh run on Node 22 produces exactly 228/228 in 21 files. The E2E section no longer overclaims: 15 assertions, with an explicit disclosure that the scripts were run from a scratch directory and are not committed because the repo has no Playwright setup. No dangling references to E2E scripts exist anywhere in the tree (grepped).

## Regression checks on this head (run locally, Node 22)

- `tsc --noEmit` clean; `vite build` clean.
- `vitest run`: **228/228 across 21 files**, fresh run.
- Single-request property: still one `getManagerMe` call per mount (`toHaveBeenCalledTimes(1)` pinned); the mounted-flag cleanup does not interact with the generation counter or the in-flight guard.
- Abort is not misread as auth failure: `dismiss()` bumps the generation and aborts; the rejection short-circuits on the generation check before classification — late settle after dismiss neither navigates nor touches the credential (pinned).
- Locale parity: both `en-US` and `zh-CN` carry the full `restore.*` key set including the forbidden state.
- CI green on this head (build on Node 18 and 20, secret/dependency scans, title lint).

## 💬 Non-blocking notes

- The StrictMode single-request property is argued correctly in code and covered by the (uncommitted) browser E2E, but no committed unit test renders under `React.StrictMode`. Worth adding one when convenient so the flag-over-abort invariant can't regress silently.
- The remaining unchecked test-plan item — a reviewer sanity pass in a real browser against a deployment — is a human step, not a code blocker.

## ✅ Highlights

- The two-axis split is documented at length exactly where the next maintainer will look, naming the server-side functions and the D14 compatibility pin by name.
- Tests are built from the wire shapes the backend provably sends, with a forward-compatibility case for the eventual semantic-status migration.
- The unmount guard became a structural guarantee instead of a documented hope.

yujiawei (approved):
# Code Review — PR #152 (octo-admin)

Reviewed at head `d3c2606`, merge-base `0393ca1`. Checked out and verified locally, and read the corresponding `octo-server` / `octo-lib` sources to settle the backend contract this round turns on.

Round three. Round one approved with six advisory items; round two (`90982be`) addressed all six but introduced a blocking defect — the new 403 `forbidden` branch was keyed on a status code `/v1/manager/me` never puts on the wire, so the state was unreachable and two tests asserted a shape the backend does not emit. **That is fixed here, and I verified the fix against the real backend rather than against the PR narrative.**

## 1. Spec compliance

**Spec: ✅** — nothing missing, nothing out of scope, no deviation.

### The round-two blocker is genuinely closed

I traced the wire shape end to end instead of taking the commit message for it:

| step | source | result |
|---|---|---|
| handler | `octo-server modules/user/api_manager.go:176` — `me()` role guard | `respondManagerForbidden(c)` |
| helper | `modules/user/api_i18n.go:155-157` | `httperr.ResponseErrorL(c, errSharedForbidden, nil, nil)` |
| transport | `pkg/httperr/respond.go:69-71` — `useSemanticStatus == false` | wire status pinned to **400** (D14) |
| semantic | `pkg/i18n/codes/shared_test.go:20` pins `err.shared.auth.forbidden` | `HTTPStatus = 403` |
| envelope | `pkg/i18n/renderer.go:56-65` (renderer injected at `main.go:203`) | `error.http_status = 403`, `error.code = "err.shared.auth.forbidden"` |
| client | `src/api/index.ts:83,85` | `ApiError.status = 403`, `transportStatus = 400`, `code = "err.shared.auth.forbidden"` |
| classify | `src/pages/Login/sessionRestore.ts:44` | `'forbidden'` — **reachable** |

`octo-server`'s own `modules/user/api_manager_me_test.go:151-158` asserts `w.Code == 400` plus the code string, so the client and server now pin the same shape from both sides. The new test at `src/api/index.test.ts:86-107` is exactly the right construction: it drives the **real** interceptor with the **real** wire body and then hands the resulting `ApiError` to `classifyRestoreFailure`, so the two halves can no longer drift apart while both stay green.

The two-axis split is also correctly reasoned, not just correctly coded: `transportStatus` answers "was the credential itself rejected" (same criterion as the global 401 interceptor at `src/api/index.ts:77`, so one module cannot give two answers), and the envelope answers "what was the authorization verdict" — which is information the transport layer genuinely does not carry under D14. `sessionRestore.ts:14-28` states which axis answers which question, as requested.

### Every other round-two item is addressed

| item | status | evidence |
|---|---|---|
| pin the `checking` escape hatch, naming *why* it is required | ✅ | `index.test.tsx:136-142`, `useSessionRestore.test.tsx:128-130` (`expect(state?.status).toBe('checking')` on the success path) |
| make the no-unmount-abort invariant structural | ✅ | `useSessionRestore.ts:58,124,136-138` + test at `:167-183` |
| tighten the `skipAuthRedirect` doc comment to name its only sanctioned caller | ✅ | `src/api/index.ts:10,15-16` |
| refresh the PR body's numbers | ❌ | still says 216 tests / `30000ms`; see P2-4 |

### Verification on this head

| claim | result |
|---|---|
| `tsc --noEmit` | ✅ exit 0 |
| `vite build` | ✅ built in 5.86s |
| `vitest run` | ✅ **228** passed across 21 files (Node 22) |
| browser E2E, 28 assertions | ⚠️ not reproducible — scripts are not in the repo, as the PR states |

**I independently checked the StrictMode property this round's change puts at risk.** The new `mounted` ref adds a cleanup function to the very effect whose single-fetch behavior is load-bearing, and the repo has no StrictMode test, so I wrote a throwaway one (not committed) that renders the hook inside `<StrictMode>` with a deferred probe. Result: `getManagerMe` is called exactly **once**, and the restore still lands (`managerProfileStatus === 'loaded'`, navigation fired) after the double-invoked cleanup has flipped `mounted` false and the re-run has flipped it back. The reasoning in the comment at `useSessionRestore.ts:127-135` is correct — flipping a flag is safe where aborting would not be, because a naive abort-on-unmount would kill probe #1 while `probing.current` still short-circuits probe #2, yielding zero fetches. The property holds; only the pin is missing (P2-3).

## 2. Code quality

**Quality: Approved** — no P0/P1. Four P2s and a nit, none of which should hold up merge.

### Security assessment

This carries `needs-human-review`, so I looked hard for a widened trust boundary and did not find one.

- **No new entry.** Restoring requires `isLoggedIn && scope === 'super' && token !== ''` (`useSessionRestore.ts:44-47`) **and** a 200 from `/v1/manager/me`. `setManagerMe` (`src/store/auth.ts:80-87`) writes only `name` / `role` / `uid` / `capabilities` / `status` — it cannot set `token`, `scope`, or `isLoggedIn`. Anyone able to forge that localStorage record could already navigate straight to `/dashboard`, where `SuperOnlyRoute` (`src/App.tsx:30-34`) checks the same two fields and `MainLayout` runs the same probe. This is strictly *more* validation than the pre-change path, not less.
- **Restore corrects stale privilege rather than trusting it.** `setManagerMe` overwrites `role` and `capabilities` from the server response, so a locally-stashed `superAdmin` cannot outlive a demotion. I grepped for client-side `role`-based gating of manager UI and found none — every gate is capability-driven (`CapabilityRoute`, `firstManagerPath`), and `normalizeManagerCapabilities` whitelists keys and coerces `=== true`. `firstManagerPath` returns one of a closed set of literal paths, so there is no open-redirect surface.
- **Revocation is preserved and pinned on both instances.** `skipAuthRedirect` suppresses only the global handler; the probe does its own `logout()` on 401. `src/api/index.test.ts:50-131` exercises the real interceptor across six cases including both envelope/transport mismatch directions. `marketplaceApi` honors the flag too (`src/api/marketplace.ts:62`) and still produces a proper `ApiError` (`:65-67`) — it passes no `transportStatus`, which is correct because that instance does no D14 wire-pinning, so the two axes coincide there.
- **The 401 path is correct against the real backend.** `octo-lib`'s `AuthMiddleware` (`pkg/wkhttp/http.go:306-330`, `authErrorSpec` at `:345-377`) answers a missing / invalid / expired token with `TransportStatus: 401` in **every** branch, so `transportStatus === 401 → expired` always fires and the credential is cleared. This is the load-bearing case and it works. (The *body* is not what the comment says — see P2-2.)
- **`isLoggedIn` in the precondition is still load-bearing** against the `/login → probe → /dashboard → SuperOnlyRoute → /login` loop, and still pinned (`useSessionRestore.test.tsx:97-105`).

### P2-1 — an infrastructure 403 is reported to the operator as a revoked account

`sessionRestore.ts:44` matches on `error.status === 403 || error.code === MANAGER_FORBIDDEN_CODE`. `ApiError.status` falls back to the transport status when there is no envelope (`src/api/index.ts:83`), so a **bare** 403 — nginx, a WAF rule, an ingress misconfiguration mid-deploy, an expired mTLS cert — satisfies the first disjunct with no backend authorization verdict behind it at all.

I confirmed this by execution rather than by reading: driving the real interceptor with `status 403` and an HTML body yields `transportStatus 403`, `status 403`, `code undefined`, and `classifyRestoreFailure` returns `'forbidden'`. The operator then gets *"This account has no admin console access — your admin permissions have been revoked"* for a proxy hiccup, and `RestoreForbiddenBody` (`src/pages/Login/index.tsx:102-118`) offers exactly one button: **Sign out**, which destroys a still-valid credential. Compare the `error` state, which correctly offers both Retry and *sign in with a password instead*.

The narrow fix is to require the envelope, since a genuine `respondManagerForbidden` always carries one:

```ts
if (error.code === MANAGER_FORBIDDEN_CODE) return 'forbidden'
if (error.status === 403 && error.code !== undefined) return 'forbidden'  // semantic-status migration
return 'retryable'
```

Independently, `forbidden` would be better with the same non-destructive escape hatch the `error` state has, so the only offered exit is not the one that costs a credential.

### P2-2 — the "bare `{"msg": …}`" 401 contract is out of date, in a commit whose whole thesis is not testing invented shapes

`sessionRestore.ts:18-20` and the comment at `src/api/index.test.ts:110` both assert that `AuthMiddleware` answers with a wire 401 and "a bare `{"msg": …}`, not even an error envelope". That stopped being true. At the `octo-lib` revision `octo-server@main` pins (`v0.0.0-20260811160929`), `AuthMiddleware` renders through `c.RenderError` (`pkg/wkhttp/http.go:309-316`, `authErrorSpec` at `:345-377`), and `octo-server` injects `pkg/i18n.ErrorRenderer` at `main.go:203`, which emits the full compatibility envelope (`pkg/i18n/renderer.go:56-65`):

```json
{"error":{"code":"err.shared.auth.token_invalid","message":"…","details":{},"http_status":401},"msg":"…","status":401}
```

Transport status is 401 in every branch, so **classification is unaffected and no behavior is wrong**. But the test at `src/api/index.test.ts:109-117` and the `WIRE.unauthenticated` fixture at `sessionRestore.test.ts:12` pin a body the backend does not send — the same category of mistake this commit exists to correct, just pointed the other way. Giving that test the real envelope is a two-line change and buys something real: it proves an envelope claiming `http_status: 401` cannot hijack the expired verdict from the transport axis.

### P2-3 — the StrictMode single-fetch invariant still has no test

`useSessionRestore.test.tsx:50` mounts `<Probe />` bare. The exactly-one-fetch assertion at `:119` therefore never exercises the setup → cleanup → setup sequence, which is precisely the sequence this round's `mounted` ref interacts with. I verified out-of-tree that the property does hold today (above), so this is a coverage gap rather than a defect — but it is a gap over an invariant that two consecutive rounds have now had to reason about by hand, and it is ~15 lines to close.

### P2-4 — a definitive 401 arriving after unmount no longer revokes the credential

`useSessionRestore.ts:87` returns on `!mounted.current` *before* classification, so if the page unmounts while a probe is in flight and the response is a 401, `logout()` never runs and the dead token stays in localStorage. Before this round the late 401 would have revoked it (at the cost of a state write after unmount), so this is a small deliberate trade rather than an oversight.

It is benign and self-healing: the token is already dead server-side, so the next call anywhere hits the global interceptor and revokes, and a later `/login` mount re-probes and lands on `expired`. Worth a sentence in the comment at `:127-135` acknowledging the trade, and the unmount test at `:167-183` only covers a late *success* — a late *401* case would document the choice.

### Nit — PR body still reports round-one numbers

216 tests and `timeout of 30000ms exceeded`; the head is 228 tests across 21 files and `PROBE_TIMEOUT_MS = 8000`. Repeat of a round-two note. The body is the record of what was verified, so it is worth 30 seconds before merge.

### Test quality

Strong, and better this round than last. The tests assert behavior rather than mock shape: `useSessionRestore.test.tsx:137-163` releases a deferred promise after `dismiss()` to prove a late success cannot navigate the user away *and* checks the signal actually aborted; `:167-183` does the same for unmount; `api/index.test.ts:69-131` drives the real interceptor through both envelope-mismatch directions. The fixture comment at `sessionRestore.test.ts:5-11` — explaining that `new ApiError('forbidden', 403)` defaults `transportStatus` to `403` and is why dead code tested green — is the kind of note that stops the same mistake from being made again. The remaining gaps are P2-2 and P2-3.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved.

The round-two blocker is properly fixed, and fixed at the right layer: the classifier now asks each status axis the question that axis can actually answer, the choice is documented at the call site, and the test drives the real interceptor output rather than a hand-built `ApiError`. The underlying change remains worth landing — it stops a redundant sign-in from stranding a server-side session against a per-uid cap, and the 401-versus-no-verdict split is the correct axis to split on.

The four P2s are follow-ups. None of them makes a user worse off than `main`, and none is reachable on the paths this feature actually runs on day to day.

## 4. Suggested follow-ups

1. Require the error envelope before classifying `forbidden`, and give that state a non-destructive exit (P2-1). This is the one with real user-visible impact.
2. Update the auth-failure wire shape in the comment and the test to the envelope `octo-lib` actually renders today (P2-2).
3. Add a `<StrictMode>` test pinning exactly one probe (P2-3).
4. Note the late-401-after-unmount trade in the comment, and cover it in the unmount test (P2-4).
5. Refresh the PR body's test count and timeout figure.

## 5. Recommended human verification

`needs-human-review` is on this PR. These are the parts I could not settle from source alone:

- **Browser confirmation on a deployment.** The E2E evidence is detailed and credible, but the scripts are not in the repo, so everything except actual browser behavior is what I verified. The StrictMode single-fetch claim I reproduced in a unit harness; the 502 / connection-refused paths and dismiss-during-probe abort are worth re-running by hand.
- **Which `octo-server` build each target deployment runs.** The wire-400-plus-envelope-403 contract is read from `main`. A deployment on an older build whose `/v1/manager/me` returned a real 403 still classifies correctly here (both disjuncts cover it), but the reverse — a build predating the D14 pin on that endpoint — is worth a moment's thought before shipping.
- **Whether any environment fronts `/api/v1/manager/me` with something that can emit a bare 403** (WAF, IP allowlist, auth proxy). That is the trigger for P2-1.
- **Confirm manager console MFA is genuinely disabled everywhere this ships.** The PR discloses this as a follow-up and the feature is merged server-side; if it is enabled anywhere, the password fallback is already broken there, which matters more now that the fallback is the escape hatch from a failed probe.
- **Cross-tab behavior** (pre-existing, unchanged): the auth store persists to `localStorage` with no `storage`-event sync, so two tabs share one credential record and the last writer wins.

## 6. Review coverage

What I did **not** examine, so it is visible rather than implied: the login page's visual presentation and theming beyond the state machine and its a11y wiring; the `dist` bundle-size warning (pre-existing); the CI matrix on Node 18/20 (I ran Node 22 locally — CI's `build (18)` and `build (20)` both pass on this head); interaction with the other open PRs touching auth (#143 space-console sign-in, #139 manager MFA) beyond confirming this branch is `MERGEABLE` and that neither overlaps these twelve files today.

PR #161, opened 2026-09-03
feat(space-admin): honor ?spaceId= on admin entry so multi-space admins land on the right space

an9xyz:
## Summary

Space admins who manage more than one space used to be dropped on `managed[0]` every time they opened the admin backend from the main app's space-management entry, and had to switch to the right space by hand. SpaceEntry now honors a `?spaceId=` query param passed by octo-web and navigates to that space when the user actually manages it, falling back to the existing default otherwise.

Pairs with [Mininglamp-OSS/octo-web#1620](https://github.com/Mininglamp-OSS/octo-web/pull/1620), which sends the param. This side is backward compatible — no param means previous behavior — so merging admin-first is safe.

## Changes

- `pages/SpaceAdmin/SpaceEntry.tsx` — read `?spaceId=` from the URL and use it to resolve the navigation target on both the already-logged-in and the fresh-login-via-session-token paths. Both paths go through one `resolveAndSyncTargetSpaceId` helper that also aligns the store before navigating.
- `pages/SpaceAdmin/spaceTarget.ts` (new) — pure `resolveTargetSpaceId(spaces, requestedSpaceId, fallbackSpaceId)`. The requested id is honored only when it is `===`-present in the manageable list (`role >= 1`, sourced from `/v1/space/my`), so unknown, left, or non-admin ids fall through to the default.
- `pages/SpaceAdmin/spaceTarget.test.ts` (new) — 7 cases on the pure helper.
- `pages/SpaceAdmin/SpaceEntry.test.tsx` (new) — 15 cases on the component: both entry branches, URL parsing (empty / absent / duplicate / percent-encoded), store-sync-before-navigate ordering, role-0 filtering, the no-manageable-space logout, and the super-admin bypass.

## Review follow-ups

- **Store sync was missing on the already-logged-in path**: the first pass only synced on fresh login, but the persisted `currentSpaceId` can differ from the resolved target just the same. Without the sync, SpaceAdminLayout's header (which renders outside the loading gate) paints the wrong `SpaceSwitcher` selection for a frame, and its effect — whose dependency array contains `currentSpaceId` — calls `setCurrentSpaceId` on itself, tearing down and re-running, which fires a redundant second `/space/my` + `/space/{id}` pair. Both paths now sync before `navigate()` (`616b257`).
- **The `status` comment overstated what the check does**: `/v1/space/my` filters `status = 1` in SQL, so a frozen space never reaches the list and the branch cannot fire; one frozen after load still carries a stale `status: 1` and passes anyway. The real backstop is SpaceAdminLayout re-fetching on mount and bouncing ids that are no longer manageable. The comment now describes the check as defensive redundancy rather than frozen-state handling (`616b257`).
- **Coverage stopped at the pure helper**: added the 15-case component test above, which covers the branches, the URL parsing, and the sync ordering. Mutation-verified — removing the store sync fails 3 cases, removing the membership check fails 6 — so the coverage is real rather than tautological (`616b257`).
- **Test filename** renamed to match its module, per repo convention (`616b257`).

Deliberately not changed, for the record:

- `nginx.conf.template`'s `/space` short link still drops the query (`return 301 /admin/space`, no `$is_args$args`). The paired web change targets `/admin/space` directly, so the primary path never touches this hop, and the standard deployment's front-door nginx already preserves the query with `return 301 /admin/space$1$is_args$args`. The residual gap only affects hand-shared `/space?spaceId=…` links hitting this container directly, and it fails safe to the old default. Left out to keep this PR frontend-only.
- Validating `fallbackSpaceId` against the same membership predicate. No reachable case produces a persisted id absent from a non-empty `mySpaces` (`setMySpaces` repairs it on every refresh), so this would add a third fallback tier for an unreachable state.

## Testing

- **338/338 tests pass** across 27 files (was 323 before this branch; 22 of them cover this change).
- **`tsc --noEmit` clean.**
- **`vite build` succeeds.**
- **Mutation-verified** as described above.
- End-to-end browser check of `/admin/space?spaceId=<x>` still pending a staging environment.

https://claude.ai/code/session_01Fm1wJX7SoAXUh5cdwPXSes


Jerry-Xin (dismissed):
## Verdict: APPROVE

Reviewed head `ee5f962` with a byte-level reading of all three changed files plus the surrounding authorization flow, and verified locally (install, tests, typecheck, build).

### Authorization — URL-param-driven space selection does NOT bypass the admin check

- `resolveTargetSpaceId` (`src/pages/SpaceAdmin/spaceTarget.ts`) only honors the requested id when it **strictly equals** a `space_id` in the manageable list (`role >= 1`, sourced from `GET /v1/space/my`). Unknown / left / non-admin ids fall back to the existing default. Role-0 spaces never appear in `mySpaces`: both the SpaceEntry token path and SpaceAdminLayout filter `role >= 1`.
- Independently of this entry change, `SpaceAdminLayout` re-fetches `/v1/space/my` on mount and on every `spaceId` change and bounces to `/space` if the URL space is no longer in the managed list (`src/pages/SpaceAdmin/SpaceAdminLayout.tsx`). So even a hand-crafted deep link to a non-managed space cannot land; the membership gate is intact and server-side data APIs are unchanged.

### Parsing / malformed input — fail-closed

- `readRequestedSpaceId()` guards SSR (`typeof window === 'undefined'`), wraps `URLSearchParams` in try/catch, and returns `''` on any failure; `null` is normalized via `|| ''`.
- Arbitrary, malformed, whitespace-only, or oversized values simply fail the strict-equality match against server-provided ids and fall back. The param is consumed only by that equality comparison — it is never reflected into redirects, DOM, or API paths, so there is no open-redirect or param-injection surface. Navigation targets internal routes `/space/<id>/members` with ids taken from the managed list or the pre-existing fallback.

### Frozen-space handling

- `status !== 1` rejects the requested space; `status === undefined` is treated as normal — a documented, deliberate choice matching SpaceAdminLayout's `detail.status === 1` "normal" predicate, avoiding over-rejection when the backend transiently omits the field. Editing on a frozen space remains blocked by `canEdit` in the layout regardless.

### Fresh-login store sync

- `loginSpace` sets `currentSpaceId = managed[0]`; syncing the store to the resolved target before `navigate()` prevents a one-frame wrong selection in SpaceSwitcher. Correct, and the guard `targetSpaceId !== managed[0].space_id` avoids a redundant write.

### Local verification on this head

- New `resolveTargetSpaceId.test.ts`: 7/7 pass (happy / no request / unknown id / empty list / frozen / explicit normal / undefined status).
- `tsc --noEmit`: clean. `vite build`: succeeds.
- Full suite: 303/323 pass. The 20 failures are confined to `src/api/index.test.ts` and `src/pages/Login/useSessionRestore.test.tsx` and reproduce identically at the merge base — pre-existing and unrelated to this change (verified in a separate worktree at the merge base).

### Non-blocking notes

1. Edge case identified in second-pass analysis: when the requested frozen space equals the fallback (`resolveTargetSpaceId(spaces, 'b', 'b')` with `b` frozen), the result is the frozen space `b`. This is behaviorally identical to base — the old code navigated to `currentSpaceId` / `managed[0]` unconditionally — and landing on a frozen space is already reachable today via SpaceSwitcher, which lists all managed spaces without a status filter. Not a blocker; if stricter semantics are wanted later, fall back to the first *normal* manageable space and add a regression test for requested == fallback == frozen.
2. Optional follow-up: component-level coverage for the two SpaceEntry paths (URL parsing + store sync); the pure helper is well covered already.

yujiawei (dismissed):
# Code Review — PR #161 (octo-admin)

Reviewed at head `ee5f962`, merge-base `e601186`. Diff is 3 files / +89 −2, all under `src/pages/SpaceAdmin/`.

**Verdict: Approve.** No blocking (P0/P1) issues. Everything below is P2 / nit — worth a follow-up commit, but nothing that should hold the merge. I also ran a second and third independent pass over the diff and reconciled them; where those passes disagreed with each other I re-read the code and adjudicated, and I've noted the disagreements explicitly so nothing is silently averaged away.

---

## 1. Scope check

The stated goal is: honor `?spaceId=` on admin entry when the user actually manages that space, otherwise keep the old default. The diff does exactly that, on both entry paths:

- `SpaceEntry.tsx:93-101` — already-logged-in path
- `SpaceEntry.tsx:140-151` — fresh-login-via-session-token path

Nothing required is missing, and no unrelated surface is touched (the earlier nginx tweak was reverted in `3caf99a`, so the diff is frontend-only).

Two behaviors go slightly beyond the headline goal, both documented in the PR description:

- rejecting a requested space with `status !== 1` (`spaceTarget.ts:18`)
- syncing `currentSpaceId` before `navigate()` on the fresh-login path (`SpaceEntry.tsx:145-150`)

Both are defensible hardening rather than scope creep, but see §3.1 and §3.2 — the first one does not actually do what its comment claims, and the second one was only applied to one of the two paths.

## 2. Security review

This landed with the security-sensitive label, so I traced the untrusted value end to end. **I found no authorization or injection problem.** Recording the reasoning so it doesn't have to be re-derived:

- **No IDOR.** `resolveTargetSpaceId` (`spaceTarget.ts:17-19`) only accepts the requested id if it is `===`-present in the candidate list. That list is never attacker-influenced: it is populated exclusively from the server's `/v1/space/my` response filtered to `role >= 1` (`SpaceEntry.tsx:118`, `SpaceAdminLayout.tsx:139` and `:168`), and the server scopes that endpoint to the caller's own memberships. A crafted `?spaceId=` for someone else's space falls straight through to the fallback.
- **No path injection / open redirect.** The value is interpolated into `` `/space/${targetSpaceId}/members` ``, but since it must equal a server-issued id verbatim, `../`-style or encoded-separator payloads can never reach the template. There is no host or scheme position in the target, so no open-redirect shape exists.
- **Defense in depth is intact.** Even if a stale in-memory list let a wrong id through, `SpaceAdminLayout` re-fetches `/space/my` on mount and bounces ids that are no longer manageable (`SpaceAdminLayout.tsx:141-144`). Server side, both the detail and member reads independently require membership.
- **No parameter leakage.** The param is read once at entry and the subsequent `navigate(..., { replace: true })` drops it, so it does not linger in history or get forwarded anywhere.
- `readRequestedSpaceId()` (`SpaceEntry.tsx:31-38`) is correctly guarded — SSR-safe and `try`/`catch`-wrapped — and `window.location.search` is the right source here because the app mounts a `BrowserRouter` (`App.tsx:283`). Worth stating because this line would silently read empty under a `HashRouter`; it does not, today.

I re-ran the PR's own gates to confirm the claims: `vitest run` → 26 files / 323 tests pass, and `tsc --noEmit` is clean.

## 3. Findings (all non-blocking)

### 3.1 P2 — the `currentSpaceId` sync is missing on the already-logged-in path

`SpaceEntry.tsx:145-150` syncs the store before `navigate()` specifically to stop `SpaceSwitcher` painting the wrong space for a frame. The identical situation exists on the already-logged-in path at `SpaceEntry.tsx:93-101`, where the persisted `currentSpaceId` can differ from the resolved target, and there the sync was not added.

Consequences with persisted current space `a` and `?spaceId=b`:

1. `SpaceAdminLayout`'s header — including `SpaceSwitcher` — renders outside the `loading` gate (`SpaceAdminLayout.tsx:264`), so the switcher paints `a` while the URL says `b`. The repair happens in a `useEffect` (`SpaceAdminLayout.tsx:131`), i.e. after paint, so this is a real visible frame, not a theoretical one.
2. More concretely: that effect's dependency array includes `currentSpaceId` (`SpaceAdminLayout.tsx:157`). Calling `setCurrentSpaceId(spaceId)` inside it mutates one of its own deps, so the effect tears down (`cancelled = true`) and re-runs — issuing a second `/space/my` + `/space/{id}` pair on every entry where URL and store disagree. This amplification is pre-existing for any direct URL paste, but `?spaceId=` makes it the common case rather than the rare one.

Suggested fix — mirror lines 148-150 on this branch, or fold it into the resolve step. An alternative worth considering for the fresh-login path: pass the resolved target into `loginSpace()` so the store commits once, instead of `loginSpace()` setting `managed[0]` and then immediately being corrected.

### 3.2 P2 — the frozen-space guard doesn't hold in either direction, and its comment overstates what it does

`spaceTarget.ts:18` rejects a requested space whose `status !== 1`. Two independent reasons this provides close to no protection:

- **For a freshly fetched list it is unreachable.** The server's `/v1/space/my` query filters `s.status = 1` in SQL (and the alternate single-space variant used by non-session credentials applies the same condition), so a frozen space is never in the response at all. The branch cannot fire.
- **For a list already in memory it doesn't fire either.** If a space is frozen *after* the list was loaded, the cached entry still carries its old `status: 1`, so the guard passes and the user is navigated there anyway — then bounced by the layout's refresh. That is exactly the case the guard was meant to catch.

The comment at `spaceTarget.ts:12-14` also justifies the check by pointing at `SpaceAdminLayout`'s "`detail.status === 1` means normal" predicate. That predicate only gates *editing* (`SpaceAdminLayout.tsx:94`); the layout renders a frozen space fine, and `SpaceSwitcher` lists every entry of `mySpaces` with no status filter at all (`SpaceSwitcher.tsx:19-38`), so a user could reach a frozen space with one click regardless. Nothing else in the app treats a frozen space as a page that "would error out."

I'd keep the guard as cheap defense-in-depth, but reword the comment to say what it is — a belt-and-braces check against a list shape the API does not currently produce — so the next reader doesn't assume frozen spaces are handled.

### 3.3 P2 — test coverage stops at the pure helper

The 7 cases in `resolveTargetSpaceId.test.ts` all exercise synthetic arrays. Two of them (`:25-33`) assert on `status: 2` / explicit-`status` shapes that `/v1/space/my` cannot return, which reads as more coverage than it is.

What is untested is where the risk actually lives: nothing renders `SpaceEntry`, so neither entry branch, the store/URL synchronization, nor the `readRequestedSpaceId()` parsing is covered. All seven tests pass whether or not the deep link works. `useSessionRestore.test.tsx` in this repo is a good local precedent for testing this kind of effect-driven entry component — one test that mounts `SpaceEntry` with a `?spaceId=` search and asserts the resulting navigation target would cover far more than the remaining helper permutations.

### 3.4 P2 — the `/space` short link still drops the query

`nginx.conf.template:19-21` is unchanged:

```nginx
location = /space {
    return 301 /admin/space;
}
```

`return` does not forward the query string, so `/space?spaceId=X` arrives at the SPA with no param and silently falls back to the old default.

This is **not** blocking, and I want to be explicit about why, because a reading of this diff in isolation suggests otherwise: the paired web change deliberately links to `/admin/space?spaceId=…` rather than the short link, and `/admin/` is served by `try_files` with no redirect hop (`nginx.conf.template:27-29`), so the primary entry path preserves the param. The failure mode is limited to hand-shared or documented `/space?spaceId=X` links, and it fails safe (old behavior, no error).

Still worth the one-token fix — `return 301 /admin/space$is_args$args;` — so the short link honors the param too. The revert in `3caf99a` justified dropping this on the grounds that production nginx doesn't carry this block; that's a claim about infrastructure that can't be checked from this repo, while the template this repo does own has the block. See §4.

### 3.5 Nit — test filename doesn't match the module

The module is `spaceTarget.ts` but the test is `resolveTargetSpaceId.test.ts`. Repo convention is module-named (`sessionRestore.test.ts`, `cooldown.test.ts`, `probeHelpers.test.ts`). Renaming to `spaceTarget.test.ts` keeps it findable.

### 3.6 Nit — the fallback id gets no validation, though the helper already has the list

`SpaceEntry.tsx:93` builds `fallbackSpaceId` from `state.currentSpaceId` (which *is* persisted, while `mySpaces` is not) and passes it through unvalidated, while the requested id is validated against the list. I tried to construct a reachable case where the persisted id is absent from a non-empty `mySpaces` and could not — `setMySpaces` repairs `currentSpaceId` on every refresh — so this is robustness, not a bug. Since the list is already a parameter, running the fallback through the same membership predicate is nearly free.

## 4. Worth a human eye before merge

- **Which nginx config production actually serves.** If any environment serves this repo's `nginx.conf.template`, §3.4 applies to the short link. The consequence is a silent no-op, not an error, which is the kind of thing that gets reported months later as "the feature never worked for me."
- **Landing order with the paired web change.** This side is backward compatible (no param → previous behavior), so merging admin-first is safe. The user-visible improvement only appears once the web side ships.
- **The manual pass the PR description lists as still pending.** Worth actually doing, since no automated test covers the wired-up path: `?spaceId=<manageable>` lands on that space and the switcher shows it immediately; `?spaceId=<not-manageable>` falls back without surfacing anything about the space; a single-space admin is unaffected; and confirm the duplicate `/space/my` + detail fetch from §3.1 in the network panel.

## 5. Reconciliation notes

Two additional independent passes were run over this diff besides my own read; both completed, so nothing here rests on a missing opinion.

**Agreed across passes:** the missing sync on the already-logged-in path (§3.1), the frozen-guard weakness (§3.2), the helper-only test coverage (§3.3), and the `/space` query-strip (§3.4).

**Disagreements, and how I settled them:**

- One pass rated §3.4 blocking, on the reasoning that the PR's primary entry path never reaches the new resolver. I settled it as non-blocking: that pass had no visibility into the paired web change, which routes to `/admin/space` precisely to skip the redirect hop. Concrete evidence outranks the inference.
- One pass reported unsafe array access at `SpaceEntry.tsx:93` and `:143` crashing on an empty list. **Refuted** — line 93 sits behind `state.mySpaces.length > 0` (`:75`) and line 143 behind an early `return` for `managed.length === 0` (`:119-123`). No crash path exists.

**Coverage / blind spots, including what none of the passes covered:**

- One pass flagged that `window.location.search` would break under a `HashRouter`. Closed: the app uses `BrowserRouter` (`App.tsx:283`).
- One pass could not verify server-side authorization on the space detail and member endpoints. Partially closed: both require membership before returning anything, and because the param must match a server-issued id exactly, delimiter-bearing ids are a non-issue for the redirect. Residual, and pre-existing / outside this diff: I did not audit the role gate on every admin-tab write endpoint.
- Not covered by any pass: no browser-level verification exists at all (acknowledged in the PR description); the no-op case where the requested space already is the current one; and the fact that the param is lost if a user lands on the no-token branch and detours through login.

yujiawei (approved):
# Code Review — PR #161 (octo-admin)

Reviewed at head `616b257`, merge-base `e601186`. Diff is 4 files, +314 / −2, all under `src/pages/SpaceAdmin/` (240 of those lines are new tests).

**Verdict: Approve.** No P0/P1. Everything below is P2 or a nit.

This is a re-review: the two earlier reviews were dismissed at `ee5f962`. I re-derived the verdict against this head rather than inheriting the previous one, and I checked which of the earlier P2s were actually addressed (four of six were — see §4).

Verified locally at this SHA, not taken from the PR description:

- `pnpm install --frozen-lockfile` → clean
- `pnpm exec tsc --noEmit` → clean
- `pnpm test` → **27 files / 338 tests pass**, no `act()` warnings
  (the PR description still says "26 files / 323 tests" — stale, see §3.5)

---

## 1. Scope check

Stated goal: honor `?spaceId=` on admin entry when the user actually manages that space, otherwise keep the previous default.

The diff does exactly that, on both entry paths, and nothing else:

- `SpaceEntry.tsx:115-123` — already-logged-in path
- `SpaceEntry.tsx:161-167` — fresh-login-via-session-token path
- `spaceTarget.ts:15-25` — the pure resolver

Nothing required is missing. No unrelated surface is touched: no new dependency, no new API call, no new i18n key, no feature flag, and no deployment config (the earlier `nginx.conf.template` tweak was reverted in `3caf99a`, so this diff is frontend-only). `readRequestedSpaceId()` is called at `SpaceEntry.tsx:91`, *before* the super-admin branch at `:93`, and the super branch ignores it — so the param cannot pull a super-admin session into the space console. That is covered by a test.

## 2. Security review

This carries the security-sensitive label, so I traced the untrusted value end to end. **I found no authorization, IDOR, path-injection, or open-redirect problem.** Recording the reasoning so it does not have to be re-derived next round:

**The gate is strict equality against a server-supplied list.** `spaceTarget.ts:21-22`:

```ts
const match = spaces.find((s) => s.space_id === requestedSpaceId)
if (!match) return fallbackSpaceId
```

The candidate list is never attacker-influenced. I checked **every** write site of `mySpaces` in the repo, not just the ones in the diff:

- `SpaceEntry.tsx:161` → `loginSpace(token, resolvedUid, name, managed)` where `managed` is `filter((s) => s.role >= 1)` at `:140`
- `SpaceAdminLayout.tsx:140` and `:169` → `setMySpaces(managed)`, both `filter((s) => s.role >= 1)` at `:139` / `:168`

Those are the only three. `mySpaces` is also **excluded from persistence** (`store/auth.ts:119-127` `partialize`), so it cannot be poisoned by editing `localStorage`.

**No path injection.** The resolved id is interpolated into `` `/space/${targetSpaceId}/members` ``, but because it must equal a server-issued id verbatim, `../` or encoded-separator payloads can never reach the template. There is no host or scheme position in the target, so no open-redirect shape exists.

**Defense in depth is intact.** Even if a stale in-memory list let a wrong id through, `SpaceAdminLayout.tsx:128-145` re-fetches `/space/my` on mount and bounces ids no longer in the managed list back to `/space`. Server side, the detail and member reads independently require membership.

**No parameter leakage.** The param is read once, and the subsequent `navigate(..., { replace: true })` drops it, so it does not linger in history or get forwarded.

`readRequestedSpaceId()` (`SpaceEntry.tsx:31-38`) is SSR-guarded and `try`/`catch`-wrapped, and `window.location.search` is the correct source because the app mounts a `BrowserRouter` (`App.tsx:283`) — worth stating explicitly, since this line would silently read empty under a `HashRouter`. `URLSearchParams.get` takes the first occurrence, so `?spaceId=<evil>&spaceId=<valid>` resolves to the invalid first value and falls back rather than being overridden; there is a test for that.

## 3. Findings (all non-blocking)

### 3.1 P2 — the resolver's stated defense-in-depth guards the wrong dimension

`spaceTarget.ts:13` says the `status` check is kept "to guard against a future caller passing an unfiltered list". But the parameter type on `:16` is:

```ts
spaces: Pick<MySpace, 'space_id' | 'status'>[],
```

`role` is structurally excluded, so the resolver **cannot** reject a `role: 0` entry — which is the dimension that would actually matter for an unfiltered list. What it does check (`status !== 1`, `:23`) is a dimension the backend already filters in SQL, as the comment itself concedes two lines earlier. So the function guards the harmless axis and is structurally unable to guard the load-bearing one.

Not exploitable today — §2 shows all three write sites filter `role >= 1` and the list is not persisted — but "manageable" is a caller convention with nothing enforcing it. Suggested fix: widen the type to `Pick<MySpace, 'space_id' | 'status' | 'role'>` and add `if (match.role !== undefined && match.role < 1) return fallbackSpaceId`, or normalize at the store write boundary in `setMySpaces`/`loginSpace` so the invariant lives in one place. Either makes the comment true.

### 3.2 P2 — the `/space` short link still drops the query

`nginx.conf.template:19-21` is unchanged:

```nginx
location = /space {
    return 301 /admin/space;
}
```

`return` does not forward the query string, so `/space?spaceId=X` reaches the SPA with no param and silently falls back to the old default.

I want to be explicit about why this is **not** blocking, because reading this diff in isolation suggests otherwise: the paired web change deliberately links to `/admin/space?spaceId=…`, and `/admin/` is served by `try_files` with no redirect hop (`nginx.conf.template:27-29`), so the primary entry path preserves the param. The failure mode is limited to hand-shared or documented `/space?spaceId=X` links, and it fails safe (previous behavior, no error).

Still worth the one-token fix — `return 301 /admin/space$is_args$args;` — since the comment on `:18` advertises that short link as shareable. One correction to the revert commit's rationale, because it will mislead the next reader: `3caf99a` justified dropping this on the grounds that "production nginx does not carry the `location = /space` return-301 block that this template shows." The block **is** in this repo's template, and `Dockerfile:10` copies that template into the image as `/etc/nginx/templates/default.conf.template`. The revert's *conclusion* is fine (the web side bypasses the hop), but its *stated reason* is not, and it is the reason a future maintainer will trust.

### 3.3 P2 — ids are interpolated into the route without encoding

`SpaceEntry.tsx:121` and `:167` build `` `/space/${targetSpaceId}/members` `` with no `encodeURIComponent`. Safe today because the value is a verbatim server-issued id, and consistent with the pre-existing `SpaceSwitcher.tsx:44`. But `SpaceEntry.test.tsx:194-197` deliberately exercises an id containing a space (`'a b'`) and asserts the raw `'/space/a b/members'`, which shows the code path does carry unencoded characters. Encoding both call sites (and `SpaceSwitcher`) costs nothing and removes the dependency on an unstated backend guarantee about id charset. A test with an id containing `/` would be the useful one to add.

### 3.4 P2 — the fallback id is never validated, though the list is right there

`SpaceEntry.tsx:115` builds `fallbackSpaceId` from `state.currentSpaceId` (which **is** persisted, while `mySpaces` is not) and passes it through unvalidated, while the requested id is validated against the list. I tried to construct a reachable case where the persisted id is absent from a non-empty `mySpaces` and could not — `setMySpaces` (`store/auth.ts:88-95`) reconciles `currentSpaceId` on every refresh. So this is robustness, not a bug. Since the list is already a parameter, running the fallback through the same predicate is nearly free. Carried over from the previous round, still unaddressed, still non-blocking.

### 3.5 Nit — the PR description no longer matches the head

It describes commit `b8f89e4`: it names the test file `resolveTargetSpaceId.test.ts` (renamed to `spaceTarget.test.ts`), does not mention `SpaceEntry.test.tsx` at all (204 lines, 15 tests), says the store sync applies to the fresh-login path only (it now applies to both), and reports "26 files / 323 tests" (actual: 27 / 338). Worth a refresh before merge so the description is a usable record.

### 3.6 Nit — test teardown assumes setup succeeded

`SpaceEntry.test.tsx:70-73` calls `root.unmount()` and `container.remove()` unconditionally. If `beforeEach` throws before `:66-68` assign them, teardown throws on a stale/undefined value and masks the real failure. Guarding with `root?.` / `container?.` keeps failures legible.

## 4. What the previous round asked for, and what landed

Recording this so the round count is legible rather than re-litigated:

| Previous P2 | Status at `616b257` |
|---|---|
| `currentSpaceId` sync missing on the already-logged-in path | **Fixed** — `SpaceEntry.tsx:53-64`, applied on both paths |
| Frozen-space comment overstated what the guard does | **Fixed** — `spaceTarget.ts:9-13` now states plainly that the branch is unreachable and that the real backstop is the layout refresh. Unusually honest comment; keep it. |
| Coverage stopped at the pure helper | **Fixed** — `SpaceEntry.test.tsx`, 15 component-level tests |
| Test filename did not match the module | **Fixed** — renamed to `spaceTarget.test.ts` |
| `/space` short link drops the query | Still open → §3.2 |
| Fallback id unvalidated | Still open → §3.4 |

**The store sync is a real fix, not cosmetic.** I verified both consequences the comment at `SpaceEntry.tsx:44-52` claims: `SpaceAdminLayout`'s `<Header>` with `<SpaceSwitcher />` (`:264`) renders outside the `loading || !detail` gate (`:332`), so a wrong selection is a genuinely painted frame; and that effect's dep array includes `currentSpaceId` (`:157`) while its body calls `setCurrentSpaceId(spaceId)` (`:131`), so it mutates its own dependency, tears down, and re-runs — issuing a second `/space/my` + `/space/{id}` pair. Pre-aligning the store removes that duplicate pair on every deep-linked entry. Both claims check out.

**The new tests are load-bearing, not decorative.** I did not take "338 tests pass" at face value — I ran four mutations against this head and confirmed each is caught:

| Mutation | Result |
|---|---|
| resolver always returns `fallbackSpaceId` (feature disabled) | **8 tests fail** |
| drop the `status !== 1` check | **1 fails** |
| drop the membership check (`!match` returns the requested id) | **5 fails** |
| drop the store sync before `navigate` | **3 fails** |

Worth noting because the fourth one is the interesting case: the two "store synced *before* navigate" tests capture `useAuthStore.getState().currentSpaceId` from inside `navigate.mockImplementation` rather than reading it after the fact, so they actually pin the ordering instead of just the end state. That is the right way to test this, and it is why the mutation is caught.

## 5. Additional findings outside this diff

Surfaced while reading the surrounding flow. **Not attributable to this PR and not part of the verdict** — filing so they are not lost:

- **A valid persisted token can still dead-end at "no token."** `token` is persisted (`store/auth.ts:121`) but `mySpaces` is not (`:119-127`). On a hard load of `/admin/space` in a tab with no `sessionStorage` (fresh tab, bookmark, browser restart), `mySpaces` is empty so the branch at `SpaceEntry.tsx:97` is skipped, and `:126` consults `sessionStorage` only — so `readSessionToken()` returns `''` and the user gets the `no-token` result despite holding a valid persisted token. The store's own comment on `:118` says `mySpaces` is re-fetched by `SpaceEntry` on every start; on this path it is not. `state.token || readSessionToken()` would close it. Pre-existing, untouched here, and PR #143 is rewriting this exact branch.
- **The `no-token` exit sends space admins to the super-admin sign-in.** `SpaceEntry.tsx:199` routes to `/login`, which is the manager login and lands on `/dashboard` (`Login/index.tsx:206`, `:257`). Also pre-existing, and also what #143 addresses.

## 6. Worth a human eye before merge

1. **Merge order against PRs #143 and #144.** Both are open, both modify `src/pages/SpaceAdmin/SpaceEntry.tsx`, and both **add their own `src/pages/SpaceAdmin/SpaceEntry.test.tsx`** — an add/add conflict with the file this PR creates. More importantly, #143 refactors the fresh-login path into an `enterWithToken()` helper that ends with a hardcoded `` navigate(`/space/${managed[0].space_id}/members`) `` — i.e. it reintroduces the exact `managed[0]` behavior this PR removes. All three are independently `MERGEABLE`, so GitHub will not warn; git will raise the conflict, but resolving it by taking #143's rewrite would silently revert this feature with no failing test to catch it (#143 replaces the test file too). Whoever merges second needs to reconcile deliberately.
2. **The manual pass the description lists as still pending.** No automated test covers the wired-up path end to end. Worth actually doing: `?spaceId=<manageable>` lands on that space and the switcher shows it on the first frame; `?spaceId=<not-manageable>` falls back without revealing anything about that space; a single-space admin is unaffected; and confirm in the network panel that the duplicate `/space/my` + detail pair described in §4 is now gone.
3. **Landing order with the paired web change.** This side is backward compatible (no param → previous behavior), so merging admin-first is safe; the user-visible improvement appears only once the web side ships.
4. **Which nginx config production actually serves** (§3.2). The consequence of getting this wrong is a silent no-op on the short link, not an error — the kind of thing reported months later as "that never worked for me."

## 7. Reconciliation notes

Two additional independent passes were run over this diff besides my own read. Both completed, so nothing here rests on a missing opinion. Severities were normalized to a single scale before deduplication.

**Agreed across passes:** the `role`-not-enforceable resolver contract (§3.1) and the `/space` query strip (§3.2). Both were raised independently by more than one pass, which is why they lead the list.

**Disagreements, and how I settled them:**

- One pass rated §3.1 blocking (P1), on the reading that a `role: 0` space could reach the resolver and let a non-admin into a space console via `?spaceId=`. **Downgraded to P2.** I enumerated all three `mySpaces` write sites (`SpaceEntry.tsx:161`, `SpaceAdminLayout.tsx:140`, `:169`) — every one filters `role >= 1` — and confirmed `mySpaces` is excluded from `partialize`, so it cannot be seeded from `localStorage`. No reachable path today; the finding is a real contract gap, not a live hole.
- One pass reported a P1 crash at `SpaceEntry.tsx:115` (`state.mySpaces[0].space_id` on an empty array). **Refuted.** That line sits inside the branch guarded by `state.mySpaces.length > 0` at `:97`, and the corresponding fresh-login access at `:165` sits behind the `managed.length === 0` early return at `:141-145`. No crash path exists. The line is also unchanged by this diff. Noting it because the previous review round refuted the identical false positive — it is a recurring artifact on this file, not a regression.
- One pass called the empty-list case at `spaceTarget.test.ts:19` a vacuous assertion. **Refuted with evidence:** mutation 3 in §4 (dropping the membership check) fails that exact test. It is load-bearing.
- One pass wanted the already-logged-in path to re-fetch `/space/my` before honoring the param, since `state.mySpaces` is an in-memory snapshot that a demotion can invalidate. **No change requested.** The same pass concedes it is not an independent IDOR: `SpaceAdminLayout` revalidates before rendering (`:128-145`) and the server authorizes the detail read regardless. The pre-existing `currentSpaceId` fallback already trusted that same snapshot, so the param widens the choice only within an already-stale set of the user's *own* spaces. An extra blocking fetch on every entry is a worse trade than one bounced request in a rare window.

**Coverage / blind spots, including what no pass covered:**

- One pass flagged that `window.location.search` would break under a `HashRouter`. Closed: `App.tsx:283` mounts `BrowserRouter`.
- One pass could not run the test suite (read-only checkout blocked Vite's temp dir) and reasoned about test quality by inspection only. Closed by the mutation results in §4, which are empirical.
- Neither pass had visibility into the paired web change; I read it directly to settle §3.2 (it targets `/admin/space?spaceId=…` and `encodeURIComponent`s the id).
- Not verified by anyone, mine included: real-browser behavior (acknowledged as pending in the description); backend id-charset guarantees, which §3.3 currently leans on; `/v1/space/my` freshness semantics; and the role gate on every admin-tab **write** endpoint, which is pre-existing and outside this diff.
- `SpaceEntry.test.tsx` mocks `react-router-dom`, `antd`, and the api module wholesale (`:11-33`). That is a defensible choice for an effect-driven entry component and the mutations show it still pins the real behavior — but it structurally cannot catch basename handling, path-segment escaping, or redirect loops. If one integration test with a real `MemoryRouter` is ever added, the id-containing-`/` case from §3.3 is the one worth spending it on.

Jerry-Xin (approved):
Re-review of delta `ee5f9629..616b257b` (1 commit: sync store on both entry paths, add SpaceEntry coverage).

**Prior 🟡 items — status:**
1. Logged-in path missing currentSpaceId sync → addressed. New `resolveAndSyncTargetSpaceId` syncs the store before navigate on BOTH entry paths (previously only the fresh-login path synced, and only conditionally). New tests assert store state at the moment of navigate.
2. spaceTarget comment/behavior mismatch → addressed. The comment now describes the status check as defensive redundancy rather than a frozen-space gate (logic unchanged — verified comment-only diff).
3. Frozen fallback when requested == current == frozen → addressed via documentation. The behavior is byte-identical to pre-PR entry (landing on persisted currentSpaceId) and is also reachable via SpaceSwitcher, so it remains non-blocking.

**Security invariants re-verified at byte level:**
- `?spaceId=` is honored only when it strictly equals a space_id in the manageable list (role>=1 from /v1/space/my); unknown/non-admin/malformed values fail closed to the default fallback. `resolveTargetSpaceId` logic is unchanged from the approved head.
- SpaceAdminLayout is untouched by this delta and still independently re-fetches /space/my on mount/spaceId change, bouncing no-longer-manageable ids back to /space.
- The param is consumed only by equality comparison; navigate targets are always server-provided ids (a matched list entry, or the existing currentSpaceId / managed[0]) — never the raw param, so no injection or open-redirect surface.
- Malformed param: URLSearchParams read is try/caught to '' → fallback; no crash surface.

**Writer census for the new sync path:** `setCurrentSpaceId` writers = SpaceSwitcher (user action), SpaceAdminLayout effect (pre-existing), and the new helper. The helper only writes the resolved target — a server-provided list entry or the existing store value — never raw URL input; both call sites pass role>=1-filtered lists. Pre-syncing also keeps the layout effect's dependency stable (spaceId === currentSpaceId on mount → no redundant re-run).

**Local verification:** `tsc --noEmit` clean, `vite build` clean, full suite 338/338 pass including the 15 new SpaceEntry tests; CI also green on Node 18/20.

Codex raw verdict REQUEST_CHANGES overridden: its sole blocker (requested == frozen == current resolves to the frozen space) is pre-existing, byte-identical base behavior already reachable without this param, adjudicated non-blocking on the prior head.

PR #163, opened 2026-09-11
refactor(i18n): rename Space to Organization across admin copy

an9xyz:
## What

Unify the product term for the top-level tenant: **空间 / Space → 组织 / Organization**, in both locales.

One screen used to show three names for the same concept — the sidebar said 「空间管理」, the page title said 「Space 管理」, and the column header said 「Space 名称」.

| | Before | After |
|---|---|---|
| Sidebar / breadcrumb (zh) | 空间管理 | 组织管理 |
| Page title (zh) | Space 管理 | 组织管理 |
| Column header (zh) | Space 名称 | 组织名称 |
| Sidebar (en) | Spaces | Organizations |
| Page title (en) | Space Management | Organizations |
| Space admin header (en) | Space Admin | Organization Admin |
| Dashboard chart (zh) | Top Space 消息量 | Top 组织消息量 |

129 string replacements across 10 namespaces × 2 locales: `spaces`, `spaceAdmin`, `nav`, `common`, `dashboard`, `appBots`, `expertMarket`, `systemMcp`, `systemSetting`, `systemSkill`.

## What is deliberately NOT renamed

Identifiers that map 1:1 to backend fields keep the English term, so support can still match them against `space_id`:

- `spaces.column.spaceId` = `Space ID`
- `appBots.detail.scope.space` = `Space: {{spaceId}}`

Routes (`/spaces`), API paths (`/v1/space/*`), request/response fields (`space_id`, `max_users`) and code comments are untouched. Two false positives were skipped where "space" means whitespace, not the product: `systemMcp.form.argsHint` ("Space-separated arguments") and `groups.remove.field.extra` ("commas or spaces").

**Only i18n JSON values changed — no keys added, removed or renamed**, so no `t()` call site needs updating. Verified by diffing the key set of every touched file against `HEAD`.

## Copy fixes found while auditing the same strings

- **Developer jargon in the UI** — the add-member hint read 「此接口绕过 max_users 限制」, exposing a backend field name to end users. Now 「管理员添加不受成员上限限制」.
- **Synonyms** — 人数上限 vs 成员上限, 空间描述 vs 简介. Unified to 成员上限 / 简介.
- **Missing subject** — ban and unban confirmations named the object, dissolve did not. Now consistent.
- **Punctuation** — `invites.copyLink.error` used a half-width comma where the rest of the app uses full-width.
- **en title vs nav** — nav said `Spaces`, the page title said `Space Management`, while Users/Groups matched in both places. Now both `Organizations`.
- **zh/en divergence** — the dashboard direct-chat empty state described two different things in the two locales; the en hint was also missing a clause the zh one had.

## Guard

`src/i18n/locales/terminology.test.ts` scans every locale value and reports the offending `file#key` on a miss, following the existing `appBots.test.ts` convention.

The guard is itself guarded, because a "scan everything and expect no hits" test fails open:

- the glob is asserted to have loaded >10 bundles per locale, so a wrong path cannot make the scan pass vacuously
- the allowlist is checked for stale entries, so it cannot rot into a place where real misses hide

Both were verified by injection: breaking the glob path makes the two scan assertions still pass (✓) while the new guards fail — which is exactly the failure mode they exist to catch.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm test` — 33 files / 369 tests pass (367 before, +2 guards)
- [x] Interpolation integrity: `{{var}}` sets compared per key against `HEAD` — 129 changes, zero broken placeholders
- [x] zh/en key parity across all 17 file pairs
- [x] Guard verified in both directions (catches a reverted string; catches its own broken glob and stale allowlist)
- [ ] Visual check of the en `Organization Admin` header on a narrow viewport — see below

## Known follow-ups (not in this PR)

1. **Cross-repo terminology drift.** This only covers octo-admin. octo-web's 「空间管理」 entry point and any copy served by octo-server still say 空间, so users may see both terms until those are aligned.
2. **`Organization Admin` is ~55px wider than `Space Admin`** and `SpaceAdminLayout.tsx:261` renders it in a bare `span` with no `whiteSpace: nowrap`. Not observed breaking, but untested on narrow viewports. A one-line defensive fix is available if wanted.
3. **English plural** — `"{{count}} organizations"` reads "1 organizations". Pre-existing (was "1 Spaces"); the call site passes a formatted *string* as `count`, and i18next skips plural resolution for string counts, so fixing it means changing `Dashboard/index.tsx:1833` to pass a number.

No screenshots: this is a pure copy change with no layout or component changes, so the before/after table above carries the same information more precisely.


Jerry-Xin (approved):
## Review: refactor(i18n): rename Space to Organization across admin copy

**Verdict: APPROVE** ✅

Independently verified at commit `3c5bb49` against the full diff — every claim below was re-checked at byte level on a clean checkout, and the test suite was executed on both this head and the merge base.

### Verification

- **Scope discipline** — all 21 changed files live under `src/i18n/locales/` (20 locale JSONs + 1 new guard test). No code identifiers, routes, API paths, or backend field names were renamed: `space_id`, `/v1/space/*`, the `/spaces` route, and component names (`Spaces`, `SpaceAdminLayout`) are untouched.
- **Key stability** — flattened key sets of all 20 changed JSON files are identical to the merge base. No keys added, removed, or renamed, so no `t()` call site can dangle.
- **Locale parity** — zh-CN/en-US key parity holds across all 17 file pairs, not just the 10 touched namespaces.
- **Interpolation integrity** — exactly 129 values changed; the `{{var}}` placeholder set per key is identical before/after. Zero broken interpolations.
- **Rename completeness** — independent re-scan of every locale value at head: zero occurrences of 空间 / `space` (zh) or `Space(s)` (en) outside the 7-entry allowlist. Each allowlist entry is legitimate: `Space ID` and `Space: {{spaceId}}` (×2 locales) are deliberately preserved so support can match them against the backend `space_id` field, and the three "commas or spaces" / "Space-separated" strings use "space" in the whitespace sense. No stale allowlist entries.
- **No half-renamed copy outside i18n** — remaining 空间/Space occurrences in `src/` are code comments (explicitly out of scope) and test-case description labels; no user-visible hardcoded literals.
- **Guard test** — `terminology.test.ts` passes 5/5. The design is sound: it guards against its own vacuous pass (bundle-count assertion per locale) and against allowlist rot (stale-entry check), and pins the renamed nav values with a positive assertion.
- **Type check** — `tsc --noEmit` clean.
- **Test suite** — head: 369 tests / 33 files, exactly +5 tests / +1 file vs the merge base (the new guard). 35 failures in my environment (localStorage undefined under Node v26 hitting `zustand/persist`) reproduce **identically** at the merge base — same 3 files, same counts, same `setItem` TypeError — so they are environmental, not regressions; zero test-behavior delta attributable to this PR. On a supported Node runtime the suite is green as claimed. (Nit: the PR body's "367 before, +2 guards" doesn't match the measured base of 364 / +5 — description drift only, nothing functional.)

### 💬 Non-blocking

- 🟡 The stale-allowlist guard only checks that an exempted key still *exists*, not that its *value* still deserves the exemption — a future edit could turn an allowlisted key into a hiding place for a regression. Consider also asserting allowlisted values still match their documented legacy usage.
- 🔵 The bundle-count assertion (`>10` per locale) cannot prove *every* bundle loaded; comparing against an explicit expected file set would make the vacuous-pass guard airtight.
- 🔵 Test-case descriptions (e.g., in the SpaceEntry tests) still use 空间. Developer-facing only and out of the guard's scope — worth folding into a future terminology sweep.
- The PR's own listed follow-ups (cross-repo drift with octo-web/octo-server copy, `Organization Admin` header width on narrow viewports, the pre-existing "1 organizations" English plural) are correctly scoped out and acknowledged here as known.

The bundled copy fixes were each verified in the diff and improve consistency: the add-member hint no longer exposes the `max_users` backend field to end users, 成员上限/简介 synonyms are unified, the half-width comma in `invites.copyLink.error` is fixed, the dissolve confirmation now names its subject like ban/unban do, and the zh/en direct-chat empty states finally describe the same thing.

Clean, complete, and well-guarded terminology migration. Approving.

yujiawei (approved):
# Code Review — PR #163 (octo-admin)

Reviewed at head SHA `3c5bb494ce55e4b352e68c6fd40703155ec9c895`, against merge-base `ba195c2`.

I re-derived every claim in the PR description from the code rather than taking it on trust. All of them hold. Details below, then five non-blocking suggestions.

## 1. Scope compliance

**Result: passes.**

| Claim | How I checked | Result |
|---|---|---|
| Only JSON *values* changed; no keys added/removed/renamed | Flattened every touched bundle at merge-base and at head, diffed the key sets | 0 added, 0 removed, 0 renamed |
| `{{interpolation}}` preserved per key | Compared the sorted placeholder set of every key on both sides | 0 differences across 129 changed values |
| zh/en key parity | Flattened all 17 file pairs | Parity holds |
| 129 replacements | Counted changed value lines under `src/i18n/locales/*/*.json` | 129 |
| Deliberate non-renames intact | `spaces.json#column.spaceId` = `Space ID`, `appBots.json#detail.scope.space` = `Space: {{spaceId}}`, both locales | Unchanged |
| Routes / API paths / request fields untouched | Diff is 21 files, all under `src/i18n/locales/` | Confirmed |
| `npx tsc --noEmit` clean | Ran it | Exit 0 |
| `npm test` 33 files / 369 tests | Ran `npx vitest run` | 33 files, 369 tests, all pass |

The seven "copy fixes found while auditing" are outside a strict rename but are declared in the PR body, touch the same strings in the same files, and each one is defensible. I don't count them as undeclared scope creep. `members.addModal.field.extra` dropping `max_users` from user-facing copy is a genuine improvement.

I also checked the two things that would make this kind of rename dangerous, and neither is present:

- **No code compares against a translated value.** No `=== t(...)`, `switch (t(...))`, or `includes(t(...))` anywhere in `src/`. The renamed strings are display-only.
- **No user-visible copy outside the bundles still says the old term.** `grep` for `空间` and `\bSpaces?\b` across `src/` leaves only: code comments, route literals (`/space`, `/spaces`), `space_id` / `X-Space-Id` identifiers, Ant Design's `<Space>` component, and test descriptions. Nothing rendered.

## 2. Guard test

`src/i18n/locales/terminology.test.ts` is unusually well built for this class of test. "Scan everything and expect zero hits" normally fails open, and the two meta-assertions (glob loaded, allowlist not stale) close the obvious holes. I verified the allowlist is exactly right by replicating the scan independently: across all 34 bundles there are exactly 7 matches, and all 7 are in `ALLOWED` — no stale entries, no missing ones.

One hole remains, plus a soundness gap in the stale check — see P2-1 and P2-2.

## 3. Findings

No P0 or P1. Everything below is a suggestion; none should block merge.

**P2-1 — `offenders()` has no positive control, so its language filter can go vacuous undetected.**

`terminology.test.ts:38` filters by a second, independent string literal:

```ts
if (!file.startsWith(`${lang}/`)) continue
```

The glob guard at `:52-53` asserts against a *different* literal (`'./zh-CN/'`). If those two ever diverge — a rename to `zh_CN`, a locale directory move — `offenders()` returns `[]` for every input, `:66` and `:70` pass vacuously, and nothing else catches it: the stale-allowlist test at `:56-63` builds its key set directly and never calls `offenders()`. (A broken `entries()` *is* caught, by that same test. The `lang` filter is the one path with no backstop.) One line closes it:

```ts
expect(offenders('zh-CN', () => true).length).toBeGreaterThan(100)
```

**P2-2 — the stale-allowlist check verifies the key exists, not that the exemption is still needed.**

`terminology.test.ts:62` only asks whether each allowlisted id is still a real key:

```ts
expect([...ALLOWED].filter((id) => !existing.has(id))).toEqual([])
```

If an exempted value is later rewritten so it no longer contains the legacy term, the exemption survives as a permanent blind spot at a live key — a future regression there would be pre-authorized. Tightening it to also require that the value still trips `isLegacy` would make the allowlist self-pruning.

**P2-3 — the zh matcher is unbounded where the en one is not.**

`terminology.test.ts:66`:

```ts
offenders('zh-CN', (v) => v.includes('空间') || /space/i.test(v))
```

`/space/i` has no word boundary, so it fires inside `workspace`, `namespace`, `whitespace`, `SpaceX`. That's not hypothetical for this codebase: the en allowlist already carries two whitespace-sense exemptions (`groups.json#remove.field.extra`, `systemMcp.json#form.argsHint`), so the first zh string that mentions whitespace in English hits it. Matching `:70`'s `/\bspaces?\b/i` would fix the English half. The blanket `空间` ban has the same shape — `磁盘空间`, `命名空间`, `内存空间` are all legitimate — though I read that one as deliberate: forcing a conscious allowlist entry is the point, and the escape hatch exists. Worth a comment saying so.

**P2-4 — zh now uses one name for two different consoles, where en distinguishes them.**

After this PR, zh renders 「组织管理」 in three places: `nav.json#spaces` (super-admin sidebar), `spaces.json#title` (super-admin page), and `spaceAdmin.json#header.title` (the org-owner console). The first two matching is the fix this PR set out to make. The third is a different product surface, and en keeps it distinct — `Organizations` vs `Organization Admin`. Low impact, since nobody sees both chromes at once, but the PR's own framing ("one screen showed three names for the same concept") argues for the mirror case too. 「组织管理台」 would restore the distinction.

**P2-5 — three code comments now describe a column label that no longer exists.**

`src/hooks/useSpaceNameMap.ts:7`, `src/api/mcp.ts:114`, and `src/api/expert.ts:83` all reference the admin list 「所属空间」 column. This PR renamed that string to 「所属组织」 (`zh-CN/common.json#table.space`). Pure doc drift, but it's the kind that makes the next `grep` for the label come back empty.

## 4. What I did not verify

Stating this explicitly, because two of the author's own follow-ups live here and I did not close either:

- **No visual check.** en `Organization Admin` is ~55px wider than `Space Admin` and renders in a bare `<span>` at `src/pages/SpaceAdmin/SpaceAdminLayout.tsx:260-262` with no `whiteSpace: 'nowrap'`, inside a flex row that also holds the logo and `<SpaceSwitcher />`. Unverified at narrow viewports — not refuted either. The author's offered one-line defensive fix seems cheap enough to just take.
- **Cross-repo drift.** octo-web's entry point and any copy served by octo-server still say 空间. I did not check those repos. Users will see both terms until they're aligned; this is worth a tracking issue rather than a blocker on this PR.
- **Backend-supplied labels.** `src/pages/SystemSetting/index.tsx:208` resolves category titles via `categoryTitle(t, key)` with a `category.generic` = `{{category}} Settings` fallback, implying the server sends keys. If it can ever send a display label instead, that path bypasses the rename. I did not trace the server contract.
- **Plural.** `kpi.scope.filtered` receives a numeric `count` (`Dashboard/index.tsx:783`) and no `_one`/`_other` variants exist, so it falls back to the base key and renders "1 organizations selected". Pre-existing — it said "1 spaces selected" before — and unchanged by this PR. Not a regression; noting it because the author flagged the sibling case at `Dashboard/index.tsx:1833`.

## 5. Verdict

**Approved.** The risky parts of a 129-string rename are key drift, placeholder damage, and renaming an identifier that code depends on. All three are clean and independently verified, the type check and full suite pass, and the guard test is genuinely load-bearing rather than decorative. The five findings above are all improvements to a change that is already correct.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #163 (octo-admin)

## Summary

This PR renames the product term "Space / 空间" to "Organization / 组织" across the admin console's user-facing copy: 20 locale JSON files (10 namespaces × en-US/zh-CN, 208 insertions / 129 deletions) plus a new guard test `src/i18n/locales/terminology.test.ts` that fails CI if the legacy term creeps back. No component, API, route, store, or type is touched — every edit is a JSON *value*; no key is added, removed, or renamed.

The rename itself is semantically clean. "Space" carries exactly one meaning in this product (the tenant container), so unlike the earlier Agent→Bot pass in this repo there is no second concept that the blanket replacement could contradict, and nothing in `src/api` or `src/store` already uses "organization/org" as a distinct entity. The three P2 items below are all one theme — the two strings the guard deliberately exempts — and none of them block.

## Verification

Static analysis only at head `3c5bb494ce55e4b352e68c6fd40703155ec9c895`; build and tests not executed in this environment (`node_modules` is absent). First review round for this head — no prior unresolved blockers on record.

- ✅ **No locale key was renamed** — flattened key sets of all 20 changed JSON files are identical at `pr-base` vs head, so every existing `t('…')` call site still resolves. This matters because bundles are statically registered in `src/i18n/index.ts:49`: a broken key would not 404, it would render the raw key path to users.
- ✅ **No interpolation drift** — for every changed value the `{{…}}` set is unchanged (`{{count}}`, `{{name}}`, `{{max}}`, `{{cap}}`, `{{range}}`, `{{spaceId}}`, `{{current}}`), so nothing starts rendering a literal `{{count}}`.
- ✅ **en-US ↔ zh-CN parity holds** — all 17 namespaces have identical key sets and identical per-key placeholder sets at head, and both locales were renamed in lockstep (`spaces.json` gains the new term on 21 lines per side).
- ✅ **No runtime dependency on the old values** — no code compares against a translated literal; menu identity uses route paths (`src/layouts/MainLayout.tsx:118` → `key: '/spaces'`, label from `t('nav:spaces')`), and nothing persists a label (`localStorage` holds only `octo-theme`, the `i18n_lang` cookie only the language code). i18next `count` pluralization is unaffected because these keys carry no `_one`/`_other` variants, before or after.
- ✅ **No existing test asserts on the renamed copy** — there are no snapshot files, and the rendered-text assertions under `src/pages/**` target i18n *keys* or Changelog data, so `npm test` does not break.
- ✅ **The new guard is enforced, not decorative** — `vite.config.ts:23` includes `src/**/*.test.{ts,tsx}` and `.github/workflows/ci.yml:90` runs `npx tsc --noEmit`, `npm test`, `npm run build`. `import.meta.glob` type-checks because `src/vite-env.d.ts:1` references `vite/client` and `tsconfig.json` includes `src`.
- ✅ **The guard is not a silent no-op** — replaying its logic over the 34 bundles: the glob loads 17 files per locale (above the `>10` assertion), all 7 `ALLOWED` ids resolve against the exact `lang/file.json#dotted.key` shape `offenders()` builds, and both offender scans return empty. The "glob really loaded" and "no dead exemptions" assertions make the two classic false-green modes (wrong glob path, mismatched id shape) fail loudly rather than pass vacuously.

## Findings

No P0/P1 issues. Three P2 items, all non-blocking and all stemming from the same exemption decision.

### P2 — `column.spaceId` header left behind in en-US (`src/i18n/locales/en-US/spaces.json:14`)

The table now reads "Organization Name | Space ID | Creator | Members" — `column.name` on `:13` was renamed but the adjacent `column.spaceId` header still says "Space ID", so one header row mixes both terms. It is rendered, not internal: `src/pages/Spaces/index.tsx:274` uses it as the column title. Change it to "Organization ID"; the raw `space_id` remains the cell value, so greppability against the backend field is preserved by the value rather than the header.

### P2 — zh-CN `column.spaceId` is both inconsistent and untranslated (`src/i18n/locales/zh-CN/spaces.json:14`)

Same key on the zh side is still the English "Space ID" while `column.name` became "组织名称", so a fully Chinese header row carries one English leftover. This is the more visible of the two because the surrounding copy is now uniformly Chinese. Suggest "组织 ID" (or "组织 ID（space_id）" if the backend-field hint is worth keeping in the header).

### P2 — The exemption list codifies the leftover, and a second rendered string has the same problem (`src/i18n/locales/terminology.test.ts:14`)

`ALLOWED` pins `spaces.json#column.spaceId` and `appBots.json#detail.scope.space` in both locales, so the guard will keep passing while the inconsistency ships. The second one is also user-visible: `src/pages/AppBots/DetailDrawer.tsx:271` renders `detail.scope.space` inside the drawer of the page whose title this PR renames to "Organization App Bots" / "组织应用 Bot" (`src/i18n/locales/en-US/appBots.json:3`), and both locales ship the identical English string "Space: {{spaceId}}" (`src/i18n/locales/en-US/appBots.json:46`, `src/i18n/locales/zh-CN/appBots.json:46`) — so the zh drawer shows an English "Space" row under a Chinese "组织应用 Bot" title. The remaining five exemptions are legitimate whitespace-sense or backend-identifier cases and should stay. If the two rendered labels are fixed, drop those two ids from `ALLOWED` so the guard covers them going forward; if the English form is a deliberate product decision, it deserves a line in the PR description rather than only a code comment.

- Rewrite-only strings stayed truthful: `en-US/spaces.json:93` and `zh-CN/spaces.json:93` both restate the same pre-existing claim about the add-members endpoint bypassing the member limit, and `dashboard.direct.filteredEmpty.title` was reworded on the en side to match the zh wording that already existed.
- `common:table.space` / `common:visibility.space` are labels only — the `space` enum values and the `scope: 'space' | 'super'` store field are untouched, and all five consuming tables (`SystemMcp`, `SkillMarket/SkillTab`, `ExpertMarket/ExpertTab`, `ExpertMarket/SquadTab`, `SystemSkill/SkillTable`) pass the enum, not the label.
- The zh-CN drive-by punctuation fix at `src/i18n/locales/zh-CN/spaces.json:124` (halfwidth → fullwidth comma in `invites.copyLink.error`) is harmless.
- `README*`, `docs/`, and `CONTRIBUTING*` contain no "Space/空间" terminology, and `RELEASING.md:20` confirms the changelog is release-drafter-generated, so no docs or changelog obligation is left open.

## Verdict: APPROVED

The rename is values-only, key-complete, placeholder-stable, locale-symmetric, and cannot break navigation, persistence, authorization, or CI; the new guard test is genuinely wired into the CI gate and is not a vacuous pass. The three P2 items are user-visible terminology inconsistencies in the two strings the guard exempts — worth fixing in this PR since it is exactly a copy-consistency pass, but they break nothing and are not a merge blocker.

